PR #7, opened 2016-07-25
Try/pod organization

aerych:
This PR moves the podspec inside the pod's source folder, and updates various paths to reflect that change. 
This cleans up some duplicated folder structure inside the Xcode development project while retaining a separation between the pod and example project code bases. 

@diegoreymendez, @bummytime Wha'cha think?


diegoreymendez:
:shipit: Awesome stuff.

PR #8, opened 2016-07-26
Removed the Aztec prefix from some classes.

diegoreymendez:
Removes the Aztec prefix from some classes, since they're already inside the module `Aztec`.

**To test:**

1. Make sure the app builds
2. Make sure the unit tests run properly.
3. Make sure the demo app hasn't lost any functionality.


aerych:
:shipit:

PR #11, opened 2016-07-27
Loads HTML into the demo editor.

diegoreymendez:
**What works:**

Loads HTML into the demo editor.

**What doesn't work:**

Unfortunately there's an issue with how we've implemented the formatting that's making it not work properly.  In any case, since there's value in having the parser wired, I'm submitting this PR now and will work in parallel on a fix for the formatting issues.

**How to test:**

Just launch the demo app, go to "Editor Demo" and make sure the HTML text is loaded.  Some styles may be present, but they will mostly not work yet.


Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -133,7 +197,7 @@ class HMTLNodeToNSAttributedString: SafeConverter {
             string.addAttribute(NSStrikethroughStyleAttributeName, value: 1, range: fullRange)
         }
 
-        string.addAttribute(NSFontAttributeName, value: font(forNode: node), range: fullRange)
+        //string.addAttribute(NSFontAttributeName, value: font(forNode: node), range: fullRange)
```
aerych: still needed?

Classes/Public/Converters/HTMLToAttributedString.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    public func convert(html: String) throws -> NSAttributedString {
+        let htmlToNode = Libxml2.In.HTMLConverter()
+        let nodeToAttributedString = HMTLNodeToNSAttributedString(usingDefaultFontDescriptor: defaultFontDescriptor)
+
+        let node = try htmlToNode.convert(html)
+
+        return nodeToAttributedString.convert(node)
+    }
+}
```
aerych: Maybe add a new line at the end of the file?

aerych:
:shipit: from me :)

diegoreymendez:
Ty @aerych.

Issue #10, opened 2016-07-26
Make the demo App load a sample HTML file.


diegoreymendez:
Title has it.  Formatting should work for the following types:

- Bold
- Italic
- Underline
- Strikethrough

PR #12, opened 2016-07-27
Load HTML in Style!

diegoreymendez:
When loading HTML data, the editor demo now renders:

- Bold Text
- Italic Text
- Underlined Text
- Striked-through text
- Links

**To test:**
Launch the app, go to the editor demo and make sure the styles mentioned above are being rendered properly.

### DIFF
--- a/Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
@@ -182,7 +182,7 @@ class HMTLNodeToNSAttributedString: SafeConverter {
         if isLink(node) {
             let linkURL: String
 
-            if let attribute = node.attributes.indexOf({ $0.name == "href" }) as? Libxml2.HTML.StringAttribute {
+            if let attribute = node.attributes.indexOf({ $0.name == HTMLLinkAttributes.Href.rawValue }) as? Libxml2.HTML.StringAttribute {
                 linkURL = attribute.value
             } else {
                 // We got a link tag without an HREF attribute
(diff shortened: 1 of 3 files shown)


Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -161,62 +155,55 @@ class HMTLNodeToNSAttributedString: SafeConverter {
 
     // MARK: - String attributes
 
-    /// Adds all HTML-meta data and style attributes from the specified element node to the
-    /// specified string.
+    /// Calculates the attributes for this node based.  Returns a dictionary which includes
```
aerych: "based"?  Incomplete sentence maybe?

Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -161,62 +155,55 @@ class HMTLNodeToNSAttributedString: SafeConverter {
...
+            attributes[NSLinkAttributeName] = linkURL
         }
 
         if isStrikedThrough(node) {
-            string.addAttribute(NSStrikethroughStyleAttributeName, value: 1, range: fullRange)
+            attributes[NSStrikethroughStyleAttributeName] = 1
+        }
+
+        if isUnderlined(node) {
+            attributes[NSUnderlineStyleAttributeName] = 1
```
aerych: Maybe `NSUnderlineStyle.StyleSingle.rawValue` instead of `1`, for both `isStrikedThrough` and `isUnderlined` `if` blocks.

aerych:
Maybe out of scope, but I'm wondering if we should go ahead and make consts for all the tag names vs writing out the strings every time.

aerych:
I had just a couple of nitpicks.  Looks good otherwise so :shipit: when ready.

diegoreymendez:
> Maybe out of scope, but I'm wondering if we should go ahead and make consts for all the tag names vs writing out the strings every time.

We should.  I'll take care of your feedback before merging.

Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
diegoreymendez: I had no clue these constants existed.  Thanks for pointing this out!

diegoreymendez:
@aerych: Applied all your recommendations, including the constants for standard tags and attributes.  Would you mind giving this another look?

Thanks!

Classes/Public/Constants/HTMLConstants.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+public enum HTMLTags: String {
+    case Bold = "b"
+    case Delete = "del"
+    case Emphasized = "em"
+    case Italic = "i"
+    case Link = "a"
+    case StrikeThrough1 = "s"
+    case StrikeThrough2 = "strike"
```
aerych: This works but would it be better to just use the name of the tag for the name of the enum case?  `HTMLTags.b` (or maybe `HTMLTags.B`?) is just as clear (if not more so) as `HTMLTags.Bold` and it would avoid having to do things like `StrikeThrough1` and `StrikeThrough2`. 
Thoughts?

diegoreymendez: Considering [there is a `<link>` tag](http://www.w3schools.com/tags/tag_link.asp), I think yours is a good idea.

As for the enum names, I was a bit torn between Swift 2 (first character upper-case) and Swift 3 style (first character lower-case).

The reason I decided to go for upper case is that Swift 3 is not here yet (and even if unlikely there's a chance that rule may never see the light of day?).

diegoreymendez:
@aerych - Ready for another look.  Check out your last (now hidden) comment for my reply since it has some relevant info.

aerych:
:shipit: sir!!!

diegoreymendez:
Ty @aerych!

diegoreymendez:
Fixes #10

PR #15, opened 2016-08-01
Try/blockquotes

aerych:
Adds basic formatting support for blockquotes.

Needs review: @diegoreymendez


Classes/Private/NSAttributedString/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -259,4 +268,8 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     private func isUnderlined(node: ElementNode) -> Bool {
         return node.name == HTMLTags.U.rawValue
     }
+
+    private func isBlockquote(node: ElementNode) -> Bool {
+        return [HTMLTags.Blockquote.rawValue].contains(node.name)
```
diegoreymendez: A bit nit-picky, but for consistency: when there's a single element to compare agains I'm using just the `==` operator above.

Classes/Public/Metrics.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
```
diegoreymendez: Maybe move this file to the `Constants` subfolder (the one that's in the same directory as this file).

Classes/Public/TextKit/AztecVisualEditor.swift
```diff
@@ -46,6 +45,49 @@ public class AztecVisualEditor : NSObject
     }
 
 
+    // MARK: - Misc helpers
```
diegoreymendez: I'd consider replacing "Misc helpers" with "Paragraph" (or something similar, since it'd be more specific to what's inside the group).

diegoreymendez:
@aerych: just added a few minor comments.

The PR is working very nicely though.  So feel free to :shipit: if there are no big changes following my feedback, or ping me if necessary.

Classes/Public/Metrics.swift
aerych: 👍  I didn't notice the Constants folder after I merged develop.  Def belongs in there.

Classes/Public/TextKit/AztecVisualEditor.swift
aerych: I'm going to leave it as is for now. I expect `paragraph` is too specific and I'm betting there will be some other helper type methods related to lists that we can group there.  Better naming should become apparent down the road.

aerych:
Thanks @diegoreymendez !!

PR #20, opened 2016-08-02
Adds several unit tests and improves TextNode's initialization.

diegoreymendez:
Adds several unit tests and improves TextNode's initialization by removing the `name` parameter (which isn't really needed, and actually dangerous to provide editing access to).

**To test:**
1. Make sure the code compiles, and that the editor still loads properly.
2. Run the unit tests and make sure they work.


aerych:
Code looks good and all tests pass. 
:shipit: sir!

diegoreymendez:
Thanks @aerych!

PR #23, opened 2016-08-03
Changes the code a bit to remove an IUO.

diegoreymendez:
Changes the code a bit to remove an IUO.  The property was only an IUO due to an initialization limitation - the property was trully non-optional and this change helps to better reflect that.


aerych:
:shipit:

PR #25, opened 2016-08-08
Cleans up trait toggling.

diegoreymendez:
Cleans up trait toggling.  This PR introduces changes to avoid accessing `rawValue` where possible.

**How to test:**
Toggle bold, italic, underline, etc... for a range of text.
Make sure that the style is applied if any section of the range does not have the style.
Make sure that the style is removed if the full range has the selected style.

### DIFF
--- a/Classes/Public/TextKit/AztecTextStorage.swift
+++ b/Classes/Public/TextKit/AztecTextStorage.swift
@@ -131,15 +131,15 @@ public extension AztecTextStorage
                                 return
                             }
 
-                            var newTraits: UInt32
-                            if assigning {
-                                newTraits =  font.fontDescriptor().symbolicTraits.rawValue | trait.rawValue
+                            var newTraits = font.fontDescriptor().symbolicTraits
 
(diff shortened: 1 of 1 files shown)


Classes/Public/TextKit/AztecTextStorage.swift
```diff
@@ -131,15 +131,15 @@ public extension AztecTextStorage
                                 return
                             }
 
-                            var newTraits: UInt32
-                            if assigning {
-                                newTraits =  font.fontDescriptor().symbolicTraits.rawValue | trait.rawValue
+                            var newTraits = font.fontDescriptor().symbolicTraits
 
+                            if assigning {
+                                newTraits.insert(trait)
```
aerych: omg that is so much nicer.

aerych:
Ship it.  Ship it NAO! And if you can ship it twice... do that too.
 :shipit:

diegoreymendez:
Thanks Eric 😄

PR #26, opened 2016-08-08
Fixes the HTML output conversion.

diegoreymendez:
Fixes the HTML output conversion.

**How to test:**
The best way to test this is to integrate it into `feature/19-html-mode`.  The conversion is not perfect due to other issues we'll need to tackle, but it should be enough to keep making progress.


jleandroperez:
Fix confirmed! :shipit:

PR #27, opened 2016-08-08
Feature: HTML Mode

jleandroperez:
@diegoreymendez + @aerych mind taking a quick look?

I've simplified `EditorDemoController` a little bit: nuked setters that weren't being used anywhere.

Suggestions would be very welcome!

### DIFF
--- a/Example/WordPress-Aztec-iOS/EditorDemoController.swift
+++ b/Example/WordPress-Aztec-iOS/EditorDemoController.swift
@@ -90,60 +90,7 @@ class EditorDemoController: UIViewController
     }
 
 
-    var titleFont: UIFont? {
-        get {
-            return titleTextField.font
-        }
-        set {
-            titleTextField.font = newValue
-            layoutTextView()
-        }
-    }
-
-
-    var titleColor: UIColor? {
-        get {
-            return titleTextField.textColor
-        }
-        set {
-            titleTextField.textColor = newValue
-        }
-    }
-
-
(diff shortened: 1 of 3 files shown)


Example/WordPress-Aztec-iOS/EditorDemoController.swift
```diff
@@ -5,20 +5,20 @@ import Aztec
 
 class EditorDemoController: UIViewController
 {
-    private var bottomConstraint: NSLayoutConstraint?
+    private var bottomConstraint: NSLayoutConstraint!
 
 
     private (set) lazy var editor: AztecVisualEditor = {
-        let e = AztecVisualEditor(textView: self.textView)
+        let e = AztecVisualEditor(textView: self.richTextView)
```
diegoreymendez: I know this wasn't introduced by your PR but I'd take the opportunity to remove the `e` variable, both for having a non standard name, and for not being really necessary.

diegoreymendez:
Just 1 optional comment @jleandroperez and then :shipit: .  Nice work.

Example/WordPress-Aztec-iOS/EditorDemoController.swift
aerych: +1 I think I did that just to inspect the variable before it was returned, but its not something that should have shipped

aerych:
Seconded! :shipit:

jleandroperez:
Thanks gentlemen!!!

PR #28, opened 2016-08-10
Text nodes no longer support attributes.

diegoreymendez:
Text nodes no longer support attributes, since they really should never have. 😄 

This PR cleans up text nodes to only accept their text content in their initializer.

**How to test:**
Build the app.
Run the unit tests and make sure they succeed.


bummytime:
Looks good @diegoreymendez :shipit:

PR #29, opened 2016-08-12
Improves the root node handling.

diegoreymendez:
Improves the root node handling mechanism.

It may still not be perfect, but it does look much better than what we had before.  I need these changes to implement the DOM editing logic for [this issue](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/issues/18).

**How to test:**
1. Review the logic and code.
2. Make sure the unit tests still run.
3. Run the app, go to the editor, make sure everything loads.
4. Still at the editor, make sure editing text and styles still works visually.
5. Switch back and forth to HTML and RichText mode.  The edits won't be shown, but make sure the initial text is loaded right.


diegoreymendez:
This was already added by a larger PR.  Closing.

Issue #18, opened 2016-08-02
Initial Editing Logic for Bold


diegoreymendez:
We need to implement the logic to update the DOM data whenever an edit takes place.  The best way to start is by focusing on the full editing logic for a singly style, in order to tackle the problem atomically.

We should expect this code to be reusable for other styles.

Scenarios to handle:

- [x] Toggling Bold
 - [x] When the full selection text is bold.
 - [x] When the full selection text is non-bold.
 - [x] When the full selection text is mixed.

diegoreymendez:
Sorry @jleandroperez, I assigned this one to you by mistake.

PR #31, opened 2016-08-15
Alpha version of bold editing logic.

diegoreymendez:
Tackles part of #18 

I was planning to split this PR into several PRs, since there are some improvements not-strictly-related to the editing logic.  However since we're two days away from the sprint's end, and there's still [so much to do](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Iteration+%233.%22) (and we're not working in production code), I'd rather have this merged and deal with any issues as they arise.

**How to test:**
1. Run the app
2. Go to the editor
3. Select non-bold text and press the B button in the toolbar.
4. Switch to HTML mode and make sure the tag is added to the selected text.

**What doesn't work:**
1. Removing bold
2. The way in which bold is added is kinda primitive now - there's no logic to merge adjacent bold elements into a single one, so the results are a lot of bold tags everywhere.  This is known and accepted for this alpha.


Classes/Private/Libxml2/Converters/In/InHTMLConverter.swift
```diff
@@ -4,6 +4,8 @@ import libxml2
 extension Libxml2.In {
     class HTMLConverter: Converter {
 
+        typealias RootNode = Libxml2.RootNode
```
aerych: nice

Classes/Private/Libxml2/Converters/In/InHTMLConverter.swift
```diff
@@ -76,7 +77,7 @@ extension Libxml2.In {
                 let nodeConverter = NodeConverter()
                 let node = nodeConverter.convert(rootNode)
```
aerych: Maybe something like this for paranoia?
```
if let node = nodeConverter.convert(rootNode) as? RootNode {
return node
}
throw Error.NoRootNode
```

Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -0,0 +1,242 @@
+import UIKit
```
aerych: Why UIKit?  Maybe Foundation?

Classes/Private/Libxml2/Data/Node.swift
```diff
@@ -14,47 +14,102 @@ extension Libxml2.HTML {
...
-            return Mirror(self, children: ["type": "text", "name": name, "text": text, "parent": parent], ancestorRepresentation: .Suppressed)
+            return newNode
         }
     }
 }
 
 // MARK: - Node Equatable
 
-func ==(lhs: Libxml2.HTML.Node, rhs: Libxml2.HTML.Node) -> Bool {
+func ==(lhs: Libxml2.Node, rhs: Libxml2.Node) -> Bool {
```
aerych: Stylin' :)

Classes/Private/Libxml2/Data/ElementNode.swift
diegoreymendez: I think this was added by default when I created the file and I never changed it.

diegoreymendez:
@aerych - Fixed the 2 comments you posted.  Let me know if there's anything else.

Classes/Private/Libxml2/Converters/In/InAttributeConverter.swift
```diff
@@ -4,8 +4,8 @@ import libxml2
 extension Libxml2.In {
     class AttributeConverter: SafeConverter {
 
-        typealias Attribute = HTML.Attribute
-        typealias StringAttribute = HTML.StringAttribute
+        typealias Attribute = Libxml2.Attribute
+        typealias StringAttribute = Libxml2.StringAttribute
```
aerych: There are lots of places where typealias's are defined for Libxml2 classes.  Maybe it would make sense to define them all once, maybe in Libxml2, and then use them through out the library?

Classes/Public/TextKit/AztecAttachmentManager.swift
```diff
@@ -25,7 +25,7 @@ import Foundation
         self.textView = textView
         self.delegate = delegate
 
-        super.init()
+        //super.init()
```
aerych: Maybe delete?

Classes/Public/TextKit/AztecAttachmentManager.swift
```diff
@@ -214,12 +214,12 @@ import Foundation
...
     private func frameForAttachmentView(attachmentView: AztecAttachmentView, forAttachment attachment: AztecTextAttachment, atRange range:NSRange) -> CGRect {
         let glyphRange = layoutManager.glyphRangeForCharacterRange(range, actualCharacterRange: nil)
-        guard let textContainer = layoutManager.textContainerForGlyphAtIndex(glyphRange.location, effectiveRange: nil) else {
+        guard let _ = layoutManager.textContainerForGlyphAtIndex(glyphRange.location, effectiveRange: nil) else {
             return CGRectZero
         }
 
         // The location of the attachment glyph
-        let glyphBoundingRect = layoutManager.boundingRectForGlyphRange(glyphRange, inTextContainer: textContainer)
+        //let glyphBoundingRect = layoutManager.boundingRectForGlyphRange(glyphRange, inTextContainer: textContainer)
```
aerych: maybe delete?

Classes/Public/TextKit/AztecAttachmentManager.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 /// Wrangles attachment layout and exclusion paths for the specified UITextView.
 ///
-@objc public class AztecAttachmentManager : NSObject
```
aerych: I'm trying to remember why I had added `@objc` and extended `NSObject`. I borrowed this class from WPTextKit and then did some tweaks so maybe it was something I missed. Or maybe I was just thinking we'd also have objc support.  
Not necessary for now regardless so 👍

aerych:
Hola @diegoreymendez 
I had a couple more comments. Not quite sure about the `typealias` thing.  If there is a reason to keep them on a per class basis then that's cool too.  
Everything works in testing so once the conflicts (and optionally the typealias) are fixed it should be good to :shipit:

Classes/Private/Libxml2/Converters/In/InAttributeConverter.swift
diegoreymendez: I'd prefer to avoid an Aztec-wide definition for this for now - for the time being I'd prefer to let each class define its own `typealiases`.

I'm open to reviewing this once we're closer to final, though.

PR #32, opened 2016-08-16
Fixed an error in the bold-editing-logic.

diegoreymendez:
There was an error with the bold editing logic that was causing a crash.

Brings us closer to closing #18.

**How to reproduce the original error:**
1. Launch the editor from `develop`.
2. Select the word "post" from the second line.
3. Make it bold.
4. Now select the full line, including the word "post".
5. Make it bold again.
6. Crash!

**How to test:**
Same steps as above, but they should no longer crash and the text should be made bold.


aerych:
:shipit: !!

PR #39, opened 2016-08-24
Implements block-level node finding, and node splitting.

diegoreymendez:
This PR implements several logical instruments we'll need to update the DOM.

- Block-level node finding basically means we can find all lowest (hierarchy-wise) block-level nodes intercepting a range.
- Splitting nodes and their children using a range.

**How to test:**
- Since this PR is mostly non-hooked logical methods, there's not much that can be tested from the GUI yet.
- Run the unit tests and make sure they succeed.

### DIFF
--- a/Classes/Private/Libxml2/Data/ElementNode.swift
+++ b/Classes/Private/Libxml2/Data/ElementNode.swift
@@ -402,7 +402,7 @@ extension Libxml2 {
         /// Removes the specified child nodes.  Only updates their parents if specified.
         ///
         /// - Parameters:
-        ///     - chldren: the child nodes to remove.
+        ///     - children: the child nodes to remove.
         ///     - updateParent: whether the children node's parent must be update to `nil` or not.
         ///             If not specified, the parent is updated.
         ///
--- a/Classes/Private/Libxml2/Data/ElementNode.swift
+++ b/Classes/Private/Libxml2/Data/ElementNode.swift
@@ -302,7 +302,7 @@ extension Libxml2 {
             var result = [Node]()
             var offset = length()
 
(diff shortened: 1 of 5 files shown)


Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -215,6 +366,322 @@ extension Libxml2 {
...
+            }
+
+            children.removeAtIndex(childIndex)
+            children.insertContentsOf(newChildren, at: childIndex)
+        }
+
+        /// Removes the specified child nodes.  Only updates their parents if specified.
+        ///
+        /// - Parameters:
+        ///     - chldren: the child nodes to remove.
```
SergioEstevao: small type in parameter name.

Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -215,6 +366,322 @@ extension Libxml2 {
...
+                    break
+                } else {
+                    break
+                }
+            }
+            
+            return result
+        }
+
+        /// Retrieves all child nodes positioned before a specified location.
```
SergioEstevao: all children nodes?

SergioEstevao:
Looking good @diegoreymendez . Just made some small notes about typos in the comments. After you fix those :shipit:

Classes/Private/Libxml2/Data/ElementNode.swift
diegoreymendez: So in theory it would be either "all children" (without the word "nodes") or "all child nodes".

diegoreymendez:
Thanks @SergioEstevao!

PR #40, opened 2016-08-25
Apply bold style

diegoreymendez:
It's now possible to apply bold-style to text runs.  It's also smart about how it wraps existing nodes, and how it inserts and breaks the bold styling (where necessary).

Related to #18.

**Known limitations:**
1. It doesn't check for pre-existing bold styling, so it's possible to end up having HTML such as `<strong><b>hello</b></strong>`.
2. Removing bold doesn't affect HTML.

**How to test:**
1. Run the editor demo.
2. Select any range of text.
3. Make it bold.  Make sure the HTML is affected by switching to HTML mode.
4. Make sure to run the unit tests too.


SergioEstevao:
:shipit:

Issue #33, opened 2016-08-17
Smart tag grouping & reorganization when adding or removing tags.


diegoreymendez:
When adding a tag to the DOM, we need to reduce tag fragmentation to the minimum.  This implies that nodes need to be added in a "smart" way.

The following is an example showing the selected text wrapped by `[` and `]`.

`<em>Hello there<u>[ man!</u> How are you?]</em>`

After applying bold, the non-smart result would be:

`<em>Hello there<u><b> man!</b></u><b> How are you?</b></em>`

The smart result we must achieve is:

`<em>Hello there<b><u> man!</u> How are you?</b></em>`

Issue #41, opened 2016-08-28
Initial Editing Logic for Italic


diegoreymendez:
Implements style editing for Italic.  The result should affect the HTML output.

Scenarios to handle:

- [x] Toggling Italic
 - [x] When the full selection text is Italic.
 - [x] When the full selection text is non-Italic.
 - [x] When the full selection text is mixed.

PR #42, opened 2016-08-28
Implements bold and italic styles.

diegoreymendez:
Implements bold and italic style editing.  This means you should be able to toggle bold and italics on and off, following the editing logic of our previous editor.

Fixes: #18 and #41.

**Known issues:**
- There's an issue related to our HTML importer not ignoring some characters that can cause node duplication if you switch back and forth several times between visual and HTML mode while editing styles.  This will be fixed separately.

### DIFF
--- a/Classes/Private/Libxml2/Data/ElementNode.swift
+++ b/Classes/Private/Libxml2/Data/ElementNode.swift
@@ -300,7 +300,7 @@ extension Libxml2 {
                                 if let previousRangeWithoutMatch = rangeWithoutMatch {
                                     rangeWithoutMatch = NSRange(location: previousRangeWithoutMatch.location, length: previousRangeWithoutMatch.length + range.length)
                                 } else {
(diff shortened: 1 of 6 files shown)


SergioEstevao:
If you try to select text and select both bold and italic for the text, while it looks correct on the visual editor when I go to the HTML view the resulting markup is wrong. Screenshots here:

screen shot 2016-08-29 at 10 45 37

screen shot 2016-08-29 at 10 45 48



It also looks that after doing the first switch from the visual to the html editor afterward the html editor get stuck in the first conversion.

diegoreymendez:
@SergioEstevao - Good catch.

Fixed the issue you spotted and added a unit test for this specific case.

SergioEstevao:
Good work @diegoreymendez :shipit:

diegoreymendez:
Thanks!

diegoreymendez:
This PR also closed #33

Issue #44, opened 2016-08-29
Initial Editing logic for Strike-through


diegoreymendez:
Implements style editing for strikethrough.  The result should affect the HTML output.

Scenarios to handle:

- [x] Toggling strikethrough
 - [x] When the full selection text is strikethrough.
 - [x] When the full selection text is not striked-through.
 - [x] When the full selection text is mixed.

PR #45, opened 2016-08-29
Adds underline and strikethrough support.

### LINKED ISSUE #43: Initial Editing Logic for Underline
Implements style editing for underline.  The result should affect the HTML output.

Scenarios to handle:

- [x] Toggling underline
 - [x] When the full selection text is underline.
 - [x] When the full selection text is non-underline.
 - [x] When the full selection text is mixed.

diegoreymendez:
Fixes #43 and #44.

The same testing procedure and limitations as #42 applies.


SergioEstevao:
:shipit:

Issue #34, opened 2016-08-17
Ordered and unordered lists.


diegoreymendez:
Enable the lists button and add the logic to edit lists in the editor.  List behavior is as follows:
### Toolbar:
- [x] OL button should highlight when cursor is in an order list
- [x] UL button should highlight when cursor is in an unordered list
#### Tapping a List Button:
- [x] **Scenario 1**: When the cursor is in a plain paragraph: 
  `Should create a new list, and turn the whole paragraph into a list item.`
- [x] **Scenario 2**: When the cursor is in a list of the same type: 
  `Should turn the item into a plain paragraph. Following list items should be renumbered.`
- [x] **Scenario 3**: When the cursor is in a list of a different type: 
  `Should change the type of the entire list, and update the list markers appropriately.`
- [x] **Scenario 4:** When the selected text starts with a paragraph then intersects a list of the same type: 
  `Should add the non-list text to the list and update the item numbering.`
- [x] **Scenario 5:** When the selected text starts with a paragraph then intersects a list of a different type: 
  `Should create a new list from the selected paragraph, but do not make changes to the other list.`
- [x] **Scenario 6:** When the selected text starts with a list of the same type and intersects a non-list paragraph: 
  `Should Remove the list.`
- [x] **Scenario 7:** When the selected text starts with a list of a different type and intersects a non-list paragraph:
  `Should change the type of the list and update its numbering, but make no changes to the non list paragraph.`
### Selection:
- [x] **Scenario 1:** Whenever there was no previous selected text (just the cursor blinking on a line), and a List was applied, the cursor should shift to match the spot at which it was previously located. In another words: we should apply a delta, to consider the new Text Marker.
- [x] **Scenario 2:** Whenever a list is applied over a Selected Range, we should expand the selection to consider any length-deltas, caused by Text Markers.
### Cursor:
- [x] If the current selection intersects a list marker, adjust the selection so it does not intersect the marker.  Move the cursor forward or backward depending on the previous cursor location.
### Typing:
- [x] Apply the current list and list item custom attributes to newly typed text (keyboard input does not automatically assign custom attributes.)
### Carriage returns:
- [x] When there is a carriage return inside a list. Insert a new list marker.  Update the numbering of any following items. 
- [x] When the cursor is on an empty list item (just the marker and no other text) and a carriage return is added.  Remove the empty list item, and insert a normal paragraph in its place.  If there are any following list items, replace those items with a new list and new item numbering. 
[...]

PR #48, opened 2016-08-30
Try/lists

aerych:
Refs #34 

Provides support for formatting and editing ordered and unordered text lists.

The list behavior exists in a few places. 
- `TextList` Simple types providing attribute names, and values for NSAttributedStrings.  I chose to uses classes vs structs initially, but structs might just as well.
- `ListFormatter` Provides various methods for formatting lists and retrieving information about a list.
- `AztecVisualEditor` assigns itself as the UITextViewDelegate in order to mange the cursor when it would overlap/intersect a list marker (bullet or number).  This conflicts with other classes who would implement UITextViewDelegate methods (like the EditorDemoController). The plan here is to proxy UITextViewDelegate calls via `AztecVisualEditorDelegate`.
- `AztecTextStorage` handles cases where a list would be modified by a carriage return or backspace via `public override func replaceCharactersInRange(range: NSRange, withAttributedString attrString: NSAttributedString)`

There are a few things left to be sorted out with the implementation. 
- The selected text range is not quite right after applying a list style.  You can see this by selecting multiple lines and tapping either list button on the format bar.  The selected text will include the previous line. 
- The blockquote style works with paragraph indentation, and as such it is incorrectly highlighting the blockquote button in addition to the list buttons when the cursor intersects a list.  This can be addressed when we update the blockquote style to look for our custom dom attribute.  Out of scope for this feature I think.
- There are issues updating the list for carriage returns and backspacing. The range of the changes is off some where causing text after the list to also be edited, or additional list items to appear or markers to fail to delete.
- Everything needs unit testing.

cc @diegoreymendez


diegoreymendez:
@SergioEstevao - @jleandroperez was using this PR to slowly add code for handling lists into Aztec.  I'm not too aware of how much remains to be copied or if that's still valid for our project, but considering you took over, maybe you can evaluate if this can be closed.

Let me know if that's the case or if we still need this to remain open.

jleandroperez:
@diegoreymendez @SergioEstevao as far as i remember, there were few helpful snippets:

- Code to avoid selecting the bullet
- Aaand `TextStorage` (i think!) snippets to deal with newlines

SergioEstevao:
I already did the first... I will have a look on the text storage for the new line methods.

SergioEstevao:
Closing this PR, all the functionality and interesting bit were already ported in the code base.

PR #49, opened 2016-08-30
Fixes attributes conversion.

diegoreymendez:
Fixes our attributes conversion logic.

When going back to HTML-mode, all attributes were being removed.

**How to test:**
1. Launch the demo editor.
2. Go to HTML mode.
3. Make sure all the HTML attributes from the original HTML are showing up.

Props to @SergioEstevao for spotting this issue.


SergioEstevao:
:shipit:

PR #50, opened 2016-08-30
Removes some unused custom attribute information.

diegoreymendez:
We were initially storing some custom attributes representing HTML tags in our editor's final visual string.

These are no longer needed as its not through this attributes that we're editing HTML tags.

**How to test:**
1. Make sure that when loading a post the styles are still working.
2. Make some simple editions to a post, and make sure when switching to HTML-mode that they still work as they do in `develop`.
3. Make sure the unit tests still run correctly.


jleandroperez:
Looking good!!

:shipit:

PR #54, opened 2016-08-31
Attachments: Alignment Support! - Part I

jleandroperez:
#### Details:
This PR switches from `NSTextAttachment` over to `AztecTextAttachment`. Why:

- Because by doing so, we gain absolute control over the way Text Attachments are laid out
- It's what our friend @aerych initially intended
- This is actually a requirement for us to properly support alignment / rotations

#### Testing
Please, try inserting an image / multiple images in the following cases:

- Very first line of the sample text
- Middle of any block (indented or not)
- Last character in the text

Please, make sure that the text properly excludes the area defined by the image. 
Also try adding newlines, and make sure the TextView moves along.

#### Pendings:
We'll be addressing the following pendings in "Part II":

- TextWrapping Modes
- Alignment / Resizing / Rotations

Needs Review: @diegoreymendez @aerych 
Thanks in advance!!


Classes/Public/TextKit/AztecAttachmentManager.swift
```diff
@@ -239,20 +246,21 @@ public class AztecAttachmentManager
...
             return
         }
 
         let view = attachmentView.view
-        if view.frame.height == 0 {
+        guard view.frame.height != 0 else {
             return
         }
 
+        let maximumWidth = container.size.width - 2 * container.lineFragmentPadding
```
aerych: maybe some parenthesis for increased readability?

jleandroperez:
Pasting below @aerych 's feedback. I'll be working on this, thanks Eric!!


```
but consider a situation where we have a line of text and the attachment falls in the middle of the line, like so:
`lorem ipsum [attachment] dolar sit amet`

Assuming we have an image that’s the width of the screen, what I think should happen is this:
The line fragment following `ipsum` should be blank space.  the exclusion path should start on the following line, and then the rest of the text would finish on the line after the exclusion path.  So something like:

lorem ipsum
[exclusion path + custom view]
dolar sit amet

but right now what we get is

lorem ipsum dolar sit amet
[exclusion path + custom view]

as shown in the draggable demo vc.

To get the line fragment to clear all the following text, I did the trick you can see here: https://github.com/wordpress-mobile/WPTextKit-iOS/blob/develop/WPTextKit/WPTextKit/WPTextAttachment.swift#L89

basically overriding attachmentBoundsForTextContainer and reserving the necessary space to clear the line.

Not sure if that’s the best approach but it worked.
```

diegoreymendez:
:shipit:

jleandroperez:
Thanks Diego!!!

PR #55, opened 2016-09-01
Add support for HTML comment nodes.

SergioEstevao:
How to test:

 - Start the demo app
 - Switch between Visual and HTML mode
 - see if the  tag is being converted properly
 - add a new comment to the HTML view, switch to visual mode, and them back to visual

Some questions to answer in the future:

 - Should comments be visible in Visual mode?
 - Should we allow editing or deleting in visual mode?
 - We still need to add an extra case for the  to be parsed visually as an separator.

### DIFF
--- a/Classes/Private/Libxml2/Data/CommentNode.swift
+++ b/Classes/Private/Libxml2/Data/CommentNode.swift
@@ -24,19 +24,19 @@ extension Libxml2 {
         }
 
         override func deleteCharacters(inRange range: NSRange) {
-            assertionFailure("No-op.  This method will be removed from this type of node.  Please remove this call.")
+            // No-op, will be removed.
         }
 
         override func replaceCharacters(inRange range: NSRange, withString string: String) {
-            assertionFailure("No-op.  This method will be removed from this type of node.  Please remove this call.")
+            // No-op, will be removed.
         }
 
         override func split(forRange range: NSRange) {
-            assertionFailure("No-op.  This method will be removed from this type of node.  Please remove this call.")
+            // No-op, will be removed.
         }
 
         override func wrap(range targetRange: NSRange, inNodeNamed nodeName: String, withAttributes attributes: [Attribute]) {
-            assertionFailure("No-op.  This method will be removed from this type of node.  Please remove this call.")
+            // No-op, will be removed.
         }
     }
 }
--- a/Classes/Private/Libxml2/Data/CommentNode.swift
+++ b/Classes/Private/Libxml2/Data/CommentNode.swift
@@ -28,8 +28,8 @@ extension Libxml2 {
             guard let textRange = text.rangeFromNSRange(range) else {
                 fatalError("The specified range is out of bounds.")
             }
-
-            text.removeRange(textRange)
+            // When a comment is deleted we take it all in one go
+            text.removeAll()
         }
 
         override func replaceCharacters(inRange range: NSRange, withString string: String) {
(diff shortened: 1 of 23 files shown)


SergioEstevao:
I had look in Calypso editor and found the following:

 - The content of the comments are not show in visual mode.
 - They are represented has a break line
 - If you delete the line the comment is removed.

Classes/Private/Libxml2/Data/CommentNode.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        override func customMirror() -> Mirror {
+            return Mirror(self, children: ["type": "text", "name": name, "text": text, "parent": parent.debugDescription], ancestorRepresentation: .Suppressed)
+        }
+
+        /// Node length.
+        ///
+        override func length() -> Int {
+            return text.characters.count
```
diegoreymendez: The length should be zero for this node, since it is intended for content between the opening and closing tags.  In fact if you run the demo app, and delete a range of text after the comment in this PR you'll notice than when switching to HTML the removed range will be a different one (due to misalignment).

This is a more graphical explanation of what length represents:

```html
<b>...child nodes go here...</b>: length is the sum of the lengths of all child nodes
<textNode>Hello there!</textNode>: length is the length of "Hello there!"
<br>: length is zero since there's no opening and closing tag
<!-- Some comment -->: length is zero since there's no opening and closing tag
```

We can rename `length()` to something more specific if necessary.

diegoreymendez: Alternatively, we could leave the length as it is, but ignore it for the editing logic, since editing comments could come in handy.

Classes/Private/Libxml2/Data/CommentNode.swift
```diff
@@ -0,0 +1,93 @@
...
+        override func replaceCharacters(inRange range: NSRange, withString string: String) {
+
+            guard let textRange = text.rangeFromNSRange(range) else {
+                fatalError("The specified range is out of bounds.")
+            }
+
+            text.replaceRange(textRange, with: string)
+        }
+
+        override func split(forRange range: NSRange) {
```
diegoreymendez: This method shouldn't be present in comment nodes.  I can see this will complicate my initial design a bit, so how to remove this cleanly is open for discussion.

diegoreymendez:
@SergioEstevao - Very nice work overall, there are just a few design considerations though that we can discuss through private messages if you want.

I think we'll need to change my initial design approach a bit.

SergioEstevao:
@diegoreymendez I'm struggling to make the comment node to be completely removed when you delete the new line. Can you have a look on the code and see what I'm missing here?

diegoreymendez:
@SergioEstevao - I went in and modified the PR a bit - sorry in advance if you don't like the changes... feel free to roll them back.  I also merged the latest from `develop` since there is an important fix in the editing logic there.

The code now properly removes the comment node.

There is still one issue that we could decide to fix in a separate PR if you want: if you place the cursor at the start of the comment node, in visual mode and start typing, some of the content will be lost when you switch to HTML mode.

This is because the editor is trying to insert the text inside the comment node, which shouldn't be allowed (but the logic to fix this will require some extra thinking).

diegoreymendez:
PS: if you want to talk about my changes let me know.

SergioEstevao:
@diegoreymendez looking good, thanks for the help!

diegoreymendez:
@SergioEstevao - On my end, feel free to :shipit:

For whatever improvement we need to make on top of this, we can work with new PRs.

PR #59, opened 2016-09-05
Fixes an issue with the default font being changed.

### LINKED ISSUE #58: Selection causes the default font to change.
There's a very weird bug that's causing the Visual Editor default font to change, if you select a styled text-range and switch to HTML and back to Visual.

Video showing the issue [here](https://cloudup.com/cyBYbeDF2fv).

diegoreymendez:
Fixes #58.

**How to test:**

1. Run the app.
2. Open the editor with sample content.
3. Select an italic or bold range of text.
4. Switch to HTML mode.
5. Switch back to visual mode and make sure the default font wasn't changed.


jleandroperez:
:shipit: once the merge conflict is gone!

PR #62, opened 2016-09-06
Attachments: Alignment Support Mark II

jleandroperez:
### Description:
In this PR we're implementing support for:

- Attachment Alignment
- Attachment Resizing
- Dynamic Resize upon Rotation

### Details:
We're updating the mechanism used by the TextAttachments. Reasoning is as follows:

- At least initially, we'll support **Inline Attachments** and **Above and Below** attachments (say, on its own "Block", replicating Apple Pages's behavior)
- We should be able to delete attachments by means of the *backspace* button
- The Exclusion Paths mechanism is a way of indicating the TextView not to allow text in specific regions. Since the Attachments live within the text itself, and not as an external entity, it's a good idea to deal with this by means of the 'Attachment Bounds'.

For all of the above, in this PR we're shifting from the **Exclusion Paths** mechanism, to something simpler: Instead of calculating the attachment's frame ourselves, we'll provide the "Required Size" that the attachment should take, and we let TextKit do the heavy lifting for us.

### Internals:
- The Attachment will take the container's full width, whenever the Alignment is set to .Left / .Center / .Right, and will occupy the actual associated view's height, always
- `AztecTextAttachment.attachmentBoundsForTextContainer` reserves the actual space, and the Associated View's size doesn't really need to match. It might be equal or smaller!

### Testing:
- Update the Alignment [here](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/62/files#diff-b3d458fc1044eeda87d98ae1bc8e4dbdR18)
- Update the Image Size [here](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/62/files#diff-b3d458fc1044eeda87d98ae1bc8e4dbdR22)
- Launch the editor, and insert an image
- Verify that after rotating the device, the image properly resizes (to the expected maximum)
- Verify that Alignment.None allows text on the right / left
- Verify that Alignment.Left / .Center / .Right takes the full TextView's width

Needs Review: @diegoreymendez 
cc @aerych 

### Notes:
- Don't worry, we'll implement the actual UI in another PR
- There are *two* TODO's in `AztecTextAttachment`, to be revisited in yet another PR

### DIFF
--- a/Classes/Public/TextKit/AztecTextAttachment.swift
+++ b/Classes/Public/TextKit/AztecTextAttachment.swift
@@ -5,19 +5,6 @@ import Foundation
 ///
 public class AztecTextAttachment: NSTextAttachment
 {
-    /// Supported Types
-    ///
-    public enum Kind {
-        case Image(image: UIImage)
-    }
-
-    /// Wrapping Options. Analog to what Apple Pages does!
-    ///
-    public enum TextWrapping {
-        case Around
-        case AboveAndBelow
-    }
-
     /// Identifier used to match this attachment with a custom UIView subclass
     ///
     private(set) public var identifier: String
@@ -26,9 +13,13 @@ public class AztecTextAttachment: NSTextAttachment
     ///
     public var kind: Kind?
 
-    /// Wrapping Mode
+    /// Attachment Size
     ///
-    public var textWrapping = TextWrapping.AboveAndBelow
+    public var size: Size = .Maximum
+
+    /// Indicates the Height to be occupied onscreen
+    ///
+    var lineHeight = CGFloat.min
 
 
     /// Designed Initializer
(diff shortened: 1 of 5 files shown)


Classes/Public/TextKit/AztecTextAttachment.swift
```diff
@@ -38,27 +34,82 @@ public class AztecTextAttachment: NSTextAttachment
...
+    ///
+    public enum Size {
+        case Thumbnail
+        case Medium
+        case Large
+        case Maximum
+
+        var targetWidth: CGFloat {
+            switch self {
+            case .Thumbnail:    return Settings.thumbnail
```
diegoreymendez: Single line after `:` to avoid the extra alignment work during maintenance.

Classes/Public/TextKit/AztecTextAttachment.swift
```diff
@@ -38,27 +34,82 @@ public class AztecTextAttachment: NSTextAttachment
...
+            case .Medium:       return Settings.medium
+            case .Large:        return Settings.large
+            case .Maximum:      return Settings.maximum
+            }
+        }
+
+        private struct Settings {
+            static let thumbnail    = CGFloat(135)
```
diegoreymendez: Same. 😄

diegoreymendez:
@jleandroperez - This is excellent!  Very nice work.

Can you check out my two comments above about alignment though? 😈  hehe

After that feel free to :shipit:

jleandroperez:
Thanks a lot for the review sir!!!

PR #65, opened 2016-09-08
Attachments: Interface

jleandroperez:
#### Details:
In this PR we're adding a super simple ViewController, that allows us to refresh both, Sizing and Alignment properties.

#### Testing:
1. Add a new image
2. Tap on it
3. Verify that the UI properly reflects any of the new settings you might select

Please, also verify that the Gestures Recognizer doesn't lock scroll events, even when performed on top of the image.

Needs Review: @diegoreymendez 
Thanks in advance!


diegoreymendez:
The new VC is working great, however, the way in which we're invoking it makes it no longer possible to copy an image to the clipboard.

My suggested change would be in line with what other apps do:

- A short tap could open up the image in full screen.
- A long tap could open up an action sheet with options (like Mail.app and Slack.app do).  One of the options should be to copy the image.

I'm ok with taking care of these in a different ticket though, so let me know if you want me to create one.

diegoreymendez:
Fix the conflicts and :shipit: !

jleandroperez:
Thanks for the review @diegoreymendez!!

PR #67, opened 2016-09-16
Using Carthage for the DemoApp and reorganizing files.

diegoreymendez:
Switches from CocoaPods to Carthage for the Example App.

The library will still be available using CocoaPods as long as we keep a working podspec file.

**Before testing:**

- Run `git status` (right after you `git checkout`) to make sure there's no old files laying around.
- You'll need to have [Homebrew](http://brew.sh) and [Carthage](https://github.com/Carthage/Carthage/blob/master/README.md) installed.
- Go to the `Example` folder in terminal and run this command `carthage update --platform iOS` to install the dependencies for testing.

**How to test:**

1. Open the `Aztec.xcworkspace` in the main folder.
2. From the scheme picker in Xcode, select the "Aztec" library.
3. Press CMD + U to run the unit tests for the library.
4. Switch to the scheme named "Example" (the example app).
5. Run the app.  Make sure everything is in place.


jleandroperez:
So happy to see CocoaPods vanishing....!!

:shipit:

SergioEstevao:
:shipit:

PR #70, opened 2016-09-20
Standard attachments

diegoreymendez:
This PR removes some of the custom attachment code we're using and tries to standardize on Apple's supported mechanism to handle attachments.

Replaces 451 lines of code with 206 - although the style and mechanisms can be improved a bit to be honest.

To test, just try out the attachment features we already had.


Aztec/Classes/Public/Extensions/NSAttributedString+Helpers.swift
```diff
@@ -2,9 +2,9 @@ import Foundation
 import UIKit
 
 
-/// NSTextStorage Helpers
+/// NSAttributedString Extension for Attachments
 ///
-extension NSTextStorage
+extension NSAttributedString
```
jleandroperez: Good call here!

Aztec/Classes/Public/TextKit/AztecTextAttachment.swift
```diff
@@ -40,22 +35,80 @@ public class AztecTextAttachment: NSTextAttachment
         super.init(coder: aDecoder)
     }
 
+    // MARK: - Origin calculation
+
+    func xPosition(forContainerWidth containerWidth: CGFloat) -> Int {
+
```
jleandroperez: Extra spacing?

Aztec/Classes/Public/TextKit/AztecTextStorage.swift
```diff
@@ -245,7 +261,7 @@ public class AztecTextStorage: NSTextStorage {
         return html
     }
 
-    public func setHTML(html: String, withDefaultFontDescriptor defaultFontDescriptor: UIFontDescriptor) {
+    func setHTML(html: String, withDefaultFontDescriptor defaultFontDescriptor: UIFontDescriptor) {
```
jleandroperez: Shouldn't this one remain public?

Aztec/Classes/Public/TextKit/TextView.swift
```diff
@@ -45,18 +40,6 @@ public class TextView: UITextView {
...
-    override public var bounds: CGRect {
-        didSet {
-            guard oldValue.size != bounds.size else {
-                return
-            }
-
-            attachmentManager.resizeAttachments()
-        }
-    }
-
```
jleandroperez: Super to see this go!

Aztec/Classes/Public/TextKit/AztecTextAttachment.swift
jleandroperez: I've noticed we've got this extra spacing ... all over. Is this something we want / our standard, or it's just by chance?

diegoreymendez: As far as I know we don't have standards on spacing, which IMHO is kind of a good thing as it means we can use it to increase readability.

Aztec/Classes/Public/TextKit/AztecTextStorage.swift
diegoreymendez: To be honest I don't think it should.  Our text storage is not designed to be used directly, but only through `Aztec.TextView`, which has the necessary interface for setting the HTML.

Aztec/Classes/Public/Extensions/NSAttributedString+Helpers.swift
diegoreymendez: Ty!

Aztec/Classes/Public/TextKit/AztecTextAttachment.swift
diegoreymendez: That said, I'll remove it from this particular method, since its not really needed.

diegoreymendez:
Ty @jleandroperez!

PR #71, opened 2016-09-20
Removes the Aztec prefix from several classes.

diegoreymendez:
Removes the Aztec prefix from several classes, since we already have the module prefix if we want to use it.

**How to test:**

1. Just make sure the app builds and the tests succeed.


jleandroperez:
Love this change!!

:shipit:

diegoreymendez:
Ty @jleandroperez!

PR #80, opened 2016-09-28
Adds unit tests, adds methods to aid in debugging.

diegoreymendez:
This PR adds some unit tests to avoid regressions with the latest link fixes.  It also adds some code to aid in debugging (Node inspection).

Also adds the concept of `leaf` nodes to group `TextNode` and `CommentNode`.

**How to test:**
1. Run the unit tests and make sure they work fine.
2. Try adding / removing links and make sure they work right.

### DIFF
--- a/Aztec/Classes/Private/Libxml2/Data/LeafNode.swift
+++ b/Aztec/Classes/Private/Libxml2/Data/LeafNode.swift
@@ -1,0 +1,10 @@
+import Foundation
+
+/// Represents any node that can be a leaf in the DOM tree.
+///
+protocol LeafNode {
+    
+    /// Returns the text representation of the node.
+    ///
+    func text() -> String
+}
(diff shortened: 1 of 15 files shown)


Aztec/Classes/Private/Libxml2/Data/LeafNode.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+/// Represents any node that can be a leaf in the DOM tree.
+///
+protocol LeafNode {
+    
+    /// Returns the text representation of the node.
+    ///
+    func text() -> String
```
diegoreymendez: I'm not a big fan of properties in protocols for read-access only... I feel like they make the most sense when write access is needed as well.

In this case, considering we only need read access, I find this:

```swift
func text() -> String
```

a bit clearer than this:

```swift
var text: String { get }
```

Does that make sense?

SergioEstevao: I see your point on the declaration side but when you are the using the property, it's more clear to just do element.text than element.text(). And has apple say it's better to make it easier on the side of using than on the side of declaring it, because you are probably going to use it a lot more.

diegoreymendez: PS: when I say `Node` relies on `func text() -> String` I mean it provides that method as something its subclasses should override.  By changing `text` to become a property I start getting name collisions due to the two different expectations from `Node` and `LeafNode`.

As I said - not very clean.  But maybe we can postpone fixing that a bit?

diegoreymendez:
Ty @SergioEstevao!

PR #81, opened 2016-09-28
Base work for copy / paste support.

diegoreymendez:
This PR makes some base changes for copy / paste support.

We're replacing:

```switf
override public func replaceCharactersInRange(range: NSRange, withString str: String)
```

with its NSAttributedString counterpart:

```
override public func replaceCharactersInRange(range: NSRange, withAttributedString attrString: NSAttributedString)
```

We're also adding support in the rich text view for pasting styled text (which still doesn't affect HTML).

**How to test:**
1. Try pasting styled text in the text view and make sure it works (doesn't update the DOM yet).
2. Edit regular text and make sure everything works as expected.

### DIFF
--- a/Aztec/Classes/Public/TextKit/TextStorage.swift
+++ b/Aztec/Classes/Public/TextKit/TextStorage.swift
@@ -87,6 +87,15 @@ public class TextStorage: NSTextStorage {
         return textStore.attributesAtIndex(location, effectiveRange: range)
     }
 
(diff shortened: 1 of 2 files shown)


diegoreymendez:
I'll be adding back the original method in a bit.

diegoreymendez:
I added the method back, so this is ready for another look.

SergioEstevao:
:shipit:

diegoreymendez:
Ty @SergioEstevao!

PR #82, opened 2016-09-30
Use enum instead of static arrays for element names and equivalents.

SergioEstevao:
Refactored the code to use an enum instead of static type arrays.

### DIFF
--- a/Aztec/Classes/Public/TextKit/TextStorage.swift
+++ b/Aztec/Classes/Public/TextKit/TextStorage.swift
@@ -14,6 +14,7 @@ public class TextStorage: NSTextStorage {
     typealias TextNode = Libxml2.TextNode
     typealias RootNode = Libxml2.RootNode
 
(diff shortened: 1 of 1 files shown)


Aztec/Classes/Public/TextKit/TextStorage.swift
```diff
@@ -14,29 +14,25 @@ public class TextStorage: NSTextStorage {
...
-    ///
-    private static let elementNamesForUnderline = ["u"]
+    enum ElementTypes: String {
+        case bold = "b"
+        case italic = "em"
+        case striketrough = "s"
+        case underline = "u"
+        case link = "a"
+
+        var equivalentNames: [String] {
```
diegoreymendez: Please add a description of why this exists, because even though it's fairly obvious to us, a newcomer could find this confusing.

SergioEstevao: @diegoreymendez  ready for another look.

PR #84, opened 2016-10-02
Add support for adding image to DOM.

SergioEstevao:
Add support for adding images to the HTML/DOM.

How to test:

 - Start the demo app
 - Go to editor demo
 - Add an image to the editor
 - Check the HTML view to see if the img tag was added
 - Go back and check if the image is still there
 - Remove the image on the visual mode
 - Check the HTML
 - Repeat the process but now remove image on the HTML mode.

### DIFF
--- a/Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
+++ b/Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
@@ -455,7 +455,7 @@ extension Libxml2 {
         }
 
         /// Returns the lowest-level text node in this node's hierarchy that wraps the specified
-        /// range.  If no child element node wraps the specified range, this method returns nil.
+        /// range.  If no child text node wraps the specified range, this method returns nil.
         ///
         /// - Parameters:
         ///     - range: the range we want to find the wrapping node of.
@@ -881,7 +881,31 @@ extension Libxml2 {
                 insert(string, atLocation: range.location)
             }
         }
-        
+
+        /// Replace characters in targetRange by a node with the name in nodeName and attributes
+        ///
+        /// - parameter targetRange: the range to replace
+        /// - parameter nodeName:    the name of the new node to create
+        /// - parameter attributes:  the attributes for the new node.
+        func replaceCharacters(inRange targetRange: NSRange, withNodeNamed nodeName: String, withAttributes attributes:[Attribute]) {
+            guard let textNode = lowestTextNodeWrapping(targetRange) else {
+                return
+            }
+            let absoluteLocation = textNode.absoluteLocation()
+            let localRange = NSRange(location:targetRange.location - absoluteLocation, length: targetRange.length)
+            textNode.split(forRange: localRange)
+            let imgNode = ElementNode(name: nodeName, attributes: attributes, children: [])
+            guard let index = textNode.parent?.children.indexOf(textNode) else {
+                assertionFailure("Can't remove a node that's not a child.")
+                return
+            }
+            guard let textNodeParent = textNode.parent else {
+                return
+            }
+            textNodeParent.insert(imgNode, at:index)
+            textNodeParent.remove(textNode)
+        }
+
         func split(atLocation location: Int) {
             
             guard location != 0 && location != length() - 1 else {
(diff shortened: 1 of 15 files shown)


diegoreymendez:
@SergioEstevao - As discussed in private, there are a few issues with inserting and deleting images in visual mode.

I recommend trying to merge [the copy/paste PR](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/85) since it provides some tools to make the fixes in this PR simpler.

diegoreymendez:
@SergioEstevao - Can you please check out the small improvements added in [this PR](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/86)?  They're targetted at this PR.

They don't solve all issues, but they do solve most of them.

diegoreymendez:
Still having trouble to (delete the images](https://cloudup.com/c6RRjoVPVmC).

I believe the real culprit here is [this line of code](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/84/files#diff-b08a98a1418f1c0ea7d9ae88558e1aaaR299).

Instead of wrapping the attachment character, we should be replacing it with a new image node.  Unfortunately we don't have methods yet for that (something like `replace(charactersInRange:withNewNodeNamed:withAttributes:`).

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -52,7 +54,7 @@ extension Libxml2 {
             /// Returns an array with all block-level elements.
             ///
             static func blockLevelNodeNames() -> [StandardName] {
-                return [.Address, .Blockquote, .Div, .Dl, .Fieldset, .Form, .H1, .H2, .H3, .H4, .H5, .H6, .Hr, .Noscript, .Ol, .P, .Pre, .Table, .Ul]
+                return [.Address, .Blockquote, .Div, .Dl, .Fieldset, .Form, .H1, .H2, .H3, .H4, .H5, .H6, .Hr, .Noscript, .Ol, .P, .Pre, .Table, .Ul, .Img]
```
diegoreymendez: Why did you add `.Img` here?  It's not a block-level node... maybe we need to change the logic of a method instead?

Let me know if I can help with this: but the bottom line is we must avoid tagging `.Img` as a block-level node.

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -449,6 +454,43 @@ extension Libxml2 {
             return self
         }
 
+        /// Returns the lowest-level text node in this node's hierarchy that wraps the specified
+        /// range.  If no child element node wraps the specified range, this method returns nil.
```
diegoreymendez: In this line "element" should be "text" probably.

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -449,6 +454,43 @@ extension Libxml2 {
...
+        ///
+        /// - Parameters:
+        ///     - range: the range we want to find the wrapping node of.
+        ///
+        /// - Returns: the lowest-level text node wrapping the specified range, or nil if
+        ///         no child node fulfills the condition.
+        ///
+        func lowestTextNodeWrapping(range: NSRange) -> TextNode? {
```
diegoreymendez: Could you add a few unit tests for this method?

I'd like to make sure any future changes won't break this.

Ideally:
- At least one for a good result.
- At least one for a nil result.

`ElementNodeTests.swift` has a lot of examples you can use.  Please let me know.

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -903,6 +949,21 @@ extension Libxml2 {
             }
         }
 
+        func replace(range targetRange: NSRange, withNodeNamed nodeName: String, withAttributes attributes:[Attribute]) {
```
diegoreymendez: Document the method, and if possible add a few unit tests.

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -903,6 +949,21 @@ extension Libxml2 {
...
+            guard let textNode = lowestTextNodeWrapping(targetRange) else {
```
diegoreymendez: The problem with this call is that it won't work for any range covering more than one node.  The ideal call here would be able to return more than one node for the specified range.

We can accept this for now, but would you mind creating a separate ticket to revisit this in the future?

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -903,6 +949,21 @@ extension Libxml2 {
...
+                return
+            }
+            let absoluteLocation = textNode.absoluteLocation()
+            let localRange = NSRange(location:targetRange.location - absoluteLocation, length: targetRange.length)
+            textNode.split(forRange: localRange)
+            let imgNode = ElementNode(name: nodeName, attributes: attributes, children: [])
+            guard let index = textNode.parent?.children.indexOf(textNode) else {
```
diegoreymendez: I'd add another guard before this one that does:

```swift
guard let textNodeParent = textNode.parent else {
    ...
}
```

So that we can avoid the optionals below.

Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
diegoreymendez: Extra info for the issue:

- The new method should also return the ranges so we can avoid using `absoluteLocation()`.  I'd base it on `lowestBlockLevelElements(...)`.
- The problem with `absoluteLocation()` is that it works fine only if you call this method in the rootNode.

diegoreymendez:
:shipit:

PR #86, opened 2016-10-04
Improvements for image insertion.

diegoreymendez:
This PR offers some modifications on top of the target PR.

Image insertion works slightly better - there are still some synch issues.

### DIFF
--- a/Aztec/Classes/Public/TextKit/TextView.swift
+++ b/Aztec/Classes/Public/TextKit/TextView.swift
@@ -418,9 +418,10 @@ public class TextView: UITextView {
         attachment.image = image
 
         // Inject the Attachment and Layout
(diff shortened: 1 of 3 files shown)


Aztec/Classes/Public/TextKit/TextView.swift
```diff
@@ -418,9 +418,10 @@ public class TextView: UITextView {
         attachment.image = image
 
         // Inject the Attachment and Layout
-        let insertionRange = NSMakeRange(position, 0)
+        let insertionRange = NSMakeRange(position, 1)
```
SergioEstevao: This make the image "overwrite" any characters that is next the insertion point.

SergioEstevao:
This make it work better but still some issues around, I will merge it anyway.

PR #88, opened 2016-10-04
Renames TextList.Kind > TextList.Style

jleandroperez:
#### Details:
Super quick rename. Figured out.. `Style` is more appropriate than `Kind`!

Please, verify that Aztec builds, and the unit tests run properly.

cc @diegoreymendez 
Thanks in advance!

Ref. #34


jleandroperez:
Thanks Diego!

PR #92, opened 2016-10-06
Add test for crash on newline after div

### LINKED ISSUE #90: Crash when adding a new line at the end of a div tag
If you tap return at the end of a div, Aztec will crash.

aztec-div-return-crash

koke:
Fixes #90, which was actually fixed by #84. This just adds a unit test

Needs Review: @diegoreymendez

### DIFF
--- a/AztecTests/HTML/ElementNodeTests.swift
+++ b/AztecTests/HTML/ElementNodeTests.swift
@@ -1376,7 +1376,16 @@ class ElementNodeTests: XCTestCase {
     }
 
     // MARK: - Bug fixes
-    // https://github.com/wordpress-mobile/WordPress-Aztec-iOS/issues/90
+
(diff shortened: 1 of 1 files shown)


AztecTests/HTML/ElementNodeTests.swift
```diff
@@ -1374,4 +1374,18 @@ class ElementNodeTests: XCTestCase {
         }
 
     }
+
+    // MARK: - Bug fixes
+    // https://github.com/wordpress-mobile/WordPress-Aztec-iOS/issues/90
+    func testInsertNewlineAfterDivShouldNotCrash() {
```
diegoreymendez: Could you document the test similarly to how other tests are documented?

It's a tool we're using to improve test maintenance and readability.  It also describes the logic so we can compare it against the actual test implementation.

diegoreymendez:
:shipit: Thanks!

Issue #75, opened 2016-09-20
Copy / Paste support.


diegoreymendez:
Add copy / paste support.

diegoreymendez:
Copy / paste is working for: bold, italic, underline, strikethrough.

I'll create separate issues for: links, images, blockquotes, lists.

PR #95, opened 2016-10-06
Copy / paste support.

diegoreymendez:
Implements copy / paste support.  Improves: #75.

Also fixes the node duplication issues we were having for styles.

**Known limitations:**

Still need to add support for pasting images.
Works for: bold, underline, italic, strike.
When copy-pasting large chunks of text, the editor becomes a bit slow.  [I've created a ticket for that](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/issues/97).

**Test 1:**

Evaluate the new unit tests.
Run the unit tests.

**Test 2:**

1. Launch the app.
2. Copy some content from the sample post.
3. Paste it back elsewhere.
4. Switch to HTML mode and see that the HTML is fine for the supported styles.

You can try pasting styled text from other apps too.


SergioEstevao:
Tests are passing, but I'm having some issues with small copy pastes inside the editor. 
For example:
 - copy a bit of the bold text, and paste inside the bold string. 
 - do the same but now with italic text

For me I get the editor stuck and/or duplication of tags (they are duplicated but they work and still valid html).

diegoreymendez:
I'll check those cases out, ty!

PR #99, opened 2016-10-07
Fixes some leaks and sends all DOM processing to a background thread.

diegoreymendez:
Title has it.

**How to test:**

Just go crazy with the editor.  Make sure to change to HTML mode frequently since that triggers a read operation on the DOM (all other are write operations to the DOM).
Also run al unit tests.


diegoreymendez:
Closing since the diff is off for some reason.

PR #111, opened 2016-10-11
Fixes a bug with the layout code.

### LINKED ISSUE #110: Image layout problem
There's a problem in when we're invalidating the layout when adding images.  This causes the app to sometimes show errors in the log, and sometimes crash.

**How to reproduce:**

Repeat the exact steps from [this video](https://cloudup.com/cm1kzHtcWnQ).

diegoreymendez:
Fixes #110.

Also reduces the amount of code in `setHTML(...)` that is executed inside the dom queue.

**Test 1:**

Repeat the exact steps from [this video](try/fix-image-loading-problem) and make sure the app doesn't crash.

**Test 2:**

Try adding images normally and make sure they're still showing up fine.


diegoreymendez:
Ty!

PR #104, opened 2016-10-11
Fix insertion of characters between images.

SergioEstevao:
How to test:
 - Put the cursor at the beginning of the editor
 - Insert one image by using the media picker
 - Insert another image by using the media picker
 - Put the cursor between images and try to write
 - No crash should happen
 - Run the unit tests just to check if everything is alright.

PR #113, opened 2016-10-11
Fixes a crash when editing between img nodes.

### LINKED ISSUE #112: The editor is crashing when editing between two image nodes.
When trying to edit in visual mode between images, the app crashes.

See [this video](https://cloudup.com/dashboard) showing the crash.

Props to @SergioEstevao for finding this one.

diegoreymendez:
Fixes #112.

Supersedes #104.

**Information:**

When we introduced [this change](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/96/commits/3a9d89b8ac65d97aa8077d2c7aad524a5e2e5e07) we lost some of the logic in the process.  The change was fine, but there was an unintentional change in the logic too.

I reintroduced the lost logic in this PR.

I renamed the condition check parameter to make it clear what it's supposed to do.

**How to test:**

Add two images in HTML mode.
Switch to visual mode, and try adding a character between them.
Make sure the app doesn't crash, and editing works.

### DIFF
--- a/Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
+++ b/Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
@@ -367,7 +367,9 @@ extension Libxml2 {
                     
                     let intersectionInChildCoordinates = NSRange(location: intersection.location - offset, length: intersection.length)
 
-                    if let childElement = child as? ElementNode {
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Ty Sergio!

PR #114, opened 2016-10-12
Refactor image attachment

SergioEstevao:
Just a first stab on implementing loading of attachment directly on the attachment instead of storage.

This PR is still not complete it's just a first refactor in order to evaluate if it's a good idea.

Still needs to implement code for image with missing URL reference.

@diegoreymendez do you mind to have a look at the code.

### DIFF
--- a/Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
+++ b/Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
@@ -24,24 +24,22 @@ extension NSAttributedString
     }
 
     /// Determine the character ranges for an attachment
+    ///
+    /// - parameter attachment: the attachment to search for
+    ///
+    /// - returns: an array of ranges where the attachement can be found
+    ///
     public func ranges(forAttachment attachment: NSTextAttachment) -> [NSRange]
     {
-        // find character range for this attachment
         let range = NSRange(location: 0, length: length)
-
-        var refreshRanges = [NSRange]()
-
+        var attachmentRanges = [NSRange]()
         enumerateAttribute(NSAttachmentAttributeName, inRange: range, options: []) { (value, effectiveRange, nil) in
-
-            guard let foundAttachment = value as? NSTextAttachment where foundAttachment == attachment else
-            {
+            guard let foundAttachment = value as? NSTextAttachment where foundAttachment == attachment else {
                 return
             }
-
-            // add this range to the refresh ranges
-            refreshRanges.append(effectiveRange)
+            attachmentRanges.append(effectiveRange)
         }
         
-        return refreshRanges
+        return attachmentRanges
     }
 }
--- a/Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
+++ b/Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
@@ -22,4 +22,26 @@ extension NSAttributedString
             }
         }
     }
+
+    /// Determine the character ranges for an attachment
+    public func ranges(forAttachment attachment: NSTextAttachment) -> [NSRange]
+    {
+        // find character range for this attachment
+        let range = NSRange(location: 0, length: length)
+
+        var refreshRanges = [NSRange]()
+
+        enumerateAttribute(NSAttachmentAttributeName, inRange: range, options: []) { (value, effectiveRange, nil) in
+
+            guard let foundAttachment = value as? NSTextAttachment where foundAttachment == attachment else
+            {
+                return
+            }
+
+            // add this range to the refresh ranges
+            refreshRanges.append(effectiveRange)
+        }
+        
+        return refreshRanges
+    }
 }
--- a/Aztec/Classes/Public/Extensions/NSLayoutManager+Attachments.swift
+++ b/Aztec/Classes/Public/Extensions/NSLayoutManager+Attachments.swift
@@ -2,37 +2,12 @@ import UIKit
 
 extension NSLayoutManager
 {
-    /// Determine the character ranges for an attachment
-    public func ranges(forAttachment attachment: NSTextAttachment) -> [NSRange]
-    {
-        guard let attributedString = self.textStorage else
-        {
-            return []
-        }
-
-        // find character range for this attachment
-        let range = NSRange(location: 0, length: attributedString.length)
-
-        var refreshRanges = [NSRange]()
-
-        attributedString.enumerateAttribute(NSAttachmentAttributeName, inRange: range, options: []) { (value, effectiveRange, nil) in
-
-            guard let foundAttachment = value as? NSTextAttachment where foundAttachment == attachment else
-            {
-                return
-            }
-
-            // add this range to the refresh ranges
-            refreshRanges.append(effectiveRange)
-        }
-
-        return refreshRanges
-    }
-
     /// Invalidates the layout for an attachment when some change happened to it.
     public func invalidateLayoutForAttachment(attachment: NSTextAttachment) {
-
-        for range in ranges(forAttachment: attachment) {
+        guard let ranges = self.textStorage?.ranges(forAttachment: attachment) else {
+            return
+        }
+        for range in ranges {
             invalidateLayoutForCharacterRange(range, actualCharacterRange: nil)
             invalidateDisplayForCharacterRange(range)
         }
--- a/Aztec/Classes/Public/Extensions/NSLayoutManager+Attachments.swift
+++ b/Aztec/Classes/Public/Extensions/NSLayoutManager+Attachments.swift
@@ -1,0 +1,40 @@
+import UIKit
+
+extension NSLayoutManager
+{
+    /// Determine the character ranges for an attachment
+    public func ranges(forAttachment attachment: NSTextAttachment) -> [NSRange]
+    {
+        guard let attributedString = self.textStorage else
+        {
+            return []
+        }
+
+        // find character range for this attachment
+        let range = NSRange(location: 0, length: attributedString.length)
+
+        var refreshRanges = [NSRange]()
+
+        attributedString.enumerateAttribute(NSAttachmentAttributeName, inRange: range, options: []) { (value, effectiveRange, nil) in
+
+            guard let foundAttachment = value as? NSTextAttachment where foundAttachment == attachment else
+            {
+                return
+            }
+
+            // add this range to the refresh ranges
+            refreshRanges.append(effectiveRange)
+        }
+
+        return refreshRanges
+    }
+
+    /// Invalidates the layout for an attachment when some change happened to it.
+    public func invalidateLayoutForAttachment(attachment: NSTextAttachment) {
+
+        for range in ranges(forAttachment: attachment) {
+            invalidateLayoutForCharacterRange(range, actualCharacterRange: nil)
+            invalidateDisplayForCharacterRange(range)
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


Aztec/Classes/Public/Extensions/NSLayoutManager+Attachments.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
+
+extension NSLayoutManager
+{
+    /// Determine the character ranges for an attachment
+    public func ranges(forAttachment attachment: NSTextAttachment) -> [NSRange]
```
diegoreymendez: This method looks like a better fit for `NSAttributedString+Attachments.string`, since it only interacts with the attributed string.

`invalidateLayoutForAttachment()` could access this method through the `textStorage`.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -106,7 +123,13 @@ public class TextAttachment: NSTextAttachment
     /// Otherwise, we'll always take the whole container's width.
     ///
     override public func attachmentBoundsForTextContainer(textContainer: NSTextContainer?, proposedLineFragment lineFrag: CGRect, glyphPosition position: CGPoint, characterIndex charIndex: Int) -> CGRect {
-
+        self.textContainer = textContainer
+        switch kind {
+        case .RemoteImage:
+            startImageFetching()
+        default:
+            print("No action to do")
```
diegoreymendez: No need to log this (as there's not much value in the log message).  Add a comment if we need to document it's ok if nothing happens.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -79,6 +89,13 @@ public class TextAttachment: NSTextAttachment
     // MARK: - NSTextAttachmentContainer
 
     override public func imageForBounds(imageBounds: CGRect, textContainer: NSTextContainer?, characterIndex charIndex: Int) -> UIImage? {
+        self.textContainer = textContainer
+        switch kind {
+        case .RemoteImage:
+            startImageFetching()
+        default:
+            print("No action to do")
```
diegoreymendez: No need to log this (as there's not much value in the log message).  Add a comment if we need to document it's ok if nothing happens.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +139,39 @@ public class TextAttachment: NSTextAttachment
 
         return CGRect(origin: CGPointZero, size: CGSize(width: width, height: onScreenHeight(width)))
     }
+
+    func startImageFetching() {
+        var remoteURL: NSURL?
```
diegoreymendez: You can change this to `let remoteURL: NSURL` and it'll work too.

Also consider renaming the variable since the URL may not be "remote" in some cases.

The advantage is we won't have an optional here.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +139,39 @@ public class TextAttachment: NSTextAttachment
...
+        switch kind {
+        case .RemoteImage(let url):
+            remoteURL = url
+        default:
+            remoteURL = nil
+        }
+
+        guard let imageURL = remoteURL else {
```
diegoreymendez: This can be removed with the change suggested above.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +139,39 @@ public class TextAttachment: NSTextAttachment
...
+            return
+        }
+        if fetchingHappening {
```
diegoreymendez: Consider moving this to the top of this method and making it a `guard`.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -23,9 +26,16 @@ public class TextAttachment: NSTextAttachment
 
     private var glyphImage: UIImage?
 
-    /// Designed Initializer
+    public var imageProvider: TextAttachmentImageProvider?
+    public var fetchingHappening: Bool = false
```
diegoreymendez: Maybe change this to `isFetchingImage` or something similar?

`fetchingHappening` sounds a bit non-standard.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +139,39 @@ public class TextAttachment: NSTextAttachment
 
         return CGRect(origin: CGPointZero, size: CGSize(width: width, height: onScreenHeight(width)))
     }
+
+    func startImageFetching() {
```
diegoreymendez: Maybe rename to `startFetchingImage()` for improved readability?

SergioEstevao:

Aztec/Classes/Public/Extensions/NSLayoutManager+Attachments.swift
```diff
@@ -0,0 +1,15 @@
+import UIKit
+
+extension NSLayoutManager
+{
+    /// Invalidates the layout for an attachment when some change happened to it.
+    public func invalidateLayoutForAttachment(attachment: NSTextAttachment) {
+        guard let ranges = self.textStorage?.ranges(forAttachment: attachment) else {
```
diegoreymendez: You can remove `self.` here.

Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
```diff
@@ -22,4 +22,26 @@ extension NSAttributedString
             }
         }
     }
+
+    /// Determine the character ranges for an attachment
+    public func ranges(forAttachment attachment: NSTextAttachment) -> [NSRange]
+    {
+        // find character range for this attachment
```
diegoreymendez: IMHO we can remove this comment - the method documentation should state what we're doing in here anyway.

Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
```diff
@@ -22,4 +22,26 @@ extension NSAttributedString
...
+        let range = NSRange(location: 0, length: length)
+
+        var refreshRanges = [NSRange]()
```
diegoreymendez: I think we should rename this variable, since this method is not really tied to refreshing.

Aztec/Classes/Public/Extensions/NSAttributedString+Attachments.swift
```diff
@@ -22,4 +22,26 @@ extension NSAttributedString
...
+
+        enumerateAttribute(NSAttachmentAttributeName, inRange: range, options: []) { (value, effectiveRange, nil) in
+
+            guard let foundAttachment = value as? NSTextAttachment where foundAttachment == attachment else
+            {
+                return
+            }
+
+            // add this range to the refresh ranges
```
diegoreymendez: The line below is actually very clear, so we can safely remove this comment too.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -79,6 +89,8 @@ public class TextAttachment: NSTextAttachment
     // MARK: - NSTextAttachmentContainer
 
     override public func imageForBounds(imageBounds: CGRect, textContainer: NSTextContainer?, characterIndex charIndex: Int) -> UIImage? {
+        self.textContainer = textContainer
+        updateImage()
```
diegoreymendez: Very nicely wrapped.

Aztec/Classes/Public/TextKit/TextStorage.swift
```diff
@@ -462,64 +462,24 @@ public class TextStorage: NSTextStorage {
...
 
-    private func downloadFailure(attachment: TextAttachment) {
-        guard let imageProvider = imageProvider else {
-            fatalError("The image provider should've been set at this point.")
-        }
+extension TextStorage: TextAttachmentImageProvider {
 
-        attachment.image = imageProvider.storage(self, missingImageForAttachment: attachment)
-        invalidateLayoutForAttachment(attachment)
+    public func image(forURL url: NSURL, inAttachment attachment: TextAttachment,
```
diegoreymendez: The `attachment` parameter is on the same line as the `url` parameter, while the other two are in new lines.

Aztec/Classes/Public/TextKit/TextView.swift
```diff
@@ -783,13 +783,13 @@ public class TextView: UITextView {
     public func changeAlignment(forAttachment attachment: TextAttachment, to alignment: TextAttachment.Alignment) {
         attachment.alignment = alignment
 
-        storage.invalidateLayoutForAttachment(attachment)
+        layoutManager.invalidateLayoutForAttachment(attachment)
```
diegoreymendez: Very nice change.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +129,43 @@ public class TextAttachment: NSTextAttachment
 
         return CGRect(origin: CGPointZero, size: CGSize(width: width, height: onScreenHeight(width)))
     }
+
+    func updateImage() {
+        switch kind {
```
diegoreymendez: This switch block can be removed safely, now.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -1,6 +1,9 @@
 import Foundation
 import UIKit
 
+public protocol TextAttachmentImageProvider {
+    func image(forURL url: NSURL, inAttachment attachment: TextAttachment, onSuccess success: (UIImage) -> (), onFailure failure: () -> ()) -> UIImage?
```
diegoreymendez: I'd suggest renaming `inAttachment` -> `forAttachment`.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +129,43 @@ public class TextAttachment: NSTextAttachment
...
     }
+
+    func updateImage() {
+        switch kind {
+        case .RemoteImage:
+            break
+        default:
+            return
+        }
+        guard !isFetchingImage else {
```
diegoreymendez: I'd consider adding `let imageProvider = imageProvider` to this guard, since none of the code below makes sense without a delegate.

Aztec/Classes/Public/TextKit/TextAttachment.swift
```diff
@@ -116,6 +129,43 @@ public class TextAttachment: NSTextAttachment
...
+                                                    self?.triggerUpdate()
+            }, onFailure: { [weak self]() in
+                self?.isFetchingImage = false
+                self?.kind = .MissingImage
+                self?.triggerUpdate()
+            })        
+    }
+
+    func triggerUpdate(){
+        self.textContainer?.layoutManager?.invalidateLayoutForAttachment(self)
```
diegoreymendez: IMHO the text attachment should not know the layout manager at all.

This should be executed automatically by our `TextStorage` [here](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/114/files#diff-b08a98a1418f1c0ea7d9ae88558e1aaaR478).

Then [this property](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/114/files#diff-ac4b458a2b7aba8c3f8e9aad368bae1dR31) would also go away.

SergioEstevao:
@diegoreymendez: Ready for another round of review.

Aztec/Classes/Public/TextKit/TextAttachment.swift
diegoreymendez: We can discuss it separately from this PR.

I think the problem is that this is not necessarily a task for an "image provider" but it certainly is a task for our `TextStorage` to take care of.

PR #117, opened 2016-10-13
Lists: Matching Calypso's Behavior

jleandroperez:
### Details:
I've updated Issue #34's **Toolbar** section's Test Cases. You can find Unit Tests for each one of those in `TextListFormatterTests` (each test has a comment referencing to the matching scenario).

Lists should, effectively, behave the same was as Calypso does, as of now.

### Testing:
- Verify that the Unit Tests run correctly
- Manually verify Scenarios 1-7
- Be happy!!

Needs Review: @diegoreymendez 
cc @aerych

### DIFF
--- a/Aztec/Classes/Public/Extensions/NSAttributedString+Lists.swift
+++ b/Aztec/Classes/Public/Extensions/NSAttributedString+Lists.swift
@@ -13,7 +13,7 @@ extension NSAttributedString
     ///     - ranges: Ranges to be filtered
     ///     - style: Style to be matched
     ///
-    /// - Returns: Returns: A subset of the input ranges that don't contain TextLists matching the input style.
+    /// - Returns: A subset of the input ranges that don't contain TextLists matching the input style.
     ///
     func filterListRanges(ranges: [NSRange], notMatchingStyle style: TextList.Style) -> [NSRange] {
         return ranges.filter { range in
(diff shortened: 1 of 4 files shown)


Aztec/Classes/Public/Extensions/NSAttributedString+Lists.swift
```diff
@@ -6,6 +6,23 @@ import UIKit
...
 extension NSAttributedString
 {
+    /// Given a collection of NSRange's, this method will filter all of those who contain a TextList, and
+    /// doesn't match the specified Style.
+    ///
+    /// - Parameters:
+    ///     - ranges: Ranges to be filtered
+    ///     - style: Style to be matched
+    ///
+    /// - Returns: A collection of NSRange's, guarranteed to either *NOT* contain TextLists, or contain TextLists
```
diegoreymendez: Suggestion to increase clarity:

`Returns: A subset of the input ranges that don't contain TextLists matching the input style.`

The description applies for both possible scenarios and is slightly easier to read.

Aztec/Classes/Public/Extensions/NSRange+Helpers.swift
```diff
@@ -34,6 +34,12 @@ extension NSRange
         }
     }
 
+    /// Returns the maximum Location.
+    ///
+    var lastLocation: Int {
```
diegoreymendez: This is fairly minor: but I'd replace `last` with `end` since the term is used in other places in our code.

It's also more frequently used in Apple's code.

jleandroperez:
@diegoreymendez thanks a lot for the review!!. Feedback addressed, ready for another peek!

PR #118, opened 2016-10-14
Lists: Updating SelectedRange Accurately

jleandroperez:
### Details:

In this PR we're introducing several tweaks, so that:
- Whenever a TextList is inserted over a line (with _no selected text_), the cursor is displaced to the right, to account for the **List Item Marker**'s length
- Whenever a TextList is applied over a **Selected Text Range**, the selection is adjusted to account for the List's Markers.
### Scenario A: No Selection
1. Tap over any line
2. Press the (Ordered / Unordered) toolbar button
3. Verify that the cursor shifts to the right, and remains at the same relative position
4. Press the (Ordered / Unordered) toolbar button again
5. Verify that the list gets removed, and the cursor is adjusted again.
### Scenario B: Selection
1. Highlight a group of paragraphs
2. Press the (Ordered / Unordered) toolbar button
3. Verify that the **selected text range** gets adjusted to the effective List Length.
4. Press the (Ordered / Unordered) toolbar button again
5. Verify that the **selected text range** gets adjusted again.

Please, try several variants (such as selecting a range of text that contains a List Item at its middle, of the same and different list kind).

Ref. #34

Diego, i'll DM you regarding Unit Tests. Could really use few hints there, tricky one to unit test!

cc @diegoreymendez


diegoreymendez:
@jleandroperez - Some caret positioning issues here: https://cloudup.com/c55BL-BC7UO

I'm also seeing some really weird caret movement when selecting large ranges of text.

I'm not sure if I understand scenario B properly, but the selection should remain the same before and after applying styles (even when the style applies to a larger text-run).

Here's an example of the expected behaviour: https://cloudup.com/cDtGp90ckqU

SergioEstevao:
@diegoreymendez I did some changes on the PR to make the selection always be kept.

What I did was:
 -  create two methods that allow you to mark and restore the current selection with an special attribute
  - then I apply whatever changes are done by lists, and then restore the selection that should have moved with the changes done by lists

Have a look and tell me what you think.

diegoreymendez:
@SergioEstevao - Working great now!

PR #122, opened 2016-10-18
Refactor HTML related enums to single type outside of element node.

SergioEstevao:
This PR refactors code related to HTML elements, consolidating all the html element related code in a single enum.

How to test:

 - Build the app
 - Run the tests
 - Run the demo app and see if behaviour is still the same.


Aztec/Classes/Private/Libxml2/Data/ElementNode.swift
```diff
@@ -93,9 +20,9 @@ extension Libxml2 {
         private(set) var attributes = [Attribute]()
         private(set) var children: [Node]
 
-        private var standardName: StandardName? {
+        private var standardName: HTMLElementType? {
```
diegoreymendez: We've lost clarity here given the name has no relation to the type.  It'd be much cleaner here if we had:

```
private var standardType: StandardElementType? {
```

The word "standard" would give a hint as to why this property can be `nil` (ie: if the type is not standard, it will be `nil` here).

diegoreymendez:
Nice work @SergioEstevao!

PR #124, opened 2016-10-21
Fixes a crashing bug when inserting images.

diegoreymendez:
Inserting images was broken as a result of [this PR](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/123).

This PR fixes it.

**How to test:**

Just insert an image and make sure the app doesn't crash.


SergioEstevao:
:shipit:

Issue #76, opened 2016-09-27
Blockquote support.


diegoreymendez:
Add blockquote support to the editor.

Task list:
- [x] #143 Improve caret calculation
- [x] #144 Caret disappears on new line
- [x] #145 Custom attributes removed on new line
- [ ] DOM implementation

diegoreymendez:
@koke - Will work on initial support for this feature in Iteration 7.  We do expect this feature could require additional time.

PR #125, opened 2016-10-21
Adds Blockquote formatter

koke:
Refactors the code a bit to move all the blockquote formatting to a dedicated formatter, like we're doing for lists. I've added a custom `Blockquote` attribute to mark blockquotes instead of relying on specific paragraph styles. I've also tweaked the style a bit to help debugging, but definitely not the final style.

Refs #76 
Needs Review: @diegoreymendez


diegoreymendez:
@koke - Let me know when this is ready for another view.

koke:
@diegoreymendez I think this is ready for another round. I added the `AttributeFormatter` from #129, but as I was writing the documentation I realized I was handling the typing attributes wrong, and the only difference between a character and paragraph attribute formatter is the application range, and I had to refactor a bit.

Handling of typing attributes is a work in progress, since Apple seems to be whitelisting them and custom attributes disappear as soon as you start typing. But I'm leaving that for another PR.

Aztec/Classes/Public/Formatters/AttributeFormatter.swift
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    func attribute(inString string: NSAttributedString, at index: Int) -> Bool {
+        let attributes = string.attributesAtIndex(index, effectiveRange: nil)
+        return present(inAttributes: attributes)
+    }
+
+    func insertEmptyAttribute(inString string: NSMutableAttributedString, at index: Int) {
+        let attributedSpace = NSAttributedString(string: "\u{200B}", attributes: attributes)
```
diegoreymendez: I'd suggest making this a constant.  I'm not sure what the definitive location for that constand would be, but a good starting place could be as a member of this class (`AttributeFormatter`).

Aztec/Classes/Public/Formatters/AttributeFormatter.swift
```diff
@@ -0,0 +1,148 @@
...
+
+/// A type that provides support for toggling compound attributes in an
+/// attributed string.
+///
+/// When you want to represent an attribute that does not have a 1-1
+/// correspondence with a standard attribute, it is useful to have a virtual
+/// attribute. Toggling this attribute would also toggle the attributes for its
+/// defined style.
+///
+protocol AttributeFormatter {
```
diegoreymendez: I think it would be great to have unit tests for the default method implementations in this file.

If you'd like to push things forwards I'm perfectly fine with handling that in a separate PR though.  Especially since there's some indirect testing through `BlockquteFormatterTests`.

Aztec/Classes/Public/Formatters/AttributeFormatter.swift
koke: 👍 It will probably move somewhere else in the future, but I'll add it there for now

Aztec/Classes/Public/Formatters/AttributeFormatter.swift
```diff
@@ -56,6 +56,12 @@ extension AttributeFormatter {
     func applicationRange(forRange range: NSRange, inString string: NSAttributedString) -> NSRange {
         return range
     }
+
+    /// The string to be used when adding attributes to an empty line.
+    ///
+    var placeholderForAttributedEmptyLine: String {
```
diegoreymendez: Why not `let placeholderForAttributedEmptyLine = "\u{200B}"`?

diegoreymendez: Or even `static`, if you prefer.

koke: Because `Extensions may not contain stored properties`, and `Static stored properties not yet supported in generic types` 😉

Aztec/Classes/Public/Formatters/AttributeFormatter.swift
diegoreymendez: Well, we agree.  My point was more about it being fine to tackle those in an upcoming PR.  But your call in any case.

Aztec/Classes/Public/Formatters/AttributeFormatter.swift
diegoreymendez: Then this should do for now.

PR #130, opened 2016-10-24
Introduces the DOM object.

diegoreymendez:
## **Description:**

Introduces a DOM object to represent the HTML document object model.

## **Details:**

The DOM object represents the HTML document object model to the eyes of the rest of the code.  `DocumentObjectModel` is the class that will handle all interaction with the HTML structure.

It also takes care of managing the HTML editing queue (which was previously handled by `TextStorage`).

It also improves styles-handling so that styles don't need to be set explicitly anymore, other than through these methods which replicate `NSTextStorage`'s default editing methods (and in fact must be forwarded directly).

```swift
func replaceCharacters(inRange range: NSRange, withString string: String, inheritStyle: Bool)
func replaceCharacters(inRange range: NSRange, withAttributedString attributedString: NSAttributedString, inheritStyle: Bool)
func setAttributes(attrs: [String : AnyObject]?, range: NSRange)
```

## **How to test:**

**Test 1:**

Run the unit tests.

**Test 2:** 

Launch the editor.
Edit styles by copy-pasting.
Make sure all works like 

**Test 3:** 

Launch the editor.
Edit styles by using the toolbar.
Make sure all works like


diegoreymendez:
Ty @SergioEstevao !

PR #135, opened 2016-10-28
Adds custom drawing of background for blockquotes.

koke:
Includes a new custom NSLayoutManager, plus some tweaks for the caret size.
When you have a line with non-default paragraph spacing, UIKit will have super
tall carets that fill the line height. The custom caret isn't exactly like the
default, but it's close enough (at least until we figure out a better approach).

This also adds a `CGRect.resize` helper, so you can resize CGRects anchoring to
other than the top left.

Refs #76 
Needs Review: @diegoreymendez


Aztec/Classes/Public/Extensions/CGRect+Helpers.swift
```diff
@@ -0,0 +1,56 @@
...
+    enum HorizontalAnchor {
+        case left
+        case center
+        case right
+    }
+
+    /// Return a rect resized to a specific size, while anchoring it to one of
+    /// their edges or centers.
+    ///
+    /// By default, changing a CGRect size will resize it anchored to its top
```
diegoreymendez: I was a bit confused by the wording here as it felt like the resize operation is more than just an assignment:

```By default, changing a CGRect size will resize it anchored to its top left corner left corner, as a CGRect is defined by its `origin`.```

I think something like this could describe the situation better:

```By default, CGRect does not support anchoring, so modifying the size will not affect the origin in any way.```

koke: Yeah, I struggled with the description there. Your description reads easier, but it's also not completely accurate. I wouldn't say that "CGRect does not support anchoring", since it's just a dumb struct. Like you said, setting the size is just an assignment. Let's try another iteration:

> CGRect is typically used in UIKit so the origin is in the upper-left corner and the rectangle extends towards the lower-right corner. This method adds the option to grow or shrink the rect   anchoring to other points.

BTW, I added the UIKit mention since I read the CGRect documentation and found this:

> In the default Core Graphics coordinate space, the origin is located in the lower-left corner of the rectangle and the rectangle extends towards the upper-right corner. If the context has a flipped-coordinate space—often the case on iOS—the origin is in the upper-left corner and the rectangle extends towards the lower-right corner.

To be completely accurate, we should ditch the left/right top/bottom enum values and be context agnostic, but I think we can let it go for this project. Maybe just add a note that the method assumes a upper-left-origin coordinate system.

Thoughts?

diegoreymendez: > CGRect is typically used in UIKit so the origin is in the upper-left corner and the rectangle extends towards the lower-right corner. This method adds the option to grow or shrink the rect anchoring to other points.

I think the first part can be removed completely, since it talks about something that's really tangential to the scope of this method.  In other words, this is an extension to (a dumb struct named) CGRect adding some support for resizing + anchoring.  This support is offered in a really broad and generic way, non specific to UIKit.

I would maintain (a slightly modified version of) the second part and make it just:

> This method adds the option to grow or shrink the rect while anchoring it to a point.

diegoreymendez: In other words, I think the description should explain what the method does, and not why we particularly needed it in the first place.

koke: > This support is offered in a really broad and generic way, non specific to UIKit

I agree that the explanation is perhaps too detailed, but this is not generic as long as the horizontal/vertical enum values mention left/right and top/bottom. CoreGraphics actually has the origin at the bottom left corner, so if someone used this method within that context, the vertical enum values would mean the opposite of what they are.

I think it's safe to assume that will not happen within the scope of this project, but that's why I added the note about UIKit. If you think it adds more confusion I can remove it.

koke:
@diegoreymendez Fixed the issue and got rid of the CGRect extension in the process. Ready for another round

koke:
> The caret to disappear

I've noticed that in the past but haven't figured out what it is yet. It's not related to the custom drawing though, as you can see the same effect if you use a simple background color.

> Blockquoted text to be left outside of the blockquote

I think this also happened before, only it's more apparent with a custom background. I have the feeling that this is going to be complicated enough to need its own PR.

koke:
I've reported the missing caret: http://openradar.appspot.com/radar?id=4998031044771840

koke:
> Blockquoted text to be left outside of the blockquote.

This seems to be a similar issue as the whitelisted typing attributes. When you type return at the end of a blockquote, TextKit drops the custom attribute but preserves the paragraph style.

SergioEstevao:
:shipit: , please create issues for the the things you found that are not right yet.

PR #138, opened 2016-11-04
Make list markers non selectable

SergioEstevao:
Refs #34 

Implements the cursor requirements:

> If the current selection intersects a list marker, adjust the selection so it does not intersect the marker. Move the cursor forward or backward depending on the previous cursor location.

Note: It still allows deletion of part of the markers that will be tackled next.


SergioEstevao:
@diegoreymendez you are right, the first issue needs to be solved here the second issue has I told in the note of the PR it's still to be tackled. FIY the markers at the moment are a bullet/number followed by a tab.

SergioEstevao:
@diegoreymendez ready for another look. Issue 1 sorted out.

PR #139, opened 2016-11-04
Specify the encoding to the HTML parser

koke:
I discovered an issue with the converter not handling non-ASCII characters properly.

To test:
- Change `content.html` to have some non-ASCII text
- Launch the editor demo and verify it shows correctly

Needs Review: @diegoreymendez


diegoreymendez:
@koke - Remember to set the label, milestone and assignees, as I use those both for tracking my pending PRs and for preparing the sprint wrap-up posts.

PR #141, opened 2016-11-07
Deletion of list markers

SergioEstevao:
Ref #34 

Make sure than when deleting list markers they are delete as an all and not partially.

How to test:

- Create a list ordered/unordered
- Place the cursor just after a marker ( beginning of paragraph)
- Tap on delete/backward 

Check if the markers is deleted has all.

This PR still not upgrades the numbering of the follow up list paragraphs.


diegoreymendez:
@SergioEstevao - Working mostly fine.  I found [this issue](https://cloudup.com/cdL73y7xa2A) though, when trying to add a list item back.

SergioEstevao:
@diegoreymendez that is expected, because I'm still not updating/removing the attributes for the list, so when you tap on the list option you first remove the list attributes and then you are reapplying it.

PR #142, opened 2016-11-07
Fix cursor caret on end of text

SergioEstevao:
The cursor caret was disappearing when it was moved to the end of the text on a text view. This PR fix this issue.

How to text:
- Open the demo app
- Select the editor demo
- Move the cursor to the end of the text
- Check if the cursor is visible and on the right position.


koke:
Good catch, but I think unfortunately this means the line fragment method wasn't the right way to calculate this. When you move the cursor to the end of a line, you can see this warning in the console:

```2016-11-08 09:52:13.538 AztecExample[78151:18366212] !!! _NSLayoutTreeLineFragmentRectForGlyphAtIndex invalid glyph index 1```

Which makes sense, since there is no glyph at the end of the text, and so the layout manager can't figure out its rect.

I guess :shipit: so it doesn't break so bad for now, but I'll look into a better solution.

Issue #147, opened 2016-11-09
Lists - Backspace deleted


SergioEstevao:
When a backspace would intersect a list marker, delete the entire list marker + the carriage return of the preceding list item (if any). If the current list item still had text, merge the remaining text with the previous list item. Update the numbering for any following items.

diegoreymendez:
Closing this as the related PR was merged in.

Issue #146, opened 2016-11-09
Lists - Carriage Return


SergioEstevao:
When there is a carriage return inside a list. Insert a new list marker. Update the numbering of any following items.
 When the cursor is on an empty list item (just the marker and no other text) and a carriage return is added. Remove the empty list item, and insert a normal paragraph in its place. If there are any following list items, replace those items with a new list and new item numbering.

PR #148, opened 2016-11-10
Update of markers when editing text

SergioEstevao:
Refs #146 and #147 

This PR implement support for updating/removing lists while editing text inside the list, specially when deleting around markers and entering new lines.

How to test:

- Start the editor demo
- Create a new ordered list spreading across several lines
- Try the following:
  - Break lines in middle and see if the markers are update correctly
  - Try Delete marker using the delete key or cutting a selection and see if markers are updated correctly
 - Enter two new empty lines in the end of the list and see if the list is interrupted.
  - Enter two new empty new lines in a go in the middle of a list and see if the list is split int half

### DIFF
--- a/Aztec/Classes/Formatters/TextListFormatter.swift
+++ b/Aztec/Classes/Formatters/TextListFormatter.swift
@@ -37,6 +37,35 @@ class TextListFormatter
         let listParagraphs = string.paragraphRanges(spanningRange: listRange)
         return applyList(ofStyle: style, toString: string, atRanges: listParagraphs)
     }
+
+    func refreshList(inString string: NSMutableAttributedString, atRange range: NSRange) -> NSRange? {
+
+        var styleOptional: TextList.Style?
+        // Load Paragraph Ranges
+
+        let paragraphRanges = string.paragraphRanges(spanningRange: range)
+        guard let firstParagraphRange = paragraphRanges.first else {
+            if let paragraphRange = string.rangeOfLine(atIndex: range.location) {
+                return removeList(fromString: string, atRanges: [paragraphRange])
+            }
+            return nil
+        }
+        if let textList = string.attribute(TextList.attributeName, atIndex: firstParagraphRange.location, effectiveRange: nil) as? TextList {
+            styleOptional = textList.style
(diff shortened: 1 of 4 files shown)


Aztec/Classes/Formatters/TextListFormatter.swift
```diff
@@ -37,6 +37,39 @@ class TextListFormatter
         let listParagraphs = string.paragraphRanges(spanningRange: listRange)
         return applyList(ofStyle: style, toString: string, atRanges: listParagraphs)
     }
+
+    func refreshList(inString string: NSMutableAttributedString, atRange range: NSRange) -> NSRange? {
```
diegoreymendez: I think this method requires some documentation explaining:

- Why does this method need to refresh the list?
- How does the refresh work?

Aztec/Classes/Formatters/TextListFormatter.swift
```diff
@@ -37,6 +37,39 @@ class TextListFormatter
...
+        }
+        if let textList = string.attribute(TextList.attributeName, atIndex: firstParagraphRange.location, effectiveRange: nil) as? TextList {
+            styleOptional = textList.style
+        }
+
+        guard let style = styleOptional else {
+            return removeList(fromString: string, atRanges: [firstParagraphRange])
+        }
+
+
```
diegoreymendez: Nitpicky: Remove 1 line break.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -455,6 +496,37 @@ public class TextView: UITextView {
         selectedRange = NSRange(location:selectionStartRange.location, length: selectionEndRange.location - selectionStartRange.location)
     }
 
+    // MARK: - List Code
+
+    private func refreshListsIfOnePrecedes(range range:NSRange) {
```
diegoreymendez: Definitely needs documentation - kinda wondering what "if one precedes" means.

SergioEstevao: @diegoreymendez ready for another look.

PR #152, opened 2016-11-11
Adds node descriptors.

diegoreymendez:
This is part of my cleanup work before I start implementing undo/redo.  The reason I'm tackling this is that I want to make sure our methods don't pass an enormous amount of parameters around.

This PR introduces the concept of node descriptors to simplify `ElementNode` searching and creation, while simplifying our calls.

**Known issues:**

While implementing this feature I noticed an issue with styles not being properly removed from the HTML source anymore.  This issue however was not introduced by this PR, so I opened a separate issue to fix it.

I uploaded a video of the problem [here](https://cloudup.com/cngEAn9uNba).

**How to test:**

Test the editor normally.  Try to come up with edge scenarios and make sure nothing breaks.
Run the unit tests.

### DIFF
--- a/AztecTests/HTML/ElementNodeTests.swift
+++ b/AztecTests/HTML/ElementNodeTests.swift
@@ -658,7 +658,8 @@ class ElementNodeTests: XCTestCase {
         
         let wrapRange = NSRange(location: 0, length: text1.characters.count)
         
-        paragraph.forceWrap(range: wrapRange, inElement: ElementNodeDescriptor(name: "b"))
+        let boldElementDescriptor = ElementNodeDescriptor(elementType: .b)
+        paragraph.forceWrap(range: wrapRange, inElement: boldElementDescriptor)
         
         XCTAssertEqual(paragraph.children.count, 2)
         
@@ -692,7 +693,8 @@ class ElementNodeTests: XCTestCase {
         
         let wrapRange = NSRange(location: 0, length: fullText.characters.count)
         
-        paragraph.forceWrap(range: wrapRange, inElement: ElementNodeDescriptor(name: "b"))
+        let boldElementDescriptor = ElementNodeDescriptor(elementType: .b)
+        paragraph.forceWrap(range: wrapRange, inElement: boldElementDescriptor)
         
         XCTAssertEqual(paragraph.children.count, 1)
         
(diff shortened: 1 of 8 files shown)


AztecTests/HTML/ElementNodeTests.swift
```diff
@@ -657,7 +658,7 @@ class ElementNodeTests: XCTestCase {
         
         let wrapRange = NSRange(location: 0, length: text1.characters.count)
         
-        paragraph.forceWrap(range: wrapRange, inNodeNamed: "b", withAttributes: [], equivalentElementNames: [])
+        paragraph.forceWrap(range: wrapRange, inElement: ElementNodeDescriptor(name: "b"))
```
SergioEstevao: We could use this PR to refactor the use of "b" for StandardElementType.b.rawValue. Just to get rid of this loose strings :)

AztecTests/HTML/ElementNodeTests.swift
```diff
@@ -691,7 +692,7 @@ class ElementNodeTests: XCTestCase {
         
         let wrapRange = NSRange(location: 0, length: fullText.characters.count)
         
-        paragraph.forceWrap(range: wrapRange, inNodeNamed: "b", withAttributes: [], equivalentElementNames: [])
+        paragraph.forceWrap(range: wrapRange, inElement: ElementNodeDescriptor(name: "b"))
```
SergioEstevao: We could use this PR to refactor the use of "b" for StandardElementType.b.rawValue. Just to get rid of this loose strings :)

AztecTests/HTML/ElementNodeTests.swift
```diff
@@ -721,7 +722,7 @@ class ElementNodeTests: XCTestCase {
 
         let wrapRange = NSRange(location: text1.characters.count, length: text2.characters.count)
 
-        paragraph.forceWrap(range: wrapRange, inNodeNamed: "b", withAttributes: [], equivalentElementNames: [])
+        paragraph.forceWrap(range: wrapRange, inElement: ElementNodeDescriptor(name: "b"))
```
SergioEstevao: We could use this PR to refactor the use of "b" for StandardElementType.b.rawValue. Just to get rid of this loose strings :)

diegoreymendez:
Thanks Sergio!

PR #155, opened 2016-11-16
Fix crash when typing return

koke:
The code in `refreshListIfNewLinesOn` would crash if you typed return at the
end of the text, as it tried to check for the next character which was out of
bounds.

Needs Review: @SergioEstevao


SergioEstevao:
:shipit:

PR #156, opened 2016-11-16
Fixes a crash with some zero-length ranges.

diegoreymendez:
Some zero length ranges were causing crashes.

**How to test:**

1. Change content.html to have: `<p>123</p>`
2. Select the range (loc: 2, len: 1)... the range for "3"
3. Apply bold
4. Remove bold

Make sure it doesn't crash.


SergioEstevao:
:shipit:

diegoreymendez:
Ty Sergio!

PR #159, opened 2016-11-18
Revert custom blockquote attribute and background

koke:
I've hit many dead ends and for now I think it's best to revert to the previous blockquote style.

When I started working on blockquotes, we had some simple indentation, which is also the style for this PR:

simulator screen shot 18 nov 2016 09 48 40

We wanted to match the style in calypso by drawing a background and border behind the blockquote:

simulator screen shot 18 nov 2016 09 50 07

Drawing the background is easy and still doable, but the problem is adding some padding. To do this, the only way I've found is using spacing in a paragraph style (you can also set the spacing via layout manager delegate methods, but it has some of the same problems):

simulator screen shot 18 nov 2016 09 56 16

That looks better, but as long as you add spacing, the caret height includes that extra spacing, and it just looks ridiculous. Overriding the text view `caretRect(for:)` method we could in theory fix this, so I tried adjusting the default caret to match the vertical position and height of the current line fragment:

simulator screen shot 18 nov 2016 09 57 39

Which mostly works, until you tap around the end of a line and the caret appears out of place. This is because when I try to calculate the line fragment for a position, I get the corresponding character which is always the character *after* the insertion point. When the insertion point is at the end of a line, you actually want the line fragment before it. `UITextPosition` has an `affinity` property that signals if we should look backwards or forward, but it's private. That works but we can't use private APIs:

simulator screen shot 18 nov 2016 10 26 30

After a week or two of banging my head against this, I can't come up with any solutions other than not using paragraph spacing.

sideshow bob annoyed grumble original content - imgur

The other roadblock is using a custom attribute to signal a blockquote. While you can set any custom attributes on an attributed string, `UITextView` has an internal whitelist for its `typingAttributes` property. So custom attributes might get dropped as you start editing. I tried overriding `typingAttributes` unsuccessfully: it looks like `UITextView.typingAttributes` is a wrapper for a private class `UITextInputController` which does most of the heavy work for the text view and can't be customized. For now I've reverted to using a simple paragraph style, and checking the `tailIndent` so it's less likely that we conflict with lists.

I can only think of two ways to deal with this in the future:
- Stick to standard attributes, and make sure any set of attributes is unique enough to match our virtual attributes (i.e. can we identify a nested list inside a blockquote?).
- Make the DOM the source of truth, then map that to a set of attributes.


SergioEstevao:
I'm closing this PR, we will keep some of the changes you made @koke, but I will use a custom ParagraphStyle instead of an attribute.

Issue #149, opened 2016-11-10
Lists - Implement in the DOM tree when reading from HTML


SergioEstevao:
Make sure that all lists are read correctly from HTML and translated to attributes strings correctly.

PR #160, opened 2016-11-21
Lists: HTML to NSAttributedString support

SergioEstevao:
#149 

This PR implements the parsing of ol, ul  and li element from HTML to attributed string.

How to test:
 - Using the sample content check if the OL and UL shows correctly in the attributed strings
 - Try to change the HTML to other valid OL and UL and see if they are parsed correctly
 - Switch from visual to HTML and see if list are represented back correctly

Note: This PR doesn't implement DOM interactions on list while using visual mode, so if lists are changed in visual mode they are not reflected to HTML mode.

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -16,6 +16,14 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     ///
     let defaultFontDescriptor: UIFontDescriptor
 
+    /// The current active list style
+    ///
+    var currentListStyle: TextList.Style?
+
+    /// The current active list number
+    ///
+    var currentListNumber: Int = 0
+
     required init(usingDefaultFontDescriptor defaultFontDescriptor: UIFontDescriptor) {
         self.defaultFontDescriptor = defaultFontDescriptor
     }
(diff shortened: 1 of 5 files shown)


Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -113,13 +113,12 @@ class HMTLNodeToNSAttributedString: SafeConverter {
         let childAttributes = attributes(forNode: node, inheritingAttributes: inheritedAttributes)
 
         for child in node.children {
-            let childContent = convert(child, inheritingAttributes: childAttributes)
-
+            let childContent = NSMutableAttributedString(attributedString:convert(child, inheritingAttributes: childAttributes))
```
diegoreymendez: I tried reverting this change and everything works fine.  Any reason why you're converting the string to be mutable here?

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -113,13 +113,12 @@ class HMTLNodeToNSAttributedString: SafeConverter {
...
             content.appendAttributedString(childContent)
         }
 
         if node.isBlockLevelElement() && !node.isLastChildBlockLevelElement() {
-            content.appendAttributedString(NSMutableAttributedString(string: "\n", attributes: childAttributes))
+            content.appendAttributedString(NSMutableAttributedString(string: "\n", attributes: inheritedAttributes))
```
diegoreymendez: Following up on the logic above, this can be safely changed to `NSAttributedString`.

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -273,4 +287,8 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     private func isImage(node: ElementNode) -> Bool {
         return node.name == StandardElementType.img.rawValue
     }
+
+    private func isNode(node:ElementNode, ofType type:StandardElementType) -> Bool {
```
diegoreymendez: This method looks like a good candidate to move to `ElementNode`.

Aztec/Classes/Libxml2/DOM/ElementNode.swift
```diff
@@ -91,7 +91,13 @@ extension Libxml2 {
         /// - Returns: true if the last child of this element is a block level element, false otherwise
         ///
         func isLastChildBlockLevelElement() -> Bool {
-            if let lastChild = children.last as? ElementNode {
+            let childrenFiltered = children.filter { (node) -> Bool in
```
diegoreymendez: This may become slightly more readable using a reversed loop... something like this:

```swift
for child in children.reversed() {
    if let elementNode = node as? ElementNode {
        return elementNode.isBlockLevelElement()
    } else if let textNode = node as? TextNode where textNode.text().isEmpty {
        continue
    } else {
        return false
    }
}
```

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
SergioEstevao: You are right, I changed it for the previous version of code, and the forgot to change it back when moved the code away from this method. I will clear this out.

SergioEstevao:

PR #163, opened 2016-12-05
Use custom paragraph style for lists.

SergioEstevao:
So I decided to change our approach for lists, instead of using a custom attribute I implemented a new ParagraphStyle class that inherits from NSMutableParagraphStyle, that implements the textList attribute in a similar way macOS does.

The big advantages of this is that we get all the automatic magic of copying attributes by the default implementation  of TextView so we don't need to applied themselves when changes happen in the list.

How to test:
 - Launch the demo app
 - Check if the initial import from HTML goes correctly
 - Play around with lists by toggling them, adding new lines, delete lines and see if all works well.

Note: This still doesn't implement the updates on the DOM so any changes done on the string after the initial import from HTML are not reflected back to the HTML


Aztec/Classes/Extensions/NSAttributedString+Lists.swift
```diff
@@ -63,13 +63,82 @@ extension NSAttributedString
     func rangeOfTextList(atIndex index: Int) -> NSRange? {
         var effectiveRange = NSRange()
         let targetRange = rangeOfEntireString
-        guard let _ = attribute(TextList.attributeName, at: index, longestEffectiveRange: &effectiveRange, in: targetRange) as? TextList else {
+        guard
+            let paragraphStyle = attribute(NSParagraphStyleAttributeName, at: index, longestEffectiveRange: &effectiveRange, in: targetRange) as? ParagraphStyle,
+            let _ = paragraphStyle.textList
```
jleandroperez: Can we remove this line?. If we're relying on an internal initialization this triggers, can we add a comment?

Aztec/Classes/Extensions/NSAttributedString+Lists.swift
```diff
@@ -63,13 +63,82 @@ extension NSAttributedString
...
 
+    /// Returns the range of the given text list that contains the given location.
+    ///
+    /// - Parameter list: The textList to search for.
+    /// - Parameter location: The location in the text list.
+    ///
+    /// - Returns: An NSRange optional containing the range of the list or nil if no list was found.
+    ///
+    func range(of list: TextList,
+               at location: Int) -> NSRange? {
```
jleandroperez: Can we have a single line for the method signature? (Unless this is an Aztec convention!)

Aztec/Classes/Extensions/NSAttributedString+Lists.swift
```diff
@@ -63,13 +63,82 @@ extension NSAttributedString
...
+    }
+
+    /// Returns the index of the item at the given location within the list.
+    ///
+    /// - Parameters:
+    ///   - list: The list.
+    ///   - location: The location of the item.
+    /// - Returns: Returns the index within the list.
+    func itemNumber(in list: TextList,
+                    at location: Int) -> Int {
```
jleandroperez: Same as above, can we have a single-line method signature?

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -47,11 +47,27 @@ class LayoutManager: NSLayoutManager {
...
+                attributes.removeValue(forKey: NSStrikethroughStyleAttributeName)
+                attributes.removeValue(forKey: NSLinkAttributeName)
+                if let font = attributes[NSFontAttributeName] as? UIFont {
+                    var traits = font.fontDescriptor.symbolicTraits
+                    traits.remove(.traitBold)
+                    traits.remove(.traitItalic)
+                    let descriptor = font.fontDescriptor.withSymbolicTraits(traits)
+                    let newFont = UIFont(descriptor: descriptor!, size: font.pointSize)
+                    attributes[NSFontAttributeName] = newFont
+                }
```
jleandroperez: Are we suuuure this is the best spot to alloc a new Font / update few values here and there? (considering this method's name is `drawBackground`)

Aztec/Classes/Extensions/NSAttributedString+Lists.swift
```diff
@@ -63,13 +63,82 @@ extension NSAttributedString
...
+        while resultRange.endLocation < self.length {
+            if
+                let paragraphStyle = attribute(NSParagraphStyleAttributeName, at: resultRange.endLocation, longestEffectiveRange: &effectiveRange, in: targetRange) as? ParagraphStyle,
+                let foundList = paragraphStyle.textList,
+                foundList == list {
+                resultRange = resultRange.union(withRange: effectiveRange)
+            } else {
+                break;
+            }
+        }
```
jleandroperez: Do we have any specific case in which the initial `longestEffectiveRange` returns an invalid value... and really has to be corrected with the loops?. If so, that would be a very good Unit Test candidate!

Aztec/Classes/Extensions/NSAttributedString+Lists.swift
SergioEstevao: no really, I sometime use this if the method signature is too wide but not in this case.

Aztec/Classes/Extensions/NSAttributedString+Lists.swift
SergioEstevao: It returns a correct value, but sometimes the NSParagraphStyle style instance can be different but is the same TextList, so we need to expand the range to grab all the TextList coverage.

Aztec/Classes/Extensions/NSAttributedString+Lists.swift
SergioEstevao: ok

Aztec/Classes/TextKit/LayoutManager.swift
SergioEstevao: You are right, my intention here was to use for the markers the same font that is used in the line, but without any styling or else you end up with bold, italic, striketrouh markers.
I can refactor this to another method in the ParagraphStyle class. 

Another alternative is to have a static font that it's always used for markers...

What do you think @jleandroperez ?

jleandroperez: Sounds good Sergio!. Main goal is to just minimize the runtime of this method. We don't really know how often it may get called

SergioEstevao:

jleandroperez:
:shipit:

SergioEstevao:
Thanks @jleandroperez !

Issue #108, opened 2016-10-11
Copy / Paste support for lists.


diegoreymendez:
Implement copy / paste support for lists.

diegoreymendez:
This was closed by mistake due to a bad refrence in a PR.

PR #166, opened 2016-12-06
Fix copy paste for lists, block quotes and images

### LINKED ISSUE #107: Copy / Paste Support for Blockquotes
Implement copy / paste support for blockquotes.

SergioEstevao:
Fix #107 and #108

This PR implements support for NSCoding in all our custom attributes to allow them to be copy pasted, inside the editor.

It keeps the original pasteboard in place in order to allow other applications to work correctly when pasting in or copy out.

How to test:
 - Copy paste segments of text inside the editor from one place to another and check if the behaviour is correct.

Note: Lists and Blockquotes are still not updating the DOM so changes made are not reflected in the HTML view.

### DIFF
--- a/Aztec/Classes/Extensions/NSAttributedString+Encoding.swift
+++ b/Aztec/Classes/Extensions/NSAttributedString+Encoding.swift
@@ -1,0 +1,19 @@
+import Foundation
+
+// MARK: - NSAttributedString Enconding methods
+//
+extension NSAttributedString
+{
+    static let pastesboardUTI = "com.wordpress.aztec.attributedString"
+
+    func archiveToData() -> Data {
+        let data = NSKeyedArchiver.archivedData(withRootObject: self)
+        return data
+    }
+
+    static func unarchive(fromData data: Data) -> NSAttributedString? {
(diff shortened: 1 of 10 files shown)


Aztec/Classes/Extensions/NSAttributedString+Encoding.swift
```diff
@@ -0,0 +1,19 @@
...
+extension NSAttributedString
+{
+    static let pastesboardUTI = "com.wordpress.aztec.attributedString"
+
+    func archiveToData() -> Data {
+        let data = NSKeyedArchiver.archivedData(withRootObject: self)
+        return data
+    }
+
+    static func unarchive(fromData data: Data) -> NSAttributedString? {
```
jleandroperez: In Swift 3 style this would probably look like...

```
static func unarchive(from data: Data) -> NSAttributedString?
```

(No need to change anything though!!)

jleandroperez:
Added two minor code comments (feel free to disregard!!). Other than that, works beautifully.

Awesome work Sergio!!!

:shipit:

PR #169, opened 2016-12-08
Implement the list attributes using the same protocols has block quotes.

SergioEstevao:
This PR consolidates our formatters to use the same set of protocols allows to get less code and a more consistent approach to paragraph attributes.

How to test:

- Apply list attributes to paragraph and check if the behaviour is correct.

Note:

This changes some of the requirements of lists defined on #34 :

- Scenario 5: all paragraphs get the style applied
- Scenario 6: all paragraphs get the style applied
- Scenario 7: all paragraphs get the style applied

I actually think this is more consistent behaviour, but lets evaluate. If we really need that behaviour we can add it to the standard attributes protocol.

### DIFF
--- a/Aztec/Classes/Constants/Metrics.swift
+++ b/Aztec/Classes/Constants/Metrics.swift
@@ -7,7 +7,9 @@ import Foundation
 enum Metrics
 {
 
-    static let defaultIndentation = CGFloat(20)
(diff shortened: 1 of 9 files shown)


SergioEstevao:
@diegoreymendez regarding scenario 2, are you sure you got the last version of the PR?

PR #170, opened 2016-12-08
Refactor layout manager code 

SergioEstevao:
Break the layout manager code in smaller methods and create some variable for block quotes colors and allow customisation.

### DIFF
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -27,7 +27,7 @@ class LayoutManager: NSLayoutManager {
         //draw blockquotes
         textStorage.enumerateAttribute(NSParagraphStyleAttributeName, in: characterRange, options: []){ (object, range, stop) in
             guard let paragraphStyle = object as? ParagraphStyle,
-                let _ = paragraphStyle.blockquote
+                  paragraphStyle.blockquote != nil
                 else {
                     return
             }
@@ -54,7 +54,7 @@ class LayoutManager: NSLayoutManager {
         // draw list markers
         textStorage.enumerateAttribute(NSParagraphStyleAttributeName, in: characterRange, options: []){ (object, range, stop) in
             guard let paragraphStyle = object as? ParagraphStyle,
-                let _ = paragraphStyle.textList
+                  paragraphStyle.textList != nil
                 else {
                     return
             }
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -4,8 +4,17 @@ import QuartzCore
 
 class LayoutManager: NSLayoutManager {
 
+    var blockquoteBorderColor: UIColor = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
+    var blockquoteBackgroundColor = UIColor(red: 0.91, green: 0.94, blue: 0.95, alpha: 1.0)
```
diegoreymendez: These should be set in the `LayoutManager` initializer call IMHO, in order to force the owner app to set a default value.

[The method that initializes the `LayoutManager` instance](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/blob/try/refactor_layoutmanager/Aztec/Classes/TextKit/TextView.swift#L56) already sets a few other defaults, so it would be in-line with the rest of the code.

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -24,26 +32,31 @@ class LayoutManager: NSLayoutManager {
...
+        guard let textStorage = textStorage else {
+            return
+        }
 
+        let characterRange = self.characterRange(forGlyphRange: glyphsToShow, actualGlyphRange: nil)
         // draw list markers
         textStorage.enumerateAttribute(NSParagraphStyleAttributeName, in: characterRange, options: []){ (object, range, stop) in
             guard let paragraphStyle = object as? ParagraphStyle,
-                  let _ = paragraphStyle.textList
+                let _ = paragraphStyle.textList
```
diegoreymendez: Note: You can leave this as it is right now if you prefer.

IMHO whenever we're not using the value checking for `paragraphStyle.textList != nil` is somewhat clearer.

Aztec/Classes/TextKit/LayoutManager.swift
SergioEstevao: I don't think this needs to go for the initialiser when we can have a reasonable default value. None of the UIKit default components force you to set a default value, but they all allow customisation by the app.

Aztec/Classes/TextKit/LayoutManager.swift
SergioEstevao: Agree on this, will change this.

PR #171, opened 2016-12-12
Applying styles at cursor position

### LINKED ISSUE #167: Applying styles to the caret position.
Applying styles to the caret position should be supported.  This should behave in a way similar to how `UITextView`'s `typingAttributes` works.

Ref: https://developer.apple.com/reference/uikit/uitextview/1618629-typingattributes

Requirements:
- Applying a style to the caret position (even with a zero-length selection, so to speak) should remember the attributes so when the user types something in, the style is maintained.
- Pasting anything should drop the saved styles.
- Changing the selection in any way should drop the saved styles.
- The toolbar should light up accordingly.

jleandroperez:
### Details:
This PR adds support to toggle Text Styles at the cursor's position, without requiring the user to preselect the paragraphs, previously.

Needs Review: @diegoreymendez 
Closes #167

### Testing: No Selection
1. Hit the `Bold`/ `Italic` / `Underline` / `Strikethrough` button
2. Verify that the Editor behaves as expected

### Testing: Selection
1. Select a snippet of text
2. Hit the `Bold`/ `Italic` / `Underline` / `Strikethrough` button
3. Verify that the selected snippet gets styled as expected

### Testing: Toolbar
1. Place the cursor over text in `Bold`/ `Italic` / `Underline` / `Strikethrough`
2. Verify that the Toolbar lights as expected

### Testing: Paste
1. Toggle any of the `Bold`/ `Italic` / `Underline` / `Strikethrough` styles
2. Paste some random text
3. Verify that the pasted text reflects the style that was previously selected.

**Note:** This behavior is different from the one outlined in the original issue. Reasons why:
- Pages does it this way ("paste and match style")
- If you explicitly select `Bold` / else, and proceed to paste, it makes sense to respect that preselected style

### Testing: Selection
1. Toggle any of the text styles
2. Select a random snippet of text
3. Type anything

Verify that the styles selected in step (1) get lost


jleandroperez:
Update: Dealing with the Paste Behavior in #172

Aztec/Classes/Extensions/UIFont+Traits.swift
```diff
@@ -0,0 +1,42 @@
...
+
+
+// MARK:- UIFont Traits Helpers
+//
+extension UIFont {
+
+    /// Returns a new instance of the current font with its SymbolicTraits Updated.
+    ///
+    /// - Parameters:
+    ///     - trait: Trait to be updated
```
diegoreymendez: Since the `trait` parameter is a mask which can potentially hold several traits, it should be named `traits` (plural) instead.

Aztec/Classes/Extensions/UIFont+Traits.swift
```diff
@@ -0,0 +1,42 @@
...
+    ///     - enable: Boolean indicating whether we should insert or remove the specified trait.
+    ///
+    /// - Returns: A new UIFont with the same descriptors as the current instance, but with its traits updated, as specified.
+    ///
+    func modifyTrait(_ trait: UIFontDescriptorSymbolicTraits, enable: Bool) -> UIFont {
```
diegoreymendez: `trait` -> `traits`

diegoreymendez: Also: nice work... the code in this method should be used [here](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/blob/develop/Example/Example/FormattingDemoController.swift#L148) and [here](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/blob/develop/Example/Example/FormattingDemoController.swift#L106).

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -301,6 +301,33 @@ open class TextView: UITextView {
     }
 
 
+    /// Get a list of format identifiers of the Typing Attributes.
+    ///
+    /// - Returns: A list of identifiers.
+    ///
+    open func formatIdentifiersForTypingAttributes() -> [String] {
```
diegoreymendez: I just realized we're using `[String]` instead of `[FormattingIdentifier]` throughout our code.  Sounds like something we should change (unless there's a reason not to, let me know).

jleandroperez:
@diegoreymendez ready for another look!.

Regarding wiring UIFont+Traits, not sure if we want to expose that extension to the Host App. (For now it's just an internal helper!)

Thanks!

diegoreymendez:
Everything seems to be working fine except:

Enabling styles works fine, but when I disable any style, the toolbar still shows it as active until I press any key or move the caret.

jleandroperez:
Thanks a lot Diego!!

PR #176, opened 2016-12-13
Paste and Match Style

jleandroperez:
### Details:
This PR adds the ability to Paste and Match the surrounding text's style. What we're doing here is:
- Adding a new `UIMenuController` item
- Simply grabbing the `UIPasteboard`'s string property, if any
- If there's no string property to be found, we'll fallback to the default behavior

### Testing:
1. Launch the Demo App
2. Switch over to Safari / Notes / Reminders
3. Copy some random text
4. Switch back to Aztec's Demo
5. Try the new `Paste and Match Style` action

Note that the original string's style should be striped out.

Needs Review: @diegoreymendez


jleandroperez:
Thank you Diego!!

Issue #60, opened 2016-09-05
Undo / Redo support.


diegoreymendez:
There's a partially-working undo support in Aztec, however the DOM structure is not properly re-created when performing more complex undo / redo operations.

Make sure full support is added.

diegoreymendez:
We no longer have a DOM structure, but we still need to make sure every action works with full undo / redo support.

PR #183, opened 2016-12-16
Improves the undo manager code.

diegoreymendez:
Relates to #60 

Improves the undo management code to avoid having to propagate `UndoManager` to every single node in the DOM tree.

Unit tests have to provide an empty block, since nodes are not supposed to exist without an undo block (by design, at least in this v1 of undo management - we can review this later on).

**How to test:**

- Delete text from single node.
- Move the caret around and repeat.
- Shake to undo any times you deem necessary.
- Check the HTML is as expected.
- Shake to redo any times you deem necessary.
- Check the HTML is as expected.


jleandroperez:
Beautiful changes!!

:shipit:

diegoreymendez:
As pointed out by @jleandroperez, undo and redo don't work properly yet when editing multiple nodes simultaneously.

I edited the description of this PR to reflect this.

Thanks @jleandroperez !

PR #186, opened 2016-12-20
TextView: Adds missing method documentation

jleandroperez:
### Details:
Added missing methods documentation, and minor changes.

Needs Review: @diegoreymendez 
Thanks in advance!

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -596,7 +596,9 @@ open class TextView: UITextView {
         formatter.toggleAttribute(inTextView: self, atRange: range)
     }
 
+
     // MARK: - Blockquotes
+
 
     /// Adds or removes a blockquote style from the specified range.
     /// Blockquotes are applied to an entire paragrah regardless of the range.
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -611,11 +613,13 @@ open class TextView: UITextView {
         formatter.toggleAttribute(inTextView: self, atRange: range)
     }
 
-    /// Refresh Lists attributes when text is deleted in the specified range
+
+    /// Toggles the Blockquote Style, whenever the deleted text is at the Location 0, which is 
+    /// also known as the "Beginning of the text"
```
diegoreymendez: I think it'd be better if we bring back the original description, and concatenate your new documentation below it as a more detailed description of what happens.

The reason being that the original description described what the external usage for this class is, and when external actors would want to use this method, whereas the new documentation explains internal details about what it really does.

diegoreymendez: Also consider changing to "...whenever the deleted text is at location 0."  The "also known as the beginning of the text" part should be pretty clear anyway.

jleandroperez:

Thanks!!!

jleandroperez:
Thanks Diego!

PR #201, opened 2016-12-27
TextView: Missing Cursor Workaround

### LINKED ISSUE #144: Caret disappears on new line
I [reported this](http://openradar.appspot.com/radar?id=4998031044771840) to Apple, but we'll need to find a workaround.

The caret disappears when you are in a block of text that has a background and both indent and spacing in its paragraph style, and you type return. It seems it's only triggered by a hardware keyboard.

jleandroperez:
### Details:
We've got several flows (adding a blockquote // newline while in a blockquote) that cause the "Cursor Missing in Action" scenario.

In this PR we're adding a workaround to, at least, temporarily prevent this bug, until a proper fix is found (OR the SDK gets patched). It's not a fancy snippet, still, gets the job done.

@diegoreymendez May i bug you, yet again? (Thanks!!)
Closes #144

### Scenario A: Newline
1. Launch the empty editor
2. Press the Blockquote button

Verify that the cursor is visible and blinking

### Scenario B: Expansion
1. Launch the empty editor
2. Press the Blockquote button
3. Hit return. The blockquote should expand

Verify that the cursor is, still, visible

### Scenario C: Backspace
1. Launch the empty editor
2. Press the Blockquote button, and enter a couple lines of text (or empty lines!)
3. Hit backspace to delete one of the new lines

Verify that the cursor shows up in the previous line (as expected), and that it does not get covered by the blockquote BG.

### Scenario D: Vertical Flicker
1. Launch the empty editor
2. Enter (lots of) empty lines. Enough to cover the screen, and cause TextView Y-Scrolling
3. Hit the Backspace Button

Verify that there's no vertical flicker in the Vertical Scrolling position.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -136,7 +136,7 @@ open class TextView: UITextView {
         insertionRange.length = 1
         refreshListAfterInsertion(of: text, at: insertionRange)
         refreshBlockquoteAfterInsertion(of: text, at: insertionRange)
-        forceRedrawCursor()
+        forceRedrawCursorAfterDelay()
     }
 
     open override func deleteBackward() {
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thank you Diego!

PR #205, opened 2016-12-28
Lists: Rendering Marker at the EOF

### LINKED ISSUE #203: Lists: Invisible New Element
### Steps:
1. Launch the empty editor demo
2. Hit the `Lists` Format Bar Action
3. Hit `Return`

Note that the cursor will be displayed with the right indentation, but the list prefix (`2.`) won't show up, until the user presses yet another character.

jleandroperez:
### Steps:
1. Launch the empty editor demo
2. Hit the `Lists` Format Bar Action
3. Hit `Return`

Verify that the "New Line"'s list marker gets displayed immediately.

Closes #203

cc @diegoreymendez 
Thanks in advance Diego!

### DIFF
--- a/Aztec/Classes/Extensions/NSAttributedString+Lists.swift
+++ b/Aztec/Classes/Extensions/NSAttributedString+Lists.swift
@@ -119,13 +119,16 @@ extension NSAttributedString
     /// - Parameters:
     ///   - list: The list.
     ///   - location: The location of the item.
+    ///
     /// - Returns: Returns the index within the list.
+    ///
     func itemNumber(in list: TextList, at location: Int) -> Int {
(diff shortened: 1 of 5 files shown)


diegoreymendez:
This is mostly working OK except for the scenario shown in [this video](https://cloudup.com/cB3sDgXwwHR).

jleandroperez:
@diegoreymendez the glitch you've mentioned (as discussed via DM!) is actually a side effect of #206 

Only real difference is: it's noticeable in this PR, because otherwise the "extra new item" would not have been visible (due to #203).

jleandroperez:
Discussed via DM with @diegoreymendez!! Proceeding with the merge, and dealing with any bugs in new issues.

Thanks as ever sir!!

PR #209, opened 2017-01-02
Decoupling AttributeFormatters from UITextView

jleandroperez:
### Details:
- AttributeFormatter's API has been updated, with Swift 3 in mind
- UITextView is nowhere to be seen, within any of the formatters
- Added some extra method documentation here and there
- Unit Tests updated accordingly

### Testing:
Please, try out `Lists (Ordered / Unordered)` and `Blockquotes`, in any possible scenarios:

- Empty Documents
- At the beginning / middle / bototm of a document

We're aware of several Lists / Blockquote glitches (all of them are filed, to the best of my knowledge).
We'll be addressing them ASAP.

Needs Review: @diegoreymendez 
Thanks in advance!

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -590,7 +590,7 @@ open class TextView: UITextView {
     ///
     fileprivate func remove(list: TextList, at range: NSRange) {
         let formatter = TextListFormatter(style: list.style)
-        formatter.toggleAttribute(inTextView: self, atRange: range)
+        formatter.toggleAttribute(in: textStorage, at: range)
     }
 
 
@@ -600,7 +600,7 @@ open class TextView: UITextView {
     ///
     open func toggleOrderedList(range: NSRange) {
         let formatter = TextListFormatter(style: .ordered)
-        formatter.toggleAttribute(inTextView: self, atRange: range)
+        formatter.toggleAttribute(in: textStorage, at: range)
     }
 
 
@@ -610,7 +610,7 @@ open class TextView: UITextView {
     ///
     open func toggleUnorderedList(range: NSRange) {
         let formatter = TextListFormatter(style: .unordered)
-        formatter.toggleAttribute(inTextView: self, atRange: range)
+        formatter.toggleAttribute(in: textStorage, at: range)
     }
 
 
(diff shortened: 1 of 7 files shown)


AztecTests/Formatters/BlockquoteFormatterTests.swift
```diff
@@ -5,64 +5,69 @@ import Gridicons
 class BlockquoteFormatterTests: XCTestCase {
     func testApplyingBlockquoteOnFirstParagraph() {
         let textView = testTextView
-        let paragraphs = paragraphRanges(inString: textView.storage)
+        let storage = textView.storage
+        let paragraphs = paragraphRanges(inString: storage)
 
         let formatter = BlockquoteFormatter()
-        formatter.toggleAttribute(inTextView: textView, atRange: NSRange(location: 1, length: 1))
+        _ = formatter.toggle(in: storage, at: NSRange(location: 1, length: 1))
```
diegoreymendez: In my modest opinion the `toggle` method definition should have `@discardableResult` added to it, so this assignment is not needed.

It sounds reasonable for some usages to let you ignore the returned range.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -674,17 +679,19 @@ open class TextView: UITextView {
         }
 
         let isBegginingOfListItem = storage.isStartOfNewLine(atLocation: range.location)
+        guard text == "\n" && beforeString == "\n" && afterString == "\n" && isBegginingOfListItem else {
```
diegoreymendez: I know this is coming from older code, but having the newline character not defined as a constant seems wrong.

We can maybe have a new file for character-constants-of-relevance.

diegoreymendez: We can opt to use [this approach](https://github.com/wordpress-mobile/WordPress-Aztec-iOS/pull/209/files#diff-3660cac0e28198d6628cd6fce9c2feb9R82) as well.

Either way we should unify a solution for these character constants.

jleandroperez: If it's okay with you, let's, instead, delay refactoring the `refreshAfterInsertion` / `refreshAfterDeletion` methods.

We may (_may_) have a master fix for all of the weirdness'ess we've been seeing around, and it would definitely involve some extra work in those precise methods.

Thanks!!!

AztecTests/Formatters/BlockquoteFormatterTests.swift
jleandroperez: Sounds awesome!! updated accordingly, thanks for catching that!

Aztec/Classes/TextKit/TextView.swift
diegoreymendez: If you're going to keep working on these methods feel free to skip for now.  Let's just make that the final cut doesn't have magical strings.

jleandroperez:
Merci @diegoreymendez!!

PR #213, opened 2017-01-09
Undo / redo improvements

diegoreymendez:
<h3>Details</h3>

This PR introduces several small improvements to undo / redo.

Improvements:
- We're no longer passing around a closure, but an `EditContext` object which provides access to the undo manager.  This is better because we now have access, from almost anywhere, to the methods [`enableUndoRegistration()`](enableUndoRegistration) and [`disableUndoRegistration()`](https://developer.apple.com/reference/foundation/undomanager/1412239-disableundoregistration).
- I'm expecting the edit context will contain more state information eventually.  If it turns out it doesn't, we can reconsider the architecture.

<h3>Testing</h3>

**Test:**

- Make sure there are no big crashes caused by this PR by randomly testing the editor.
- Not all operations support undo operations, so testing in the editor may be tricky.  For now I suggest to just run the unit tests and take them as an indication that the implemented undo operations are working fine.


diegoreymendez:
> ...switch between html mode and back you loose your undo, redo stack

Yes.  I'm not sure what's triggering that, but I'm leaving that as lower priority for the time being.

> ...One thing to consider is to bunker all the changes done in the html node as an undo/redo step?

Once undo / redo is working completely in visual mode, I'll work on trying to make it work just as well in HTML mode.

diegoreymendez:
Thanks Sergio!

PR #214, opened 2017-01-10
Blockquotes changes update DOM

SergioEstevao:
Changes made to block quotes update the DOM structure.

@diegoreymendez this is a first stab on updating the DOM when applying block quotes, this is still not final but I would like to have some feedback from you of the changes I made.

My main questions are:

 - Should we move the formatters inside the TextStorage class instead of having on the view.
 - I'm not very happy with the way we are removing elements from the DOM, you have some comments around about using TextStorage.setAttributes: 
     

> We should be calculating what attributes to remove in 
>     `TextStorage.setAttributes()` but since that may take a while to implement, we need this workaround until it's ready.

What was your idea there?

### DIFF
--- a/Aztec/Classes/Extensions/NSRange+Helpers.swift
+++ b/Aztec/Classes/Extensions/NSRange+Helpers.swift
@@ -5,6 +5,17 @@ import Foundation
 //
 extension NSRange
 {
+    /// Checks if the receiver contains the specified range.
+    ///
+    /// - Parameters:
+    ///     - range: the range that the receiver may or may not contain.
+    ///
+    /// - Returns: `true` if the receiver contains the specified range, `false` otherwise.
+    ///
+    func contains(range: NSRange) -> Bool {
+        return intersect(withRange: range) == range
+    }
+    
     /// Returns the intersection between the receiver and the specified range.
     ///
     /// - Important: the main difference with NSIntersectionRange is that this method considers any contact (even a
--- a/Aztec/Classes/Libxml2/DOM/ElementNode.swift
+++ b/Aztec/Classes/Libxml2/DOM/ElementNode.swift
@@ -1436,7 +1436,7 @@ extension Libxml2 {
         ///
         fileprivate func forceWrapChildren(intersectingRange targetRange: NSRange, inElement elementDescriptor: ElementNodeDescriptor) {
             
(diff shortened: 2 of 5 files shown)


diegoreymendez:
Found a crash.  Steps to reproduce:

1. Launch the empty editor
2. Focus the content field
3. Tap the blockquote toolbar button before you type anything
4. Type in 2 characters

Expected behaviour: the two characters should come up.
Current beahviour: the app is crashing in `ElementNode:1069`.

diegoreymendez:
When typing in the second character, and before it's added, this is the DOM structure I'm seeing (notice the extra blockquotes):

```
(lldb) po self
▿ RootNode
  ▿ super : ElementNode
    ▿ super : Node
      - name : "aztec.htmltag.rootnode"
      - rawParent : nil
      ▿ editContext : Optional<EditContext>
        ▿ some : <EditContext: 0x61000003f220>
    - attributes : 0 elements
    ▿ children : 3 elements
      ▿ 0 : ElementNode
        - type : "element"
        - name : "blockquote"
        - parent : "Optional(Aztec.Libxml2.RootNode)"
        - attributes : 0 elements
        ▿ children : 1 element
          ▿ 0 : TextNode
            - type : "text"
            - name : "text"
            - text : "​"
            - parent : "Optional(Aztec.Libxml2.ElementNode)"
      ▿ 1 : ElementNode
        - type : "element"
        - name : "blockquote"
        - parent : "Optional(Aztec.Libxml2.RootNode)"
        - attributes : 0 elements
        ▿ children : 1 element
          ▿ 0 : ElementNode
            - type : "element"
            - name : "blockquote"
            - parent : "Optional(Aztec.Libxml2.ElementNode)"
            - attributes : 0 elements
            - children : 0 elements
      ▿ 2 : ElementNode
        - type : "element"
        - name : "blockquote"
        - parent : "Optional(Aztec.Libxml2.RootNode)"
        - attributes : 0 elements
        ▿ children : 1 element
          ▿ 0 : TextNode
            - type : "text"
            - name : "text"
            - text : "a"
            - parent : "Optional(Aztec.Libxml2.ElementNode)"
  - name : "aztec.htmltag.rootnode"
  ▿ children : 3 elements
    ▿ 0 : ElementNode { ... }
    ▿ 1 : ElementNode { ... }
    ▿ 2 : ElementNode { ... }
```

PR #215, opened 2017-01-12
Undo support for removing child nodes.

diegoreymendez:
Adds support for undoing child removal.

**How to test:**

Unfortunately testing this straight in the editor is prone to errors.  So the best way to make sure this works as expected is to review the provided unit test and make sure the logic is correct.


diegoreymendez:
Thanks Sergio!

PR #219, opened 2017-01-17
Make sure typing attributes are correct after applying styles.

SergioEstevao:
How to test:

- Start the editor demo 
- Go to a line toggle blockquote/list on a paragraph
- Check if the format bar is being updated correctly.


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -558,6 +558,12 @@ open class TextView: UITextView {
     }
 
 
+    private func updateTypingAttributes() {
+        if (textStorage.length > 0){
+            let location = max(0,min(selectedRange.location, textStorage.length-1))
```
diegoreymendez: I'd just recommend to add an explanation as to why we're doing `max(min())`, so that when we come back to it we immediately understand the reasons behind this choice.

PR #222, opened 2017-01-18
Implement bold italic with formatters

SergioEstevao:
This PR changes the way we apply bold and italic to the storage, by using CharacterFormatters instead of specific code in the TextView and TextStorage classes.

How to test:

- Apply bold and/or italic to sections of test and see if everything works ok.
 - Check if the HTML is updated correctly also.

### DIFF
--- a/Aztec/Classes/Formatters/FontFormatter.swift
+++ b/Aztec/Classes/Formatters/FontFormatter.swift
@@ -1,0 +1,62 @@
+import Foundation
+import UIKit
+
+class FontFormatter: CharacterAttributeFormatter {
+
+    let elementType: Libxml2.StandardElementType
+    
+    let traits: UIFontDescriptorSymbolicTraits
+
+    init(elementType: Libxml2.StandardElementType, traits: UIFontDescriptorSymbolicTraits) {
+        self.elementType = elementType
+        self.traits = traits
+
+    }
+
+    func apply(to attributes: [String : Any]) -> [String: Any] {
+
+        var resultingAttributes = attributes
+        guard let font = attributes[NSFontAttributeName] as? UIFont else {
+            return attributes
+        }
+        let newFont = font.modifyTraits(traits, enable: true)
+        resultingAttributes[NSFontAttributeName] = newFont
+        
+        return resultingAttributes
+    }
+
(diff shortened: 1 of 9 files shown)


Aztec/Classes/Formatters/FontFormatter.swift
```diff
@@ -0,0 +1,62 @@
...
+        guard let font = attributes[NSFontAttributeName] as? UIFont else {
+            return attributes
+        }
+        let newFont = font.modifyTraits(traits, enable: true)
+        resultingAttributes[NSFontAttributeName] = newFont
+        
+        return resultingAttributes
+    }
+
+    func remove(from attributes:[String: Any]) -> [String: Any] {
```
diegoreymendez: Add a space after `from attributes:`, and add unit tests.

- One that removes a present attribute.
- One that tries to remove a missing attribute.

Aztec/Classes/Formatters/FontFormatter.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    let traits: UIFontDescriptorSymbolicTraits
+
+    init(elementType: Libxml2.StandardElementType, traits: UIFontDescriptorSymbolicTraits) {
+        self.elementType = elementType
+        self.traits = traits
+
+    }
+
+    func apply(to attributes: [String : Any]) -> [String: Any] {
```
diegoreymendez: This is a really good opportunity to add more unit tests, since these methods are so atomic.

Consider adding: 1 unit test adding an attribute that's not in the provided list, and 1 unit test to add an attribute that's already in there.

Aztec/Classes/Formatters/FontFormatter.swift
```diff
@@ -0,0 +1,62 @@
...
+            return attributes
+        }
+
+        let newFont = font.modifyTraits(traits, enable: false)
+        resultingAttributes[NSFontAttributeName] = newFont
+
+        return resultingAttributes
+    }
+
+    func present(in attributes: [String : Any]) -> Bool {
```
diegoreymendez: Add a few unit tests:

- One where the attribute is present.
- One where it's not.

SergioEstevao:
@diegoreymendez just added the tests you requested, ready for another look.

PR #226, opened 2017-01-26
Implement striketrough and underline using formatters

SergioEstevao:
This PR implements strike trough and underline using formatters.
It also fix some bugs where applying one character formatter make all the characters in range getting the attributes at the start of the range.

How to test:
 - Run the demo app
 - Apply/Remove underline and/or strike trough to sections of text:
  - Places without the attribute
  - Places where the attribute is partial applied
  - Places with the attribute

### DIFF
--- a/Aztec/Classes/Formatters/AttributeFormatter.swift
+++ b/Aztec/Classes/Formatters/AttributeFormatter.swift
@@ -62,7 +62,7 @@ extension AttributeFormatter {
         return present(in: attributes)
     }
 
-    /// Indicates wheter the Formatter's Attributes are present in the full range provided
+    /// Indicates whether the Formatter's Attributes are present in the full range provided
     ///
     /// - Parameters:
     ///   - text: the attributed string to inspect for the attribute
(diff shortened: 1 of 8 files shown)


Aztec/Classes/Formatters/AttributeFormatter.swift
```diff
@@ -62,6 +62,23 @@ extension AttributeFormatter {
         return present(in: attributes)
     }
 
+    /// Indicates wheter the Formatter's Attributes are present in the full range provided
```
diegoreymendez: wheter -> whether

Aztec/Classes/Formatters/AttributeFormatter.swift
```diff
@@ -62,6 +62,23 @@ extension AttributeFormatter {
...
+    ///
+    /// - Parameters:
+    ///   - text: the attributed string to inspect for the attribute
+    ///   - range: the range to inspect
+    /// - Returns: true if the attributes exists on all of the range
```
diegoreymendez: For consistency, add an empty comment line both above and below this line.

PR #229, opened 2017-01-30
Standardize check for formatting

SergioEstevao:
Instead of using multiple methods to check for formatters active, this PR switch the approach to use a data driven approach where the mapping of formatters and typing attributes is done using a dictionary.

How to test:
 - Start the demo app
 - Navigate around and see if the toolbar is updated correctly.
 - Apply style to new ranges and location and see if the toolbar updates correctly.


jleandroperez:
@SergioEstevao this PR is absolutely beautiful. By all means, :shipit:

PR #231, opened 2017-01-31
Fixing Insert Link Crash

jleandroperez:
### Testing:
1. Launch the Aztec Empty Editor
2. Hit the **Insert Link** button
3. Verify that the app doesn't crash anymore
4. Run the Unit Tests and verify that nothing breaks

Needs Review: @SergioEstevao 
Thanks in advance!!

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -700,36 +700,47 @@ open class TextView: UITextView {
     ///
     /// - Parameter range: The NSRange to inspect
     ///
-    /// - Returns: the NSURL if available
+    /// - Returns: The NSURL if available
     ///
     open func linkURL(forRange range: NSRange) -> URL? {
         let index = maxIndex(range.location)
         var effectiveRange = NSRange()
-        if let attr = storage.attribute(NSLinkAttributeName, at: index, effectiveRange: &effectiveRange) {
-            if let url = attr as? URL {
-                return url
-            } else if let urlString = attr as? String {
-                return URL(string:urlString)
-            }
+        guard index > storage.length,
+            let attr = storage.attribute(NSLinkAttributeName, at: index, effectiveRange: &effectiveRange)
+            else {
+                return nil
         }
+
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -700,37 +700,48 @@ open class TextView: UITextView {
...
     open func linkURL(forRange range: NSRange) -> URL? {
         let index = maxIndex(range.location)
         var effectiveRange = NSRange()
-        if let attr = storage.attribute(NSLinkAttributeName, at: index, effectiveRange: &effectiveRange) {
-            if let url = attr as? URL {
-                return url
-            } else if let urlString = attr as? String {
-                return URL(string:urlString)
-            }
+        guard index > storage.length,
```
SergioEstevao: Maybe here you want ``` index < storage.length```

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -700,37 +700,48 @@ open class TextView: UITextView {
...
     ///
-    /// - Returns: The full Range required by the link.
+    /// - Returns: The full Link's Range.
     ///
     open func linkFullRange(forRange range: NSRange) -> NSRange? {
         let index = maxIndex(range.location)
         var effectiveRange = NSRange()
-        if storage.attribute(NSLinkAttributeName, at: index, effectiveRange: &effectiveRange) != nil {
-            return effectiveRange
+        guard index > storage.length,
```
SergioEstevao: Same here: ```index < storage.length```

SergioEstevao:
You may also need to fix the method maxIndex to:
```
func maxIndex(_ index: Int) -> Int {
        if index >= storage.length {
            return max(0,storage.length - 1)
        }

        return index
    }
```

To avoid returning a negative index.

jleandroperez:
@SergioEstevao thanks for the review!! ready for another look sir.

jleandroperez:
Thank you Sergio!!

PR #248, opened 2017-02-07
Updates the format bar with keyboard commands.

### LINKED ISSUE #244: Toolbar not updated with keyboard commands
It seems that using keyboard commands to apply styles (CMD + B, CMD + I, etc) is not properly refreshing our toolbar.

diegoreymendez:
<h3>Description</h3>

Updates the tooblar properly when CMD+i, CMD+b, CMD+u are pressed, even if there's no selection.

Fixes #244.

<h3>Testing:</h3>

**Test 1:**

1. Open the empty editor.
2. Without typing anything press CMD + b.
3. Make sure the toolbar is updated properly.

**Further tests:**

Try the same as test 1 with CMD + i and CMD + u.
Also try the same as test 1 in the editor with sample content.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -36,6 +36,15 @@ public protocol TextViewMediaDelegate: class {
         urlForImage image: UIImage) -> URL
 }
 
+public protocol TextViewFormattingDelegate: class {
+
+    /// Called a text view command toggled a style.
+    ///
+    /// If you have a format bar, you should probably update it here.
+    ///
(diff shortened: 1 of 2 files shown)


diegoreymendez:
> While all works perfectly, instead of creating a new delegate couldn't you just use the current delegate and send an textViewDidChangeSelection callback?

Yes but since the selection didn't really change, I don't want to trigger that callback.  While we don't do any special processing when the `textViewDidChangeSelection` callback is triggered, other apps using this library may do so, and we could end up creating confusion as to what's happening.

diegoreymendez:
Thank you @SergioEstevao !

PR #257, opened 2017-02-09
FormatBar: Marking Formatter as weak

jleandroperez:
### Details:
This (brief!) PR simply marks the Format Bar's delegate as **weak**, in order to patch a retain cycle that's affecting WPiOS.

Needs Review: @diegoreymendez 
Thanks in advance!


jleandroperez:
Thank you!

PR #258, opened 2017-02-09
Issue/media attachment improvements

SergioEstevao:
This PR improved the interaction with media by replacing the long press gesture for a double tap on image. First tap makes the image go to selection mode and second tap on it shows the related options.

How to test:
 - Start demo app
 - Tap on a image ( see if image get an overlay saying tap to edit)
 - Tap on the image ( see if image displays the action sheet)
 - Tap outside the image to see if overlay goes away.


jleandroperez:
Looking great sir!

:shipit:

PR #260, opened 2017-02-09
Update toolbar with grid icons and adding source code item.

SergioEstevao:
How to test:

 - Check if the icons on the toolbar have the gridicons look
 - Check if the source code button on the toolbar works correctly.


Example/Example/EditorDemoController.swift
```diff
@@ -322,6 +326,19 @@ extension EditorDemoController {
         richTextView.isHidden = false
         htmlTextView.isHidden = true
     }
+
+    fileprivate func toggleSourceCodeMode() {
```
diegoreymendez: Check out my proposed changes in #263

Example/Example/EditorDemoController.swift
```diff
@@ -43,6 +43,11 @@ class EditorDemoController: UIViewController {
         textView.translatesAutoresizingMaskIntoConstraints = false
         textView.isHidden = true
 
+        let toolbar = self.createToolbar(htmlMode: true)
```
SergioEstevao: the keyboardDismissMode is not part of the toolbar interface.

PR #262, opened 2017-02-09
Copy/Paste link from pasteboard to link editor if available.

### LINKED ISSUE #250: Insert Link: Prefill URL field with clipboard contents if appropriate
It would be nice if the editor was smart enough to recognize when the clipboard contains a URL, and pre-fill that field when inserting a link.  The common flow is to open the browser, copy a link, and then switch back to the WordPress app to insert it in a post.

SergioEstevao:
Fixes #250 

How to test:

- Open Safari and open a site
- Tap on share icon on Safary and select the option Copy
- Jump to the Aztec demo app
- tap on the link button in a area where a link doesn't exist
- check that the link URL is pre-filled with the link copied from Safari

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -3,7 +3,7 @@ import Aztec
 import Gridicons
 import Photos
 import UIKit
-
+import MobileCoreServices
 
 class EditorDemoController: UIViewController {
     static let margin = CGFloat(20)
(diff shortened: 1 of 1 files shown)


diegoreymendez:
@jleandroperez - Taking over the review, with full evilness. :)

jleandroperez:
@diegoreymendez thanks for bearing evilness in mind!!

(I mean, thanks for taking over!!)

PR #269, opened 2017-02-13
Refactor HTML parsing to NSAttributedString to use formatters.

SergioEstevao:
This PR refactor the HTML parsing code to use formatters for all the HTML elements.

How to test:
 - Run the demo app
 - Check that all the HTML was correctly parsed to the Visual mode
 - Run the unit tests.


jleandroperez:
Looking good sir!

:shipit:

PR #270, opened 2017-02-14
Add possibility to add an overlay icon on top of attachments.

SergioEstevao:
How to test:

 - Start the demo app
 - Scroll to an image or insert one
 - tap on it and see if the new Pencil icon shows correctly
 - Tap out and see if it goes away

### DIFF
--- a/Aztec/Classes/TextKit/TextAttachment.swift
+++ b/Aztec/Classes/TextKit/TextAttachment.swift
@@ -67,6 +67,25 @@ open class TextAttachment: NSTextAttachment
         }
     }
 
+    /// An image to display overlaid on top of the image has an action icon.
+    ///
+    open var overlayImage: UIImage? {
+        willSet {
+            if newValue != overlayImage {
+                glyphImage = nil
+            }
+        }
+    }
+
+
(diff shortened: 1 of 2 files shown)


Aztec/Classes/TextKit/TextAttachment.swift
```diff
@@ -222,9 +241,19 @@ open class TextAttachment: NSTextAttachment
             path.stroke()
         }
 
+        var imagePadding: CGFloat = 0
+        if let overlayImage = overlayImage {
+            UIColor.white.set()
+            let center = CGPoint(x: origin.x + (size.width / 2.0), y: origin.y + (size.height / 2.0))
+            let path = UIBezierPath(arcCenter: center, radius: overlayImage.size.width * 2.0/3.0, startAngle: 0, endAngle: CGFloat.pi * 2, clockwise: true)
+            path.stroke()
+            overlayImage.draw(at: CGPoint(x: center.x - (overlayImage.size.width / 2.0), y: center.y - (overlayImage.size.height / 2.0)))
```
jleandroperez: My 2 cents: we could really use some rounding for the *x / y* positions (otherwise the asset might look blurry!)

PR #271, opened 2017-02-14
Group toolbar elements on the iPad.

SergioEstevao:
This PR, groups the toolbar format elements by category, so you have the following the groups

 - Media
 - Character attributes: strong, em, underline, strikethrough
 - Paragraph attributes: blockquote, list ordered, list unordered
 - Link
 - Code

This is only visible on non .compact screen sizes.

How to test:
 - Run the demo app
 - On the app test the different orientations and multitask modes
 - Test also on the iPhone normal and plus to see if all works ok.


diegoreymendez:
Ping @SergioEstevao, since this PR has feedback.

SergioEstevao:
@astralbodies I've been looking at this, but still couldn't find out why the iPhone+ does this. The code looks correct and it's returning the right toolbar for the size, but for some reason it get's the wrong layout... Will continue to investigate.

SergioEstevao:
@astralbodies ready for another look, it looks the responder chain needs to be update to force an update of the toolbar.

Issue #276, opened 2017-02-15
Missing space around formatted text


SergioEstevao:
With this HTML code:
`It has <strong>bold text</strong> and <em>italic text</em> and <del>strikethrough text</del> and <span style="text-decoration: underline;">underlined text</span>.`

 We are getting this result:

![aztec-ios-paragraphs2](https://cloud.githubusercontent.com/assets/651601/22978619/56aa8c7e-f38b-11e6-868c-6cf27ea7fca4.png)

PR #280, opened 2017-02-15
Check for space on start of the text when doing sanitise.

SergioEstevao:
This PR adds checks for empty space on the start of text block.

How to text:

Insert this text on the source code view:
```It has <strong>bold text</strong> and <em>italic text</em> and <del>strikethrough text</del> and <span style="text-decoration: underline;">underlined text</span>.```

Then switch to visual mode and see if the space between bold and italic and strikethrough text are all correct.

### DIFF
--- a/Aztec/Classes/Libxml2/Converters/In/InNodeConverter.swift
+++ b/Aztec/Classes/Libxml2/Converters/In/InNodeConverter.swift
@@ -145,10 +145,9 @@ extension Libxml2.In {
         private func sanitize(_ text: String) -> String {
 
             let hasAnEndingSpace = text.hasSuffix(String(.space))
-            let hasAnStartingSpace = text.hasPrefix(String(.space))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Aztec/Classes/Libxml2/Converters/In/InNodeConverter.swift [hunk 1]
```diff
@@ -144,7 +144,8 @@ extension Libxml2.In {
 
         private func sanitize(_ text: String) -> String {
 
-            let hasAnEndingSpace = text[text.index(before: text.endIndex)] == Character(.space)
+            let hasAnEndingSpace = text.hasSuffix(String(.space))
+            let hasAnStartingSpace = text.hasPrefix(String(.space))
```
diegoreymendez: The spelling is off: "An" -> "A"

diegoreymendez:
This PR fixes #276

PR #287, opened 2017-02-16
More tag

### LINKED ISSUE #275: Implement the More Tag
At the moment the more tag doesn't have a visual representation.
It should be parsed and displayed visually.

SergioEstevao:
Fixes #275 

This PR implements the visualisation of the more tag using a custom attachment.

Notes: 
  - this still doesn't provide a way to insert more tag in the visual editor, it only parses them from HTML.
  - the more tag needs to be exactly like this: ```<!--more-->```. This is case sensitive and space sensitive like Calypso.

How to test:
 - Run the demo app
 - Check that in the sample content the more tag is presented correctly
 - Remove it on the visual mode and check if it updates the DOM
 - Add more tags

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -26,6 +26,10 @@ class HMTLNodeToNSAttributedString: SafeConverter {
         self.defaultFontDescriptor = defaultFontDescriptor
     }
 
+    private lazy var defaultAttributes: [String: Any] = {
+        let defaultFont = UIFont(descriptor: self.defaultFontDescriptor, size: self.defaultFontDescriptor.pointSize)
+        return [NSFontAttributeName: defaultFont]
+    }()
     // MARK: - Conversion
 
     /// Main conversion method.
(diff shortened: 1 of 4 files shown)


Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -97,6 +98,13 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     /// - Returns: the converted node as an `NSAttributedString`.
     ///
     fileprivate func convertCommentNode(_ node: CommentNode, inheritingAttributes inheritedAttributes: [String:Any]) -> NSAttributedString {
+        if node.comment == "more" {
+            var attributes = inheritedAttributes;
+            let moreAttachment = MoreAttachment()
+            moreAttachment.message = NSAttributedString(string: NSLocalizedString("MORE", comment:"Text for the center of the   more divider"), attributes: defaultAttributes)
```
diegoreymendez: There are extra spaces here in the `comment` parameter in the `NSLocalizedString` call.

Aztec/Classes/TextKit/MoreAttachment.swift
```diff
@@ -0,0 +1,86 @@
...
+    override open func attachmentBounds(for textContainer: NSTextContainer?, proposedLineFragment lineFrag: CGRect, glyphPosition position: CGPoint, characterIndex charIndex: Int) -> CGRect {
+
+        let padding = textContainer?.lineFragmentPadding ?? 0
+        let width = lineFrag.width - padding * 2
+        let height:CGFloat = 44.0
+
+        return CGRect(origin: CGPoint.zero, size: CGSize(width: width, height: height))
+    }
+
+    fileprivate func invalidateLayout(inTextContainer textContainer: NSTextContainer?) {
```
diegoreymendez: This doesn't seem to be used.  And as a side note I don't think the attachment should deal with this anyway so I'd recommend to remove it.

PR #294, opened 2017-02-17
Fixes editing around the more tag.

### LINKED ISSUE #293: Editing around the "more" comment tag is broken.
Editing around the "more" comment tag causes the text to be inserted at odd positions in the HTML tree.

diegoreymendez:
<h3>Description:</h3>

Fixes the DOM synchronization when editing text around the "more" tag.

Fixes #293.

<h3>Testing:</h3>

1. Run the app.
2. Place the cursor right before of after the "more" tag.
3. Insert text.
4. Switch to HTML and back.
5. Make sure everything's ok (styles can be slightly off but that's unrelated).

### DIFF
--- a/Aztec/Classes/Libxml2/DOM/ElementNode.swift
+++ b/Aztec/Classes/Libxml2/DOM/ElementNode.swift
@@ -235,10 +235,10 @@ extension Libxml2 {
         ///
         /// - Returns: an array of pairs of child nodes and their ranges in child coordinates.
         ///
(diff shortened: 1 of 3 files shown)


SergioEstevao:
Still not working correctly if you put the cursor after the more tag. It insert on the incorrect place.
Where is my sample html `Before<!--more/>After`

diegoreymendez:
Ready for round 2.

SergioEstevao:
Working well for more, but still failing for the ```<hr/>``` tag, here is an example HTML: ```Before<hr/>After```

PR #299, opened 2017-02-19
Check for attachment and insert image if needed in DOM.

SergioEstevao:
This PR adds code to detect image attachments and applies/insert img elements to the DOM

### How to test

- Start the demo app
- Tap on the insert image button
- Wait for image to be inserted
- Switch to HTML mode
- Check that the img tag is present

### DIFF
--- a/Aztec/Classes/Libxml2/DOMString.swift
+++ b/Aztec/Classes/Libxml2/DOMString.swift
@@ -384,6 +384,12 @@ extension Libxml2 {
                 self?.applyElement(.u, spanning: range)
             }
         }
+
+        func applyImage(imageURL: URL?, spanning range:NSRange) {
+            performAsyncUndoable { [weak self] in
+                self?.setImageURLInDOM(imageURL, forRange: range)
+            }
+        }
         
         // MARK: - Styles to HTML elements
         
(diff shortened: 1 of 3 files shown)


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -355,6 +355,10 @@ open class TextStorage: NSTextStorage {
                 let targetStyle = targetValue as? NSNumber
 
                 processUnderlineDifferences(in: domRange, betweenOriginal: targetStyle, andNew: sourceStyle)
+            case NSAttachmentAttributeName:
+                if let attachment = sourceValue as? TextAttachment {
+                    dom.applyImage(imageURL: attachment.url ,spanning: domRange)
```
diegoreymendez: The spacing before and after the comma is off.

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -384,6 +384,12 @@ extension Libxml2 {
                 self?.applyElement(.u, spanning: range)
             }
         }
+
+        func applyImage(imageURL: URL?, spanning range:NSRange) {
```
diegoreymendez: Document the method.

diegoreymendez:
Fixes part of #304

PR #302, opened 2017-02-20
Fix blockquote formatting

SergioEstevao:
How to test:
 - Start the demo app
 - Open the empty demo
 - Write a line with some text
 - Tap on the blockquote option
 - check if blockquote is applied visually
 - switch to html mode
 - Check that blockquote tag is present
 - switch to visual mode
 - Tap on the blockquote option
 - Check the blockquote is removed visually
 - Switch to html mode
 - Check that the html doesn't have a blockquote

### DIFF
--- a/AztecTests/TextStorageTests.swift
+++ b/AztecTests/TextStorageTests.swift
@@ -149,4 +149,21 @@ class TextStorageTests: XCTestCase
         XCTAssertEqual(attachment.url, url)
         XCTAssertEqual(html, "<img src=\"https://wordpress.com\" class=\"alignleft size-medium\">")
     }
+
+    func testBlockquoteToggle() {
+        let storage = TextStorage()
+        storage.append(NSAttributedString(string: "Apply a blockquote"))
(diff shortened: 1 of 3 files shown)


AztecTests/TextStorageTests.swift
```diff
@@ -149,4 +149,21 @@ class TextStorageTests: XCTestCase
         XCTAssertEqual(attachment.url, url)
         XCTAssertEqual(html, "<img src=\"https://wordpress.com\" class=\"alignleft size-medium\">")
     }
+
+    func testBlockquoteToggle() {
+        let storage = TextStorage()
+        storage.append(NSAttributedString(string: "Apply a blockquote"))
+
+        let _ = storage.toggleBlockquote(storage.rangeOfEntireString)
```
diegoreymendez: To be honest I think `toggleBlockquote()` should have a `@discardableResult`.

diegoreymendez:
Fixes part of #304

PR #303, opened 2017-02-20
Fix link formatting

SergioEstevao:
How to test:
 - Open demo app
 - Open the demo with content
 - Check if link is being parsed correctly from sample contentn
 - Tap on link
 - Check if URL and text of the link is correct
 - Change url and text and tap update link
 - Check if visual changes are correct
 - Switch to HTML and see if html is correct.
 - Switch to visual mode again
 - Tap on the link
 - Choose remove link
 - Check if visual mode is correct
 - Switch to html mode and check if html is correct.

### DIFF
--- a/Aztec/Classes/Libxml2/DOMString.swift
+++ b/Aztec/Classes/Libxml2/DOMString.swift
@@ -410,7 +410,7 @@ extension Libxml2 {
         func applyLink(_ url: URL?, spanning range: NSRange) {
             var attributes: [Libxml2.Attribute] = []
             if let url = url {
-                attributes.append(Libxml2.StringAttribute(name:"href", value: url.absoluteString))
+                attributes.append(Libxml2.StringAttribute(name: HTMLLinkAttribute.Href.rawValue, value: url.absoluteString))
             }
             self.apply(element: .a, spanning: range, attributes: attributes)
         }
--- a/Aztec/Classes/Libxml2/DOMString.swift
+++ b/Aztec/Classes/Libxml2/DOMString.swift
@@ -298,6 +298,16 @@ extension Libxml2 {
                 self?.removeBlockquoteSynchronously(spanning: range)
             }
         }
+
+        /// Disables link from the specified range
+        ///
+        /// - Parameter range: the range to remove
+        ///
+        func removeLink(spanning range: NSRange) {
+            performAsyncUndoable { [weak self] in
+                self?.removeSynchronously(element:.a, at: range)
+            }
+        }
         
         // MARK: - Remove Styles: Synchronously
         private func removeSynchronously(element: StandardElementType, at range: NSRange) {
@@ -331,7 +341,7 @@ extension Libxml2 {
         
         // Apply Styles
         
-        func apply(element: StandardElementType, spanning range: NSRange, attributes: [Attribute] = []) {
+        fileprivate func apply(element: StandardElementType, spanning range: NSRange, attributes: [Attribute] = []) {
             performAsyncUndoable { [weak self] in
                 self?.applyElement(element, spanning: range, attributes: attributes)
             }
@@ -391,6 +401,20 @@ extension Libxml2 {
             }
         }
 
+        /// Applies a link to the specified range
+        ///
+        /// - Parameters:
+        ///   - url: the url to link to
+        ///   - range: the range to apply the link
+        ///
+        func applyLink(_ url: URL?, spanning range: NSRange) {
+            var attributes: [Libxml2.Attribute] = []
+            if let url = url {
+                attributes.append(Libxml2.StringAttribute(name:"href", value: url.absoluteString))
+            }
+            self.apply(element: .a, spanning: range, attributes: attributes)
+        }
+
         /// Replaces the characteres in the specified range for a an img element pointing to the provided URL
         ///
         /// - Parameters:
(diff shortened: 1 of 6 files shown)


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -493,6 +497,31 @@ open class TextStorage: NSTextStorage {
...
+        let removeStyle = originalStyle != nil && newStyle == nil
+
+        if addStyle {
+            var attributes: [Libxml2.Attribute] = []
+            if let url = newStyle {
+                attributes.append(Libxml2.StringAttribute(name:"href", value: url.absoluteString))
+            }
+            dom.apply(element: .a, spanning: range, attributes: attributes)
+        } else if removeStyle {
+            dom.remove(element: .a, at: range)
```
diegoreymendez: Following up on my observation above, can we change this to:

```swift
dom.removeLink(spanning: range)
```

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -493,6 +497,31 @@ open class TextStorage: NSTextStorage {
...
+
+        let addStyle = originalStyle == nil && newStyle != nil
+        let removeStyle = originalStyle != nil && newStyle == nil
+
+        if addStyle {
+            var attributes: [Libxml2.Attribute] = []
+            if let url = newStyle {
+                attributes.append(Libxml2.StringAttribute(name:"href", value: url.absoluteString))
+            }
+            dom.apply(element: .a, spanning: range, attributes: attributes)
```
diegoreymendez: `TextStorage` was designed so it doesn't really need to know about HTML elements.  The mapping between styles and DOM elements is withing `DOMString`'s responsibility.

Can we replace this call with a more generic:

```swift
dom.applyLink(url: newUrl, spanning: range)
```

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -341,6 +331,11 @@ extension Libxml2 {
         
         // Apply Styles
         
+        func apply(element: StandardElementType, spanning range: NSRange, attributes: [Attribute] = []) {
```
diegoreymendez: This should really be private, as the mapping between styles and DOM elements should be handled by `DOMString` itself.  Read my comments below.

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -493,6 +497,31 @@ open class TextStorage: NSTextStorage {
...
 
+    /// Processes differences in link styles, and applies them to the DOM in the specified
+    /// range.
+    ///
+    /// - Parameters:
+    ///     - range: the range in the DOM where the differences must be applied.
+    ///     - originalStyle: the original link URL object if any.
+    ///     - newStyle: the new link URL object if any.
+    ///
+    private func processLinkDifferences(in range: NSRange, betweenOriginal originalStyle: URL?, andNew newStyle: URL?) {
```
diegoreymendez: Consider renaming the params to better represent the data:

`betweenOriginal originalStyle` -> `betweenOriginal  originalUrl`
`andNew newStyle` -> `andNew newUrl`

SergioEstevao:

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -341,6 +341,11 @@ extension Libxml2 {
         
         // Apply Styles
         
+        fileprivate func apply(element: StandardElementType, spanning range: NSRange, attributes: [Attribute] = []) {
+            performAsyncUndoable { [weak self] in
```
diegoreymendez: Only the public methods should have queueing.

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -396,6 +401,20 @@ extension Libxml2 {
...
+        /// Applies a link to the specified range
+        ///
+        /// - Parameters:
+        ///   - url: the url to link to
+        ///   - range: the range to apply the link
+        ///
+        func applyLink(_ url: URL?, spanning range: NSRange) {
+            var attributes: [Libxml2.Attribute] = []
+            if let url = url {
+                attributes.append(Libxml2.StringAttribute(name:"href", value: url.absoluteString))
```
diegoreymendez: Replace "href" with `HTMLLinkAttribute.Href.rawValue`.

Aztec/Classes/Libxml2/DOMString.swift
SergioEstevao: The idea here is to avoid repetitition of the performaAsyncUndoable everywhere. If in the future we change the way we do this, we can change in a single place.

Aztec/Classes/Libxml2/DOMString.swift
SergioEstevao: good one.

Aztec/Classes/Libxml2/DOMString.swift
diegoreymendez: The problem if you try to centralize it here, is that you can't call this method from methods that already queued the operation (or you can, but you'd be queueing twice).

The class was designed so that only public methods would queue, an all internal operations would just be synchronous.  This is to avoid double queueing in a very simple and clear way.

I'd prefer us to stick to the existing architecture for now, and if we want to change this, do it separately.

SergioEstevao:

diegoreymendez:
Fixes part of #304

PR #306, opened 2017-02-22
Fixes addition of images.

### LINKED ISSUE #305: Inserting text after images is not working.
Inserting text after images is not working and is causing data loss.

**Steps to reproduce:**

1. Add an image to an empty post.
2. Type some text after it.
3. When the image finishes downloading, switch to HTML mode and back.

You'll notice any text after the image is lost.

diegoreymendez:
Fixes adding images to the DOM so that only differences will be applied.  Also removes some weird issues that were causing images to be duplicated.

Fixes #305 

**How to test the previous issue:**

1. Launch the editor in 0.5a4.
2. Open the empty editor demo.
3. Insert and image and type just "a".
4. Switch to HTML mode.

You'll notice the image tag is duplicated.

### DIFF
--- a/Aztec/Classes/Libxml2/DOMString.swift
+++ b/Aztec/Classes/Libxml2/DOMString.swift
@@ -242,7 +242,7 @@ extension Libxml2 {
         /// - Parameters:
         ///     - range: the range to remove the style from.
         ///
-        func removeImageURL(spanning range: NSRange) {
+        func removeImage(spanning range: NSRange) {
             performAsyncUndoable { [weak self] in
                 self?.removeImageSynchronously(spanning: range)
             }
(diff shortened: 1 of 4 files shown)


SergioEstevao:
@diegoreymendez has you can see above, some tests are failing.

Beside this, there is still an issue when you try to insert text before an img.

Try this:
 - Start the empty demo 
 - Insert a single image
 - Add text after the image
 - Switch to html and see the text is inserted after the image.
 - Switch to visual
 - Insert text before the image
 - switch back to html
  - see that text was added in the wrong place.

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -6,7 +6,10 @@ extension Libxml2 {
     /// a string.
     ///
     /// Any requests made to this class are performed in its own queue (sometimes synchronously,
-    /// sometimes asynchronously).
+    /// sometimes asynchronously).  Public methods are resopnsible for queueing requests, while all
```
SergioEstevao: Small typo: responsible

PR #310, opened 2017-02-22
Issue/182 implement headers

### LINKED ISSUE #182: Add support for different font sizes.
Add support for different font sizes.

SergioEstevao:
Fixes #182 

This PR implement supports for the H1 tag.

How to test:
 - Open the demo app
 - Check that the header was applied correctly
 - Check if editing inside the header works correctly
 - Check if editing around the header (before, after) works correctly
 - Add a new line on the end of header text and check if style gets broken
 - Add a new line on the middle of the header and check if style is kept on the new line

@diegoreymendez can you have a look I'm having issue with editing around the header, the mapping looks incorrect.

### DIFF
--- a/Aztec/Classes/Formatters/HeaderFormatter.swift
+++ b/Aztec/Classes/Formatters/HeaderFormatter.swift
@@ -1,0 +1,59 @@
+import Foundation
+import UIKit
+
+class HeaderFormatter: ParagraphAttributeFormatter {
+
+    let elementType: Libxml2.StandardElementType = .h1
+
+    let placeholderAttributes: [String : Any]?
+
+    init(placeholderAttributes: [String : Any]? = nil) {
+        self.placeholderAttributes = placeholderAttributes
+    }
+
+    func apply(to attributes: [String : Any]) -> [String: Any] {
+        var resultingAttributes = attributes
+        let newParagraphStyle = ParagraphStyle()
+        if let paragraphStyle = attributes[NSParagraphStyleAttributeName] as? NSParagraphStyle {
+            newParagraphStyle.setParagraphStyle(paragraphStyle)
+        }
+        newParagraphStyle.paragraphSpacing += Metrics.defaultIndentation
+        newParagraphStyle.paragraphSpacingBefore += Metrics.defaultIndentation
+        resultingAttributes[NSParagraphStyleAttributeName] = newParagraphStyle
+
+        if let font = attributes[NSFontAttributeName] as? UIFont {
+            let newFont = font.withSize(36)
+            resultingAttributes[NSFontAttributeName] = newFont
+        }
+
+        return resultingAttributes
+    }
+
+    func remove(from attributes:[String: Any]) -> [String: Any] {
+        var resultingAttributes = attributes
+        let newParagraphStyle = ParagraphStyle()
+        guard let paragraphStyle = attributes[NSParagraphStyleAttributeName] as? ParagraphStyle,
+            paragraphStyle.headerLevel != 0 else {
+            return resultingAttributes
+        }
+        newParagraphStyle.setParagraphStyle(paragraphStyle)
+        newParagraphStyle.paragraphSpacing -= Metrics.defaultIndentation
+        newParagraphStyle.paragraphSpacingBefore -= Metrics.defaultIndentation
+        resultingAttributes[NSParagraphStyleAttributeName] = newParagraphStyle
+
+        if let font = attributes[NSFontAttributeName] as? UIFont {
+            let newFont = font.withSize(16)
+            resultingAttributes[NSFontAttributeName] = newFont
+        }
+
+        return resultingAttributes
+    }
+
+    func present(in attributes: [String : Any]) -> Bool {
+        if let paragraphStyle = attributes[NSParagraphStyleAttributeName] as? ParagraphStyle {
+            return paragraphStyle.headerLevel != 0
+        }
+        return false
+    }
+}
+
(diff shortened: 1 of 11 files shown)


diegoreymendez:
@SergioEstevao - I'm tackling the latest issue you found in #312.

diegoreymendez:
@SergioEstevao - I think we can separate this from the issue in #312, since that's caused by a problem with `UITextView`, and is not related to this.

SergioEstevao:
I still need to had a small fix when we delete a new line after a heading. I will ask for review after that.

SergioEstevao:

Aztec/Classes/Formatters/HeaderFormatter.swift
```diff
@@ -0,0 +1,87 @@
...
+            case .h2: return 24
+            case .h3: return 21
+            case .h4: return 16
+            case .h5: return 14
+            case .h6: return 11
+            }
+        }
+    }
+
+    let elementType: Libxml2.StandardElementType = .h1
```
diegoreymendez: This should be removed (I just checked and it's not being used).

Aztec/Classes/Formatters/HeaderFormatter.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    var headerLevel: HeaderType = .h1
```
diegoreymendez: A few changes here:

1. Change `var` to `let`.
2. Remove the `= .h1` assignment, since the initializer is taking care of it.

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -435,6 +448,23 @@ extension Libxml2 {
             rootNode.replaceCharacters(inRange: range, withElement: elementDescriptor)
         }
         
+        private static let headerLevels: [StandardElementType] = [.h1, .h2, .h4, .h5, .h6]
```
diegoreymendez: We're placing constants at the top of the files mostly, so it would be good to to the same for this one.

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -435,6 +448,23 @@ extension Libxml2 {
...
+        private func elementTypeForHeaderLevel(_ headerLevel: Int) -> StandardElementType? {
+            if headerLevel < 1 && headerLevel > DOMString.headerLevels.count {
+                return nil
+            }
+            return DOMString.headerLevels[headerLevel - 1]
+        }
+
+
+        func applyHeader(_ headerLevel:Int, spanning range:NSRange) {
```
diegoreymendez: This should be inside the `// MARK: - Apply Styles` group.

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -435,6 +448,23 @@ extension Libxml2 {
             rootNode.replaceCharacters(inRange: range, withElement: elementDescriptor)
         }
         
+        private static let headerLevels: [StandardElementType] = [.h1, .h2, .h4, .h5, .h6]
+        private func elementTypeForHeaderLevel(_ headerLevel: Int) -> StandardElementType? {
```
diegoreymendez: This should not be in the `// MARK: - Images` group.

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -3,8 +3,13 @@ import UIKit
 
 open class ParagraphStyle: NSMutableParagraphStyle {
 
+    enum EncodingKeys: String {
```
diegoreymendez: Make this private, since it's not necessary to access it from the outside.

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -96,15 +111,29 @@ open class ParagraphStyle: NSMutableParagraphStyle {
         copy.setParagraphStyle(originalCopy)
         copy.textList = textList
         copy.blockquote = blockquote
+        copy.headerLevel = headerLevel
 
         return copy
     }
 
+    open override var hash: Int {
```
diegoreymendez: What is the hash being used for?  I wasn't able to spot the need for this.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -605,6 +644,37 @@ open class TextView: UITextView {
...
+        var afterString = String(.newline)
+
+        if afterRange.endLocation < storage.length {
+            afterString = storage.attributedSubstring(from: afterRange).string
+        }
+
+        return afterString == String(.newline) && storage.isStartOfNewLine(atLocation: location)
+    }
+
+    @discardableResult func ensureRemovalOfSingleLineParagraphAttributes(insertedText text: String, at range: NSRange) -> Bool {
```
diegoreymendez: I have to admit I'm not able to quicly grasp why this is necessary by just reading the code.

Can you document this method, including information and possibly an example of why this is needed?

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -605,6 +644,37 @@ open class TextView: UITextView {
         return false
     }
 
+    private func shouldRemoveSingleLineParagraphAttributes(insertedText text: String, at location: Int) -> Bool {
```
diegoreymendez: I have to admit I'm not able to quicly grasp why this is necessary by just reading the code.

Can you document this method, including information and possibly an example of why this is needed?

Aztec/Classes/TextKit/ParagraphStyle.swift
SergioEstevao: This is just good practice when you override the isEqual method, this can be used by the system to insert things in Dictionaries so if we have extra properties those should be taken in account.

SergioEstevao:

diegoreymendez:
This actually fixes #324, and partially fixes #182.

PR #311, opened 2017-02-22
Fixes some issues with text fragmentation.

diegoreymendez:
Fixes some issues that were causing `TextNode` fragmentation.

**How to test:**

1. Open the empty editor demo.
2. Type in "Abcdef"
3. In Xcode, place a breakpoint in `ElementNode:1289`
4. In the editor, type 1 last character "g".

Xcode should break.  Type:

`po self`

Make sure the whole text (minus the last "g") is in a single text node.

### DIFF
--- a/Aztec/Classes/Libxml2/DOM/ElementNode.swift
+++ b/Aztec/Classes/Libxml2/DOM/ElementNode.swift
@@ -1304,8 +1304,7 @@ extension Libxml2 {
 
                         let offset = intersection.location == 0 ? 0 : 1
 
-                        let textNode = TextNode(text: string, editContext: editContext)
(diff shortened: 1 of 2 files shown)


SergioEstevao:
Still seeing issues.
For example switch to html mode and insert this text: 
```<blockquote>quote</blockquote><p>Next Line</p>```
Then switch to visual mode and try to tap at the start of the text, just before quote...
All the styling is lost and the blockquote goes away.

diegoreymendez:
@SergioEstevao - This PR is only to fix the `TextNode` fragmentation issue.  We can continue testing for other issues in your main PR.

The blockquote issue you mention is also present in `develop` so we can treat that separately.

PR #315, opened 2017-02-22
Add HR to the list of elements that cannot have children.

SergioEstevao:
This PR is just a precaution, I didn't find a scenario where this happen, but just in case.


diegoreymendez:
This change is necessary to prevent problems with text insertion.  Nice job.

PR #316, opened 2017-02-23
Set the typing attributes to whatever is on the selection range when the html is set.

### LINKED ISSUE #312: Style loss issues
As reported by @SergioEstevao:

```
<blockquote>quote</blockquote><p>Next Line</p>
```

I tested this issue and I can confirm it in 0.5a4.

It may not be limited to `blockquotes`.

SergioEstevao:
Fixes #312

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -255,7 +255,8 @@ open class TextView: UITextView {
         if storage.length > 0 && selectedRange.location < storage.length {
             typingAttributes = storage.attributes(at: selectedRange.location, effectiveRange: nil)
         }
-        delegate?.textViewDidChangeSelection?(self)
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -252,6 +252,10 @@ open class TextView: UITextView {
         font = defaultFont
         
         storage.setHTML(html, withDefaultFontDescriptor: font!.fontDescriptor)
+        if storage.length > 0 && selectedRange.location < storage.length {
+            typingAttributes = storage.attributes(at: selectedRange.location, effectiveRange: nil)
+        }
+        delegate?.textViewDidChangeSelection?(self)
```
diegoreymendez: If this is only for the toolbar, consider calling `TextViewFormattingDelegate`'s `textViewCommandToggledAStyle()` instead.

PR #317, opened 2017-02-23
Change attachment interface.

SergioEstevao:
This PR changes the attachment interface when doing copy/paste to allows integration on the main WP app and other apps.

For details of why the change check this [PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/6758) on the main WordPress app.

How to test:
 - Start the demo app
 - Use the empty demo
 - Switch to the Photos a[[
 - Copy an image on the Photos app
 - Switch to the Aztec demo
 - Paste the image
 - Check if the integration worked.

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -579,11 +579,14 @@ extension EditorDemoController: TextViewMediaDelegate
         return Gridicon.iconOfType(.image)
     }
     
-    func textView(_ textView: TextView, urlForImage image: UIImage) -> URL {
+    func textView(_ textView: TextView, urlForAttachment attachment: TextAttachment) -> URL {
         
         // TODO: start fake upload process
-        
-        return saveToDisk(image: image)
+        if let image = attachment.image {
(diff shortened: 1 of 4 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -579,11 +579,14 @@ extension EditorDemoController: TextViewMediaDelegate
...
-    func textView(_ textView: TextView, urlForImage image: UIImage) -> URL {
+    func textView(_ textView: TextView, urlForAttachment attachment: TextAttachment) -> URL {
         
         // TODO: start fake upload process
-        
-        return saveToDisk(image: image)
+        if let image = attachment.image {
+            return saveToDisk(image: image)
+        } else {
+            return URL(string: "placeholder://")!
```
diegoreymendez: I'm a bit confused about this URL we're returning here.  Can you explain why this is necessary?

SergioEstevao:
If for some reason we don't have an image we need to put an placeholder URL... this probably can be better done by changing the interface of the protocol to allow the return of an optional URL.

PR #321, opened 2017-02-24
Removes the EditableNode protocol.

### LINKED ISSUE #307: Remove EditableNode protocol
<h3>Description:</h3>

Replace the `EditableNode` protocol with a data-oriented replacement.

<h3>Rationale:</h3>

The `EditableNode` protocol is an attempt to provide a clear separation between nodes that have inline editable text content (such as bold nodes), and nodes that don't (such as `comment` nodes: their content may or may not be editable inline).

This is not a very good approach since:

- It relies on specific node subclasses being marked as `EditableNode`, which makes it not very flexible for customization (as in the case of comment nodes).
- Some `ElementNode` instances, such as `img` or `br`, don't have textual content that can be edited inline, while some others do.  This makes it impossible to mark the `ElementNode` class as either editable or not.

<h3>Suggested Implementation:</h3>

A data oriented approach would solve both issues.  We could add a method to query whether a certain node has inline editable text content or not, and eventually offer hooks to customize it.

This could be provided to the DOM tree through the `EditContext` object.

### LINKED ISSUE #319: Crash on second line of empty editor
Follow these steps to reproduce the crash:

1. Open the empty editor demo
2. Write "Hello World"
3. Press enter once
4. Type CMD + B
5. Try typing anything

Crash!

### LINKED ISSUE #325: Pasting text on a selection is not working properly.
[Here's a video](https://cloudup.com/c3uIljRssSG) showing the problem.

Steps to reproduce:

1. Open the empty editor.
2. Paste "Hello there!
3. Select the word "Hello"
4. Paste again while the selection is active.
5. Switch to HTML mode and back.

You'll notice the text order has been changed.

diegoreymendez:
<h3>Description</h3>

Removes the EditableNode protocol.  All nodes should have proper editing logic in place now.

Closes #307.
Fixes #319.
Fixes #325.

<h3>Testing:</h3>

**Test 1:**

Review the new unit test.  Run the unit tests.

**Test 2:**
1. Open the empty editor demo
2. Write "Hello World"
3. Press enter once
4. Type CMD + B
5. Try typing anything
Unfortunately we'll have to rely on unit tests for this one, as there's no simple scenario we can test against.

**Test 3:**

1. Open the empty editor.
2. Insert an image.
3. Edit text around the image.
4. See that it doesn't crash.


diegoreymendez:
This is on hold until I can fix the crashes reported in #319.

SergioEstevao:
@diegoreymendez Tests are failing.

diegoreymendez:
@SergioEstevao - Ready for another look.  The unit test was actually broken.

diegoreymendez:
@SergioEstevao - Added a new unit tests, and fixed the one that was still failing.

PR #327, opened 2017-02-25
Updates the caret position after pasting into the editor.

### LINKED ISSUE #301: When pasting, the cursor does not move to the end of the pasted content
Copy-paste something into the editor. You'll see that the cursor stays in front of the text.

diegoreymendez:
Fixes #301

To test:

Just paste something and make sure the caret position is moved to the end of the pasted content.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -123,7 +123,7 @@ open class TextView: UITextView {
     }
 
 
-    // MARK: - Intersect copy paste operations
+    // MARK: - Intercept copy paste operations
 
     open override func copy(_ sender: Any?) {
         super.copy(sender)
@@ -143,6 +143,7 @@ open class TextView: UITextView {
         string.loadLazyAttachments()
 
         storage.replaceCharacters(in: selectedRange, with: string)
+        delegate?.textViewDidChange?(self)
         selectedRange = NSRange(location: selectedRange.location + string.length, length: 0)
     }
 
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -141,7 +141,9 @@ open class TextView: UITextView {
         }
 
         string.loadLazyAttachments()
+
         storage.replaceCharacters(in: selectedRange, with: string)
+        selectedRange = NSRange(location: selectedRange.location + string.length, length: 0)
```
SergioEstevao: We should call to the TextView delegate to say that the selection changed.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -156,6 +158,7 @@ open class TextView: UITextView {
         string.loadLazyAttachments()
 
         storage.replaceCharacters(in: selectedRange, with: string)
+        selectedRange = NSRange(location: selectedRange.location + string.length, length: 0)
```
SergioEstevao: We should call to the TextView delegate to say that the selection changed.

SergioEstevao: I went back to double check, and the textViewDidChangeSelection is actually being called when you change the selecteRange, what is not being called is textViewDidChange.

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: I went back to double check, and the textViewDidChangeSelection is actually being called when you change the selecteRange, what is not being called is textViewDidChange.

diegoreymendez:
@SergioEstevao - Ready for another look.

PR #329, opened 2017-02-27
Experience with a custom html output.

SergioEstevao:
@diegoreymendez this an experience just for you to have a look on a possible custom HTML output, and possible trade-offs.

The big advantages I see here are:

 - Total customisation of output
 - Full control of indentation and new lines
 - Tags are closed correctly when they are single element: HR, BR, IMG, etc...
 - No dependency of libxml methods for output.

Some disadvantages:
 - We may need to need implement something for HTML entities likes &NBSP and things like < > & and other.
 - A bit more code on our side, but not that much.


diegoreymendez:
@SergioEstevao - Can we close this for now?  Or do you want to keep it open?

SergioEstevao:
I will close the PR but keep the branch alive, just for future reference.

PR #330, opened 2017-02-28
Improve attributes formatter interface

SergioEstevao:
This PR improves and cleanups the AttributeFormatter interface by allowing the following:

 - Allows apply and remove of formatters attributes directly, not only by toggling
 - Shares more common code between character formatters and paragraph formatters
 - Cleans up the code on TextView that was using toggles instead of direct apply and remove of attributes.

How to test:
 - Run the demo app
 - Interact with different styles and check if they update correctly
 - Try applying and remove newlines inside and around paragraph styles (list, block quotes, headers)
 - Check if unit tests pass

### DIFF
--- a/Aztec/Classes/Formatters/AttributeFormatter.swift
+++ b/Aztec/Classes/Formatters/AttributeFormatter.swift
@@ -53,7 +53,7 @@ protocol AttributeFormatter {
 
     func applicationRange(for range: NSRange, in text: NSAttributedString) -> NSRange
 
-    func worksInEmtpyRange() -> Bool
+    func worksInEmptyRange() -> Bool
 }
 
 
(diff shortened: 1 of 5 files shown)


diegoreymendez:
Found a problem:

1. Run the empty editor demo.
2. Enable bold
3. Type "Line 1"
4. Enable italic
5. Type "Line 2"
6. Switch to HTML and back.

You'll notice there's some style loss.  It seems like only the last selected style is maintained.

SergioEstevao:
@diegoreymendez Is this only happening on this branch?

SergioEstevao:
After you type Line 1 you press Enter?

diegoreymendez:
[Here are the steps to reproduce](https://cloudup.com/cl743fHc7MP).  Sorry I didn't post this with the report.

diegoreymendez:
As discussed this issue was not introduced by this PR and is currently present in `develop`.

PR #338, opened 2017-03-02
More unicode fixing

SergioEstevao:
This PR adds some more conversion methods between location on NSString and index on String.

How to test:
 - Run the unit tests.
 - Run the demo app
 - Add and remove emoji:
   -  On the end start of styles, 
   - On the end/start of lists and blockquotes
   - On the end of lines.
 - Check if no crashes occur

### DIFF
--- a/Aztec/Classes/Extensions/String+RangeConversion.swift
+++ b/Aztec/Classes/Extensions/String+RangeConversion.swift
@@ -24,4 +24,8 @@ extension String
             else { return nil }
         return from
     }
+
+    func isLastValidLocation(_ location: Int) -> Bool {
+        return index(before: endIndex) == indexFromLocation(location)
+    }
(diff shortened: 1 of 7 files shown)


Aztec/Classes/Extensions/String+RangeConversion.swift
```diff
@@ -1,19 +1,49 @@
...
         return from ..< to
                 
     }
+
+    func indexFromLocation(_ location: Int) -> String.Index? {
+        guard
+            let from16 = utf16.index(utf16.startIndex, offsetBy: location, limitedBy: utf16.endIndex),
+            let from = from16.samePosition(in: self)
+            else { return nil }
+        return from
```
jleandroperez: NICE!!!

PR #339, opened 2017-03-02
Fix: Broken Styling after Emoji

### LINKED ISSUE #337: Bold / Italics: Breaking after Emoji
1. Launch the empty editor demo
2. Enter a few emoji's
3. Hit newline
4. Enter some text
5. Press Italics or Bold

As a result, the editor will *not* toggle the formatting.

cc @SergioEstevao

jleandroperez:
### Details:
Internally, each time the user enters an emoji, iOS will automatically apply a custom font to the Emoji-Range.

Problem is that immediately after the Emoji, the "Original" font is not restored *exactly* the way it was. In some scenarios, the mechanism we use to toggle Bold / Italics would break. In other scenarios (such as in WPiOS), iOS would switch us over to **Courier New** after an Emoji.

The goal of this PR is to implement a simple workaround: whenever we detect that the Active Font's family (available via the Typing Atttributes) won't match with the one in our Default Font parameter, we'll fall back to the **Default Font**.

### Scenario: Style After Emoji
1. Launch the empty editor demo
2. Enter a few emoji's
3. Press Italics or Bold

Verify that the text entered immediately after the emoji gets formatted as expected.

### Scenario: Style After Header-Emoji
1. Launch the empty editor demo
2. Press the **H** eader format bar button
3. Enter a few emoji's
4. Press Italics or Bold

Verify that the text entered immediately after the emoji gets formatted as expected (Heading size / Italics or Bold).

### Scenario: Moving the cursor around
1. Launch the empty editor demo
2. Enter some text
3. Enter a few emoji's
4. Move the cursor back to the text
5. Move the cursor after the last emoji
6. Toggle Bold / Italics

Verify that the text entered gets formatted as expected.

Closes #337

Needs Review: @SergioEstevao @diegoreymendez 
Thanks in advance!


jleandroperez:
@diegoreymendez @SergioEstevao gentlemen, ready for review. Any feedback would be more than welcome.

I realize this is not the cutest fix, but... tested under all of the scenarios i could come up with, and it's definitely looking fine!!

SergioEstevao:
@jleandroperez 
At the moment this is breaking Header formatting because the path reverts to the default font every time it detects anything that doesn't; match the default font.
I think an improved solution could be if we detect an Emoji font, loop back the index position of the caret until you find a non-emoji font and copy the style that is there.
This probably should limited to a single line/paragraph.

jleandroperez:
@SergioEstevao thanks for catching that up!!! Workaround updated (+ test scenarios updated). 
Ready for another look!

SergioEstevao:
@jleandroperez It's looking better but now I'm seeing another issue:

When you select the header style, and then tap some emojis and follow up with regular text. If you then tap on the header style to toggle the emoji, switch but the regular text not.

jleandroperez:
@SergioEstevao thanks for spotting that one! Since that one is also happening in develop, i've opened #358 to deal with it.

jleandroperez:
Thanks for the review sir!!!

PR #344, opened 2017-03-08
Removing NSLinkAttribute when entering text after or before the anchor

### LINKED ISSUE #251: There's no way to interrupt adding a link.
When adding a link, the style is never interrupted.  There's simply no mechanism to do so.

<H3>Steps to reproduce</h3>

1. Open the empty editor.
2. Insert a link.
3. Continue typing... you'll notice there's no way to interrupt the link style.

jleandroperez:
### Details:
In this PR we're implementing a simple mechanism that will remove the `NSLinkAttribute` whenever we're not editing text **within** a Link's Anchor. 

Meaning that: inserting a link, and entering some text immediately after it will no longer further extend the linkified range.

Closes #251

### Testing:
1. Launch Aztec
2. Insert a link
3. Add some text. Verify it does not get linkified
4. Move the cursor at the beginning of the link
5. Enter `Some Random Text` and verify it does not get linkified
6. Verify that any newly entered text within the Link's Anchor gets properlylinkified

Needs Review: @SergioEstevao 
Thanks in advance sir!


jleandroperez:
Thanks for the review @SergioEstevao!!

PR #346, opened 2017-03-08
Fix attachment difference processing.

SergioEstevao:
How to test:

 - Use the empty editor demo
 - Insert one image
 - then just after the first insert a second image.
 - Switch to html and check both img tag are there.


diegoreymendez:
This is related to #347 .

Issue #331, opened 2017-02-28
Padding around images


0nko:
When adding multiple images to a post, they appear right on top of each other. Also, it's difficult to position the cursor before or after an image.

![image](https://cloud.githubusercontent.com/assets/1522856/23411433/9521b7c4-fdd2-11e6-9534-583bf3e5c48b.png)

SergioEstevao:
@diegoreymendez  to create spacing we will need to add a space or new lines character around images. 
I don't think this should be the default of the insertImage in TextView but when reacting to the command of the toolbar we could insert the image and add space around it. 

This can be even more critical when we are inserting multiples images in one go in a post.

diegoreymendez:
@SergioEstevao - Any thoughts on either adding white padding to images programmatically or modifying the layout code to do just that?

I think we should be able to achieve a good result without adding actual characters.

SergioEstevao:
We can add white space around the image when generating the glyph, so that
visually it will look better.

But the thing is that for the layout system that attachment always take the
full space.

So it's always going to be hard for the user to put cursor on the middle of
two images.

PR #351, opened 2017-03-09
Padding between images

SergioEstevao:
Fixes: #331 

This PR adds the possibility to have padding between images/attachments and other content. This is mainly to avoid that images inserted one after the other are "stitched" together.
This PR also improves interaction when tapping in between the images to allow the cursor to stay in the middle of two images.

How to test:
 - Start the empty demo app
 - Insert three images one after the other.
 - Check if they have margin between them vertically.
 - Tap in the middle empty area between the images and check if the caret position is made between the two images.


SergioEstevao:

jleandroperez:
@SergioEstevao MUCH better!!

:shipit:

Issue #324, opened 2017-02-24
Add support for the header tags.


diegoreymendez:
Add support for the header tags.

diegoreymendez:
Fixed by #310.

PR #356, opened 2017-03-14
Improve headers toolbar

SergioEstevao:
Related to #324 

This PR adds support for selection of the header type to apply when using the format toolbar.
It uses a custom input view that allows to choose what header the user wants to apply.

How to test:
 - Start the demo app
 - Start the editor with content
 - Tap on one of the paragraphs with header
 - Tap on the Header button on the toolbar
 - Check if the selection made on the input view is correct
 - Change the style and see if the style is correct.

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -273,8 +273,10 @@ class EditorDemoController: UIViewController {
         // Filter multiple header identifier to single header identifier
         identifiers = identifiers.map({ (identifier) -> FormattingIdentifier in
             switch identifier {
-            case .header1, .header2, .header3, .header4, .header5, .header6: return .header
-            default: return identifier
+            case .header1, .header2, .header3, .header4, .header5, .header6:
+                return .header
+            default:
+                return identifier
             }
         })
         toolbar.selectItemsMatchingIdentifiers(identifiers)
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -356,7 +356,7 @@ extension EditorDemoController : Aztec.FormatBarDelegate {
             showImagePicker()
         case .sourcecode:
             toggleEditingMode()
-        case .header:
+        case .header1, .header2, .header3, .header4, .header5, .header6:
             toggleHeader()
         }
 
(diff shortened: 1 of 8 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -268,6 +270,13 @@ class EditorDemoController: UIViewController {
         } else {
             identifiers = richTextView.formatIdentifiersForTypingAttributes()
         }
+        // Filter multiple header identifier to single header identifier
+        identifiers = identifiers.map({ (identifier) -> FormattingIdentifier in
+            switch identifier {
+            case .header1, .header2, .header3, .header4, .header5, .header6: return .header
+            default: return identifier
```
jleandroperez: can we move the `return` (in both cases) to a newline, to aid readability?

Example/Example/EditorDemoController.swift
```diff
@@ -406,9 +416,53 @@ extension EditorDemoController : Aztec.FormatBarDelegate {
...
+            .header3 : .h3,
+            .header4 : .h4,
+            .header5 : .h5,
+            .header6 : .h6,
+        ]
+        for (key,value) in mapping {
+            if identifiers.contains(key) {
+                return value
+            }
+        }
```
jleandroperez: (No need to change anything!!!) This one can be also synthesized as:

```
        for (key,value) in mapping where identifiers.contains(key) {
            return value
        }
```

Example/Example/OptionsTableView.swift
```diff
@@ -0,0 +1,61 @@
+
+
+import UIKit
```
jleandroperez: I (think) in the other sources the `import` goes in the first line?

Example/Example/OptionsTableView.swift
```diff
@@ -0,0 +1,61 @@
...
+        super.init(frame: frame, style: .plain)
+        self.delegate = self
+        self.dataSource = self
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+    }
+
+
```
jleandroperez: Extra new lines!

Example/Example/EditorDemoController.swift
SergioEstevao: Will do.

SergioEstevao:

jleandroperez:
:shipit:

PR #357, opened 2017-03-14
Refactor attachment touch/selection to be part of the text view.

SergioEstevao:
This PR refactors the attachment touch detection to use delegate callback from the UITextView instead of relaying on the client to do that detection. 

This avoid code duplication in users of the library, it also hides implementation details from users of the library ( no need to know about margin, moving cursor and other implementation details).

How to test:
 - Start a new post
 - Insert multiple images
 - Tap on the image and the text outside and see if the interaction works correctly
 - Tap on the white space between images and check if the cursor/caret moves to the correct position.

@diegoreymendez mind to have a look on this? The only main architecture question I have is if the new new delegate methods should be part of the current media delegate or I should create a new one. What do you think?


diegoreymendez:
I'm seeing some strange problems with:

1. Several images all showing the "TAP TO EDIT" message simultaneously.
2. The editor moving down when I tap on an image.

Video here: https://cloudup.com/cQtMebML_5L

SergioEstevao:

SergioEstevao:
@diegoreymendez ready for another look

PR #362, opened 2017-03-15
Implement insertion of HR elements.

### LINKED ISSUE #360: Support insertion of Horizontal Rulers
Add an option on the toolbar and an API to allow insertion of Horizontal Rulers, hr HTML elements.

SergioEstevao:
Closes #360

This PR adds the possibility of adding HR elements using the toolbar and by a function call on RichTextView.

- How to test:
 - Start the demo app
 - Insert the Horizontal Ruler using the toolbar icon
 - Check if the ruler is inserted visually and in code.
 - Remove the ruler visually and check if html code is updated correctly.

### DIFF
--- a/Aztec/Classes/Libxml2/DOMString.swift
+++ b/Aztec/Classes/Libxml2/DOMString.swift
@@ -493,6 +493,23 @@ extension Libxml2 {
             rootNode.replaceCharacters(inRange: range, withElement: elementDescriptor)
         }
 
+        /// Applies horizontal ruler to the specified range.
+        ///
+        /// - Parameters:
+        ///     - range: the range to apply the style to.
+        ///
+        func insertHorizontalRuler(at range: NSRange) {
+            performAsyncUndoable { [weak self] in
+                self?.insertHorizontalRulerSynchronously(at: range)
+            }
+        }
+
+        private func insertHorizontalRulerSynchronously(at range: NSRange) {
+            let elementDescriptor = ElementNodeDescriptor(elementType: .hr)
+
+            rootNode.replaceCharacters(inRange: range, withElement: elementDescriptor)
+        }
+
         // MARK: - Styles to HTML elements
         
         /// Applies a standard HTML element to the specified range.
--- a/Aztec/Classes/TextKit/TextStorage.swift
+++ b/Aztec/Classes/TextKit/TextStorage.swift
@@ -177,7 +177,7 @@ open class TextStorage: NSTextStorage {
         
         attributedString.enumerateAttribute(NSAttachmentAttributeName, in: fullRange, options: []) { (object, range, stop) in
 
-            guard let object = object else {
+            guard let object = object, !(object is LineAttachment) else {
                 return
             }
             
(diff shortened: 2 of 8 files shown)


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -185,11 +185,15 @@ open class TextStorage: NSTextStorage {
...
                 return
             }
-            
+
             guard let attachment = object as? NSTextAttachment else {
                 assertionFailure("We expected a text attachment object.")
                 return
             }
+
+            if object is LineAttachment {
```
diegoreymendez: You could move this to the guard in line 180, like this:

```swift
guard let object = object, !(object is LineAttachment) else {
    return
}
```

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -356,6 +360,13 @@ open class TextStorage: NSTextStorage {
 
             processUnderlineDifferences(in: domRange, betweenOriginal: sourceStyle, andNew: targetStyle)
         case NSAttachmentAttributeName:
+            if sourceValue is LineAttachment || targetValue is LineAttachment {
```
diegoreymendez: We should avoid checking this here with an `if` block.

It's a precondition of this method (`processAttributesDifference`) that either the source value or the target value must be non-nil (in the same way it's a precondition for the other attributes).

The rationale is that `enumerateAttributeDifferences()` is responsible of only signaling real differences: an attribute that doesn't change is NOT a difference, and would signal that the method is not working right.

If you want to make sure that this precondition is true, an assertion would be more appropriate.  For instance we could check at the start of this method:

```swift
assert(sourceValue != nil || targetValue != nil)
```

Or you could add a more specific assertion for your attribute.

```swift
assert(sourceValue is LineAttachment || targetValue is LineAttachment)
```

* My mistake for not documenting in this method that it is actually a precondition.

SergioEstevao: But the check here is not for nil but to see if we are handling an LineAttachment instead of an TextAttachment (images).

SergioEstevao:
@diegoreymendez can you give it another look?

Aztec/Classes/TextKit/TextStorage.swift
SergioEstevao: @diegoreymendez I added a test and it looks it's working ok, do you mind to have a look on it?

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -493,6 +493,23 @@ extension Libxml2 {
             rootNode.replaceCharacters(inRange: range, withElement: elementDescriptor)
         }
 
+        /// Applies horizontal ruler to the specified range.
+        ///
+        /// - Parameters:
+        ///     - range: the range to apply the style to.
+        ///
+        func insertHorizontalRuler(at range: NSRange) {
```
diegoreymendez: Can we change the name of this method to better depict that it's replacing existing contents?

Insert makes it sound as if this method didn't affect existing content.

`replaceWithHorizontalRuler(_ range: NSRange)` maybe?

SergioEstevao:

PR #363, opened 2017-03-15
Remove all Attachments API

jleandroperez:
### Details:
This PR implements a new API, that would nuke all of the TextAttachments. This is required in order to address [this WordPress iOS Issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/6827).

### Testing:
Please, just run the unit tests, and verify everything looks smooth.

@SergioEstevao although we've discussed about only removing the links tied up to the `Old Selected Site`, since WordPress.com uses links that look like... `lanteanartest.files.wordpress.com`, i'm afraid we can't really match by domain.

For Mark I, let's replicate the old behavior, and build on top of that!

Needs Review: @SergioEstevao 
Thanks in advance!!

### DIFF
--- a/Aztec/Classes/TextKit/TextStorage.swift
+++ b/Aztec/Classes/TextKit/TextStorage.swift
@@ -725,6 +725,23 @@ open class TextStorage: NSTextStorage {
         }
     }
 
+    /// Removes all of the TextAttachments from the storage
+    ///
+    open func removeTextAttachments() {
+        var ranges = [NSRange]()
+        enumerateAttachmentsOfType(TextAttachment.self) { (attachment, range, _) in
+            ranges.append(range)
+        }
+
+        var delta = 0
+        for range in ranges {
+            let corrected = NSRange(location: range.location - delta, length: range.length)
+            replaceCharacters(in: corrected, with: NSAttributedString(string: ""))
+            delta += range.length
+        }
+    }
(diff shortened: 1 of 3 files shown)


diegoreymendez:
Any reason why we can't just upload ALL images to the new site?

In my humble opinion when moving a post from one site to another, we should pick one of these options:

- Upload all images regardless of their original URL.
- Just use the original URLs without moving anything.
- Let the user pick which images to upload, and which images to leave in their original locations.

It's fine to remove all images if we want a temporary solution, but I think for our users it's the worst option to have.

jleandroperez:
@diegoreymendez i haven't really given it much thought (just attacking the original issue, which is, replicate the Hybrid's behavior).

Since the images have been already uploaded, and the original site might either have rules preventing hotlinking, i think a proper solution is to **RE** upload the images (since they've been already uploaded).

Nevertheless, for a v1, let's go ahead and recreate the current behavior, if it's okay with everyone.

diegoreymendez:
Definitely OK with having a temporary solution.

I'll open a discussion issue for this, though.

jleandroperez:
Thanks @diegoreymendez!!

SergioEstevao:
:shipit:

jleandroperez:
Thanks for the review!

PR #370, opened 2017-03-16
Fix crash when trying to delete all content

SergioEstevao:
This PR tries to fix the crash on delete all text on the textview when a comment is present in the HTML.


SergioEstevao:
Closed in favour of #374

PR #376, opened 2017-03-20
Add the possibility to parse and create spans for colors.

### LINKED ISSUE #375: Add parsing support for color style.
Related to #285 

Add support for parsing of color style in HTML.

Ex: ```<span style="color: #ff0000;">Colors</span>```

SergioEstevao:
Closes #375 

This PR adds support for parsing of span elements with the color attribute.

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -211,6 +211,15 @@ class HMTLNodeToNSAttributedString: SafeConverter {
             attributeValue = attachment
         }
 
+        if node.isNodeType(.span) {
+            if let elementStyle = node.valueForStringAttribute(named: "style") {
+                let styles = parseStyle(style: elementStyle)
+                if !styles.isEmpty, let colorString = styles["color"] {
+                    attributeValue = UIColor(hexString: colorString)
+                }
+            }           
+        }
+
         for (key, formatter) in elementToFormattersMap {
             if node.isNodeType(key) {
                 if let standardValueFormatter = formatter as? StandardAttributeFormatter,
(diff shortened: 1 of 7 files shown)


Aztec/Classes/Extensions/UIColor+Parsers.swift
```diff
@@ -0,0 +1,27 @@
+import UIKit
+
+extension UIColor {
+
+    /// Creates a color based on a hexString. If the string is not a valid hexColor it return nils
```
diegoreymendez: nils -> nil

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -240,6 +249,22 @@ class HMTLNodeToNSAttributedString: SafeConverter {
         .h3: HeaderFormatter(headerLevel: .h3),
         .h4: HeaderFormatter(headerLevel: .h4),
         .h5: HeaderFormatter(headerLevel: .h5),
-        .h6: HeaderFormatter(headerLevel: .h6)
+        .h6: HeaderFormatter(headerLevel: .h6),
+        .span: ColorFormatter()
```
diegoreymendez: Does this mean we won't be able to parse other CSS attributes in spans? (like bold, italic, etc?)

Aztec/Classes/Extensions/UIColor+Parsers.swift
```diff
@@ -0,0 +1,27 @@
...
+    /// Example of colors: #FF0000, #00FF0000
+    ///
+    convenience init?(hexString: String) {
```
diegoreymendez: Can you add a few unit tests for this method?  At least for RGB 24 bit, which I believe is the most common one.

diegoreymendez:
Nice work @SergioEstevao.  Let me know when ready.

Aztec/Classes/Extensions/UIColor+Parsers.swift
SergioEstevao: Will do!

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
SergioEstevao: Humm, good catch, I think I will change this a bit and made styles be parsed in a different way. 

Ideally they can be applied to any element so I will check styles in a dedicated pass just for the style attribute in elements. 
 
I will update the PR.

diegoreymendez: @SergioEstevao - If we want to leave it like this for this PR I'm fine with it.  I just wanted to raise awareness that we may need to change this a bit moving forward.

diegoreymendez: IMHO the fix for this should probably be separated, so that we can split the work atomically.

SergioEstevao:

PR #377, opened 2017-03-21
Fixes 350 crash when using visual only element

### LINKED ISSUE #350: Crash: Strikethrough + Blockquote
### Steps:
1. Launch the empty editor
2. Toggle Strikethru
3. Toggle Blockquote

Crash onscreen!
cc @diegoreymendez 

screen shot 2017-03-09 at 2 00 12 pm

SergioEstevao:
Fixes #350 

This PR at the moment fixes the crash by removing the visual only attributes to the zero width space.

This space is used when applying paragraph attributes that need a character in order to be visible.

@diegoreymendez I think we are missing a visual to DOM mapping somewhere in the process but I'm struggling to find here to do it.

### DIFF
--- a/Aztec/Classes/NSAttributedString/VisualOnlyElementFactory.swift
+++ b/Aztec/Classes/NSAttributedString/VisualOnlyElementFactory.swift
@@ -31,7 +31,7 @@ class VisualOnlyElementFactory {
     func zeroWidthSpace(inheritingAttributes inheritedAttributes: [String:Any]? = nil) -> NSAttributedString {
         var attributes = inheritedAttributes ?? [String:Any]()
 
-        attributes[VisualOnlyAttributeName] = VisualOnlyElement.zeroWidthSpace.rawValue
+        //attributes[VisualOnlyAttributeName] = VisualOnlyElement.zeroWidthSpace.rawValue
 
         return NSAttributedString(.zeroWidthSpace, attributes: attributes)
     }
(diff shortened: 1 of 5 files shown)


Aztec/Classes/NSAttributedString/VisualOnlyElementFactory.swift
```diff
@@ -31,7 +31,7 @@ class VisualOnlyElementFactory {
     func zeroWidthSpace(inheritingAttributes inheritedAttributes: [String:Any]? = nil) -> NSAttributedString {
         var attributes = inheritedAttributes ?? [String:Any]()
 
-        attributes[VisualOnlyAttributeName] = VisualOnlyElement.zeroWidthSpace.rawValue
+        //attributes[VisualOnlyAttributeName] = VisualOnlyElement.zeroWidthSpace.rawValue
```
diegoreymendez: Why are we removing the visual-only attribute here?

diegoreymendez:
As a side note: mapping DOM to Visual is impossible... why would you need that though?  The DOM should not be aware of anything visual.

Aztec/Classes/NSAttributedString/VisualOnlyElementFactory.swift
SergioEstevao: This fixes the issue, but of course it's not ideal, that is why I wanted some guidance on the visual to Dom mappings.

SergioEstevao:
Closing in favor of #378

PR #384, opened 2017-03-21
Comment Tag Support

### LINKED ISSUE #367: Comment Tag Support
### Details:
- [x] Add a visual representation for the `<!-- xxx -->` tag.
- [x] Support a customization mechanism for Apps to render certain comments differently.
- [x] Nuke MoreAttachment, since it'll no longer be needed
- [x] Reimplement MoreAttachment Rendering by means of a CommentAttachment customization

jleandroperez:
### Details:
- Replaced **MoreAttachment** with **CommentAttachment**
- Updated **TextAttachment**'s delegate property, for consistency's sake
- **TextStorageAttachmentsDelegate** got a couple new methods
- **TextStorage**'s Attachments Processing loop got refactored!
- **TextStorage** .attachmentsDelegate is now IUO. Added a couple asserts here and there, since it's mandatory!
- Implemented custom Renders for CommentAttachments: `CommentAttachmentRender` and `MoreAttachmentRender`.
- And of course, updated the Editor Demo!

### Scenario: More Comment
1. Launch Aztec > Emtpy Editor
2. Insert a More Comment Tag, by means of the format bar
3. Verify that a full width attachment shows up, with dashed lines: `----- MORE -----`
4. Switch to HTML
5. Switch back to rich text mode. Verify that the attachment is still there.

### Scenario: Regular Comment
1. Launch Aztec > Emtpy Editor
2. Insert a `<!--comment-->` tag, in **raw html mode**.
3. Switch to Rich Mode
4. Verify that a `[COMMENT]` attachment shows up, wherever the attachment is.

### Scenario: Media Attachment
Please, perform a smoke test. Verify that Media Attachments still work as expected.

Closes #367

Needs Review: @diegoreymendez @SergioEstevao 
Thanks in advance!!!


Example/Example/MoreAttachmentRender.swift
```diff
@@ -0,0 +1,96 @@
...
+//
+
+import Foundation
+import UIKit
+import Aztec
+
+
+// MARK: - MoreAttachmentRender: Renders More Comments!
+//
+class MoreAttachmentRender {
```
SergioEstevao: Couldn't this class be part of the open API library to allows users of the library to use it, and avoid code replication.

Example/Example/CommentAttachmentRender.swift
```diff
@@ -0,0 +1,86 @@
...
+//  Created by Jorge Leandro Perez on 3/22/17.
+//  Copyright © 2017 Automattic Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import Aztec
+
+
+class CommentAttachmentRender {
```
SergioEstevao: Couldn't this class be part of the open API library to allows users of the library to use it, and avoid code replication.

Example/Example/CommentAttachmentRender.swift
```diff
@@ -0,0 +1,86 @@
+//
```
SergioEstevao: This class comments should be removed.

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -171,53 +197,52 @@ open class TextStorage: NSTextStorage {
     /// - Returns: the preprocessed string.
     ///
     fileprivate func preprocessAttachmentsForInsertion(_ attributedString: NSAttributedString) -> NSAttributedString {
-        
+        assert(attachmentsDelegate != nil)
```
SergioEstevao: Maybe out of scope of this PR, but I think we should some kind of default implementation of the attachments Delegate that just work, and then allow users of the library to customize/override it.

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -40,9 +42,33 @@ protocol TextStorageAttachmentsDelegate {
...
+    /// Provides the Bounds required to represent a given attachment, within a specified line fragment.
+    ///
+    /// - Parameters:
+    ///     - storage: The storage that is requesting the bounds.
+    ///     - attachment: CommentAttachment about to be rendered.
+    ///     - lineFragment: Line Fragment in which the glyph would be rendered.
+    ///
+    /// - Returns: Rect specifying the Bounds for the comment attachment
+    ///
+    func storage(_ storage: TextStorage, boundsForComment attachment: CommentAttachment, with lineFragment: CGRect) -> CGRect
```
SergioEstevao: This protocol is starting to feel a bit mixed up and over complex, between images and comments, maybe two different protocols?

Example/Example/MoreAttachmentRender.swift
diegoreymendez: Can we rename these classes to `Renderer` (the one that renders) instead of `Render` (a render operation)?

Aztec/Classes/TextKit/TextStorage.swift
diegoreymendez: Just to add some info on this, @jleandroperez had a pretty nice idea for improving this mechanism in the future.

I'd encourage us as a team to review the architecture of all our delegates as an improvement.  Maybe after closed beta.

Example/Example/CommentAttachmentRender.swift
jleandroperez: Thanks for spotting that!!

jleandroperez:
@diegoreymendez @SergioEstevao Update!

- Renamed Comment + More **Render** -> **Renderer**
- Removed few new class'ess headers

Are you guys both okay with promoting **CommentAttachmentRenderer** and **MoreAttachmentRenderer** to Aztec Level, instead of Example App Level?

@SergioEstevao RE: new / different internal delegates, idea to perhaps (vastly) improve the API is to do something as follows:

```
extension TextView {
   open func register(attachment: TextAttachmentRenderer)
}
```

```
protocol TextAttachmentRenderer {
    func canRender(attachment: TextAttachment) -> Bool
    func render(attachment: TextAttachment) -> UIImage
    func bounds(attachment: TextAttachment, textView: TextView, lineFragment: CGRect) -> CGRect
}
```

We'll likely address such refactor after the internal beta, but just wanted to note that the TextStorage / TextAttachnent / TextView delegate chain are a temporary solution, and a quite generic implementation is around the corner.

Thanks in advance!

diegoreymendez:
> Are you guys both okay with promoting CommentAttachmentRenderer and MoreAttachmentRenderer to Aztec Level, instead of Example App Level?

I'd rather us wrap this PR as it is first, since the decision on that is not clear.

SergioEstevao:
@jleandroperez If we wrap up the PR like this, will the More option on the toolbar and from html parsing will stop working on the main app? or we still something visually?

PR #386, opened 2017-03-23
Pre tag support

SergioEstevao:
This PR add support for the pre tag on the html.

At the moment when the tag is detected the following is done:
 - Text is show with monospace font.
 - Reading of text from HTML doesn't clean up extra space or newlines.

Need help with: 
 - How to avoid that newlines added on visual mode, are converted to ```<br/>```


SergioEstevao:
@diegoreymendez I was able to sort out my questions above, so this is now ready for review.

SergioEstevao:

Aztec/Classes/Libxml2/DOM/Data/TextNode.swift
```diff
@@ -216,6 +216,10 @@ extension Libxml2 {
         // MARK: - EditableNode
 
         func append(_ string: String) {
+            if !shouldSanitizeText() {
```
diegoreymendez: No biggie but this could look nicer with a `guard`.

```swift
guard shouldSanitizeText() else {
    append(sanitizedString: string)
    return
}
```

Aztec/Classes/Libxml2/DOM/Data/TextNode.swift
```diff
@@ -250,7 +254,22 @@ extension Libxml2 {
             }
         }
 
+        func shouldSanitizeText() -> Bool {
```
diegoreymendez: Can we add documentation to explain why this method is necessary?  While on topic it really makes sense, but if you grab this on its own, it may not be too obvious at first.

Also, for readability and code reusability it may be nice to change this method to:

```swift
func shouldSanitizeText() -> Bool {
    return hasParent(.pre)
}
```

Aztec/Classes/Libxml2/DOM/Data/TextNode.swift
```diff
@@ -250,7 +254,22 @@ extension Libxml2 {
...
+                if let node = parentNode, node.name == Libxml2.StandardElementType.pre.rawValue {
+                    return false
+                }
+                parentNode = parentNode?.parent
+            }
+            return true
+        }
+
         override func replaceCharacters(inRange range: NSRange, withString string: String, preferLeftNode: Bool) {
+            if !shouldSanitizeText() {
```
diegoreymendez: Again, a `guard` would be nice here.

SergioEstevao:

PR #393, opened 2017-03-29
Extend title view for big titles

### LINKED ISSUE #385: Title does not stay with end of text while typing if it extends past screen width
img_3578

The full title is "What Steven Ballmer and I have in common", yet the text field doesn't show the end of the title while typing.

@diegoreymendez I'm only assigning you since you're leading iOS and know how to prioritize this. Feel free to pass this off to anyone.

SergioEstevao:
Closes #385
Relate to this: https://github.com/wordpress-mobile/WordPress-iOS/issues/6824

On this PR I'm converting the Title to an UITextView to allow multiple lines and adjusting the size of it depending of the text inserted.

I'm also adding a placeholder label because UITextView doesn't have placeholder labels!

How to test:
 - Start the demo app
 - On any of the demos app check the title placeholder shows correctly
 - Enter text on the title that expand to multiple lines and see if layout adapts correctly.

screen shot 2017-03-29 at 14 07 10


Example/Example/EditorDemoController.swift
```diff
@@ -158,22 +170,42 @@ class EditorDemoController: UIViewController {
 
     }
 
+    // MARK: - Title and Title placeholder position methods
+    func updateTitleHeight() {
+        let sizeThatShouldFitTheContent = titleTextField.sizeThatFits(CGSize(width:self.view.frame.size.width - ( 2 * Constants.margin), height: CGFloat.greatestFiniteMagnitude))
```
jleandroperez: Maybe move the CGSize instantiation one line upwards, to aid readability? 
+ No need to use `self` nor `size`. `view.frame.width` should do the trick!

Example/Example/EditorDemoController.swift
```diff
@@ -158,22 +170,42 @@ class EditorDemoController: UIViewController {
...
+        let insets = titleTextField.textContainerInset
+        titleHeightConstraint.constant = max(sizeThatShouldFitTheContent.height, titleTextField.font!.lineHeight + insets.top + insets.bottom)
+    }
+
+    func updateTitlePlaceholderVisibility() {
+        self.titlePlaceholderLabel.isHidden = !self.titleTextField.text.isEmpty
```
jleandroperez: No need to use `self`

jleandroperez:
Only thing i'm not really sure about is: since we don't currently support "Scrolling the Title" to get it out of the way, in the multiline title scenario, users might end up with something like this:

screen shot 2017-03-29 at 5 34 07 pm

Note that "Line 1 ... 3" cannot be scrolled, and the edition area ends up being quite tight (this is on the iPhone SergioEstevao!).

Thoughts on this one? (Maybe we wanna delay this a bit until we can also implement the scrollabe title behavior, or merge it, but cap the mechanism to just one line, temporarily?)

SergioEstevao:
@jleandroperez I will tackle the title view scrolling out of the way next. Just after this PR is in.

PR #397, opened 2017-04-04
Unknown HTML Support

### LINKED ISSUE #369: Unknown HTML support in visual mode.
At the moment if the content has a non supported tag like:

<span>
<table>
<video>

We are simple ignoring tag and not showing anything on the visual representation.

Following the Android implementation we should detect these tags and show a ? character (or maybe something like the name of the tag. Ex: ```<table>```. And allow users visualise it's content.

jleandroperez:
### Details:
This PR implements **Unknown HTML** Support: tags not recognized by Aztec (defined in **EditContext**'s `knownElements` property) will be rendered, in visual mode, as `[HTML]`.

The way "Unknown HTML" attachments get displayed is 100% customizable. We've also revamped the mechanism used to render "Comment Attachments" + "More Attachment". In order to keep the Delegate's fingerprint low, a new protocol was introduced: `TextAttachmentImageProvider`.

Host Application gets to register "Attachment Image Providers", and customize the way attachments will be displayed.

Please, verify the following test scenarios pass, along with the new Unit Tests.

Closes #369 

Thanks in advance!!
Needs Review: @diegoreymendez  @SergioEstevao 
🔥 

#### Scenario: Unsupported HTML
1. Switch to Raw Edition Mode
2. Enter the following snippet:
   `AZTEC <unknown>payload<something>else</something></unknown> LINE`
3. Switch to Visual Edition Mode
4. Verify that the unknown HTML shows up as `[HTML]`
5. Try editing the document. Add text on top / at the bottom
6. Switch to HTML mode
7. Verify everything looks as expected

#### Scenario: Paste
1. Switch to Raw Edition Mode
2. Enter the following snippet:
   `Head <unknown>payload</unknown> Tail`
3. Switch to Visual Mode
4. Select the `[HTML]` text, and copy it
5. Try pasting the `[HTML]` (visual attachment) into any position / range
6. Verify that the raw HTML looks as expected

#### Scenario: Replace
1. Switch to Raw Edition Mode
2. Enter the following snippet:
   `<unknown>payload</unknown> LINE <second>test</second>`
3. Switch to Visual Mode
4. Select the **first** `[HTML]` and copy it
5. Select the **Second** `[HTML]` and paste over it
6. Verify that the raw HTML looks as follows:
   `<unknown>payload</unknown> LINE <unknown>payload</unknown>`

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -54,16 +54,15 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     ///
     fileprivate func convert(_ node: Node, inheritingAttributes attributes: [String:Any]) -> NSAttributedString {
 
-        if let textNode = node as? TextNode {
+        switch node {
+        case let textNode as TextNode:
             return convertTextNode(textNode, inheritingAttributes: attributes)
-        } else if let commentNode = node as? CommentNode {
+        case let commentNode as CommentNode:
             return convertCommentNode(commentNode, inheritingAttributes: attributes)
-        } else {
-            guard let elementNode = node as? ElementNode else {
-                fatalError("Nodes can be either text or element nodes.")
-            }
-
+        case let elementNode as ElementNode:
             return convertElementNode(elementNode, inheritingAttributes: attributes)
+        default:
+            fatalError("Nodes can be either text, comment or element nodes.")
         }
     }
 
@@ -112,8 +111,18 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     ///
     /// - Returns: the converted node as an `NSAttributedString`.
     ///
-    fileprivate func convertElementNode(_ node: ElementNode, inheritingAttributes attributes: [String:Any]) -> NSAttributedString {
-        return stringForNode(node, inheritingAttributes: attributes)
+    fileprivate func convertElementNode(_ node: ElementNode, inheritingAttributes attributes: [String: Any]) -> NSAttributedString {
+        guard !node.isSupportedByAztec() else {
+            return stringForNode(node, inheritingAttributes: attributes)
+        }
+
+        let converter = Libxml2.Out.HTMLConverter()
+        let attachment = HTMLAttachment()
+
+        attachment.rootTagName = node.name
+        attachment.rawHTML = converter.convert(node)
+
+        return NSAttributedString(attachment: attachment, attributes: attributes)
     }
 
     // MARK: - Node Styling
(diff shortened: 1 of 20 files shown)


Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -54,16 +54,15 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     ///
     fileprivate func convert(_ node: Node, inheritingAttributes attributes: [String:Any]) -> NSAttributedString {
 
-        if let textNode = node as? TextNode {
+        switch node {
```
SergioEstevao: Love this change!

Aztec/Classes/Libxml2/DOMString.swift
```diff
@@ -471,6 +471,36 @@ extension Libxml2 {
...
+
+        // MARK: - Raw HTML
+
+        /// Replaces the specified range with a given Raw HTML String.
+        ///
+        /// - Parameters:
+        ///   - range: the range to insert the HTML
+        ///   - rawHTML: String representing a raw HTML Snippet, to be converted into Nodes.
+        ///
+        func replace(_ range: NSRange, rawHTML: String) {
```
SergioEstevao: replace(range: withRawHTML:) maybe will be a better signature.

SergioEstevao:
Just some suggestions:

 - instead of writing HTML could we write the name of the node we are representing? For example if it's a table we will write [TABLE]?
 - maybe integrate some kind of callback when a user taps on this attachments, in order in the future to implement some kind of editors or visualisers of the content.

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
jleandroperez: Thank you sir!!

Aztec/Classes/Libxml2/DOMString.swift
jleandroperez: Good call! updating `withComment` as well, for concistency's sake!

jleandroperez:
LOVED the "Display root unknown tag name" suggestion. Implemented, looking nice!

+ Updated DOMString helper methods. RE: Callback mechanism, i'll address that in another PR, i'd love to revisit the delegates we've got, and maybe simplify them.

Ready for another look @SergioEstevao, thanks a loooot!!

SergioEstevao:
One last thing can you add an example on the content.html file? Maybe a table?

jleandroperez:
@SergioEstevao content.html updated! ready for another look!

jleandroperez:
Thanks a lot Sergio!!!

PR #402, opened 2017-04-07
EditContext: Updatting supported tags

jleandroperez:
### Details:
Adding support for the `<del>` tag. Please, verify it does not show up as "Unknown HTML" (`[DEL]`).

Needs Review: @SergioEstevao 
Thanks in advance!!


jleandroperez:
Thanks Sergio!

PR #423, opened 2017-04-12
Add UI tests for simple text formatting

rachelmcr:
This PR adds a suite of UI tests for simple text formatting in the Aztec demo app.

They test each of the text formatting buttons in the toolbar (including the hr and more tags). The tests for lists are currently disabled, pending lists being fully implemented in Aztec.

Pinging @hoverduck and @catehstn for a review. :)

### DIFF
--- a/Example/AztecUITests/AztecUITests.swift
+++ b/Example/AztecUITests/AztecUITests.swift
@@ -117,7 +117,7 @@ class AztecSimpleTextFormattingTests: XCTestCase {
         // Copy link to be auto-filled in URL field
         UIPasteboard.general.string = "https://wordpress.com/"
         app.scrollViews.otherElements.buttons[elementStringIDs.linkButton].tap()
-        app.alerts.buttons.element(boundBy: 3).tap()
+        app.alerts.buttons[elementStringIDs.insertLinkConfirmButton].tap()
         app.buttons[elementStringIDs.sourcecodeButton].tap()
 
         let text = getHTMLContent()
--- a/Example/AztecUITests/AztecUITests.swift
+++ b/Example/AztecUITests/AztecUITests.swift
@@ -17,6 +17,9 @@ class AztecSimpleTextFormattingTests: XCTestCase {
 
         let tablesQuery = app.tables
         tablesQuery.staticTexts[elementStringIDs.emptyDemo].tap()
+
+        let richTextField = app.textViews[elementStringIDs.richTextField]
+        richTextField.tap()
     }
 
     override func tearDown() {
@@ -25,86 +28,86 @@ class AztecSimpleTextFormattingTests: XCTestCase {
     }
 
     func testSimpleBoldText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.boldButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<strong>text</strong>"
+        let expected = "<strong>1</strong>"
         XCTAssertEqual(expected, text)
     }
 
     func testSimpleItalicText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.italicButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<em>text</em>"
+        let expected = "<em>1</em>"
         XCTAssertEqual(expected, text)
     }
 
     func testSimpleUnderlineText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.underlineButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<u>text</u>"
+        let expected = "<u>1</u>"
         XCTAssertEqual(expected, text)
     }
 
     func testSimpleStrikethroughText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.strikethroughButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<del>text</del>"
+        let expected = "<del>1</del>"
         XCTAssertEqual(expected, text)
     }
 
     func testSimpleBlockquoteText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.blockquoteButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<blockquote>text</blockquote>"
+        let expected = "<blockquote>1</blockquote>"
         XCTAssertEqual(expected, text)
     }
 
     // Enable this test after unordered lists are fully implemented
     func testSimpleUnorderedListText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.unorderedlistButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<ul><li>text</li></ul>"
+        let expected = "<ul><li>1</li></ul>"
         XCTAssertEqual(expected, text)
     }
 
     // Enable this test after ordered lists are fully implemented
     func testSimpleOrderedListText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         app.scrollViews.otherElements.buttons[elementStringIDs.orderedlistButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<ol><li>text</li></ol>"
+        let expected = "<ol><li>1</li></ol>"
         XCTAssertEqual(expected, text)
     }
 
     func testSimpleLinkedText() {
-        enterTextInField(text: "text")
+        enterTextInField(text: "1")
         selectAllTextInField()
 
         // Copy link to be auto-filled in URL field
@@ -113,14 +116,11 @@ class AztecSimpleTextFormattingTests: XCTestCase {
         app.alerts.buttons[elementStringIDs.insertLinkConfirmButton].tap()
 
         let text = getHTMLContent()
-        let expected = "<a href=\"https://wordpress.com/\">text</a>"
+        let expected = "<a href=\"https://wordpress.com/\">1</a>"
         XCTAssertEqual(expected, text)
     }
 
     func testHorizontalRuler() {
-        let richTextField = app.textViews[elementStringIDs.richTextField]
-        richTextField.tap()
-
         app.scrollViews.otherElements.buttons[elementStringIDs.horizontalrulerButton].tap()
 
         let text = getHTMLContent()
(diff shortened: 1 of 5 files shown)


rachelmcr:
Also, h/t @SergioEstevao for the accessibility improvements. I used the model from https://github.com/wordpress-mobile/WordPress-iOS/pull/6967 to add accessibility labels and IDs to the Aztec demo app, for use in these tests.

diegoreymendez:
Wow, nice contribution @rachelmcr!  Really appreciated!

I'm adding myself as reviewer mostly to point out a few minor standardization changes I've spotted.

Example/AztecUITests/XCTest+Extensions.swift
```diff
@@ -0,0 +1,45 @@
...
+
+extension XCUIElement {
+    /**
+     Enters text in the field and then selects all entered text
+     - Parameter text: the text to enter into the field
+    */
+    func enterAndSelectText(text: String) -> Void {
+        let app = XCUIApplication()
+
+        self.typeText(text)
```
diegoreymendez: No need for `self.` in this line, and the next one.

Example/AztecUITests/XCTest+Extensions.swift
```diff
@@ -0,0 +1,45 @@
...
+        self.press(forDuration: 1.2)
+        app.menuItems.element(boundBy: 1).tap()
```
diegoreymendez: The UI tests are failing for me, because it seems like the menu item bound by 1 is recognized as one of the auto-correction options, instead of "Select All".

Ony way to work around this issue would be to replace this line with something like:

```swift
app.menuItems["Select All"].tap()
```

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+
+        // In UI tests it is usually best to stop immediately when a failure occurs.
+        continueAfterFailure = false
+        // UI tests must launch the application that they test. Doing this in setup will make sure it happens for each test method.
+        XCUIDevice.shared().orientation = .portrait
+        app = XCUIApplication()
+        app.launch()
+
+        let tablesQuery = app.tables
+        tablesQuery.staticTexts[ elementStringIDs.emptyDemo ].tap()
```
rachelmcr: Thanks! I've fixed these.

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+    func testSimpleBoldText() {
+        richContentTextView = app.textViews[ elementStringIDs.richTextField ]
+        richContentTextView.tap()
+        richContentTextView.enterAndSelectText(text: "text")
+
+        app.scrollViews.otherElements.buttons[ elementStringIDs.boldButton ].tap()
+        app.buttons[ elementStringIDs.sourcecodeButton ].tap()
+
+        htmlContentTextView = app.textViews[ elementStringIDs.htmlTextField ]
+        let text : String = htmlContentTextView.value as! String
```
rachelmcr: Thanks! These are fixed, too.

Example/AztecUITests/XCTest+Extensions.swift
rachelmcr: Got it, thanks!

Example/AztecUITests/XCTest+Extensions.swift
rachelmcr: Argh, text prediction. I was seeing those failures earlier but after some updates stopped seeing them (even with prediction enabled), so I thought it wasn't an issue anymore.

I was trying to avoid relying on the string `"Select All"`, so the tests will work in other languages. I was playing with pasting the text instead of typing, to avoid text prediction issues, so I'm going to try that out and see how it goes.

catehstn: Agree relying on text prediction is not the best - also need to consider what if it's not there.

catehstn: Another option is to turn off autocorrect at the start of the tests.

I don't think XCUI allows us to leave the app and go to settings: http://osxdaily.com/2015/01/06/disable-auto-correct-ios/

But apparently it is a property you can change on a text input: http://stackoverflow.com/questions/11789594/how-to-programmatically-turn-off-auto-correction-in-iphone-sdk

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        app.buttons[elementStringIDs.sourcecodeButton].tap()
+
+        htmlContentTextView = app.textViews[elementStringIDs.htmlTextField]
+        let text: String = htmlContentTextView.value as! String
+        let expected = "<strong>text</strong>"
+        XCTAssertEqual(expected, text)
+    }
+
+    func testSimpleItalicText() {
+        richContentTextView = app.textViews[elementStringIDs.richTextField]
```
catehstn: this shows up in multiple places: I would add this to a helper eg enterTextInField().

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        richContentTextView.tap()
+        richContentTextView.enterAndSelectText(text: "text")
+
+        app.scrollViews.otherElements.buttons[elementStringIDs.italicButton].tap()
+        app.buttons[elementStringIDs.sourcecodeButton].tap()
+
+        htmlContentTextView = app.textViews[elementStringIDs.htmlTextField]
```
catehstn: I think you could also use another helper here which is getHtmlContent() returning a string.

Example/AztecUITests/AztecUITests.swift
catehstn: also good because then whatever happens to stop autocorrect only happens in one helper (with explanation).

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        let text: String = htmlContentTextView.value as! String
+        let expected = "<blockquote>text</blockquote>"
+        XCTAssertEqual(expected, text)
+    }
+
+    // Enable this test after unordered lists are fully implemented
+    func testSimpleUnorderedListText() {
```
catehstn: IMO for lists need multiple element tests also.

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        app.buttons[elementStringIDs.sourcecodeButton].tap()
+
+        htmlContentTextView = app.textViews[elementStringIDs.htmlTextField]
+        let text: String = htmlContentTextView.value as! String
+        let expected = "<ul><li>text</li></ul>"
+        XCTAssertEqual(expected, text)
+    }
+
+    // Enable this test after ordered lists are fully implemented
+    func testSimpleOrderedListText() {
```
catehstn: ditto.

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+
+    func testSimpleLinkedText() {
+        richContentTextView = app.textViews[elementStringIDs.richTextField]
+        richContentTextView.tap()
+        richContentTextView.enterAndSelectText(text: "text")
+
+        // Copy link to be auto-filled in URL field
+        UIPasteboard.general.string = "https://wordpress.com/"
+        app.scrollViews.otherElements.buttons[elementStringIDs.linkButton].tap()
+        app.alerts.buttons.element(boundBy: 3).tap()
```
catehstn: why is this boundBy: 3? does it not have an a11y label?

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        app.buttons[elementStringIDs.sourcecodeButton].tap()
+
+        htmlContentTextView = app.textViews[elementStringIDs.htmlTextField]
+        let text: String = htmlContentTextView.value as! String
+        let expected = "<a href=\"https://wordpress.com/\">text</a>"
+        XCTAssertEqual(expected, text)
+    }
+
+    func testHorizontalRuler() {
```
catehstn: IMO should have a test for this that has text either side also.

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        app.scrollViews.otherElements.buttons[elementStringIDs.horizontalrulerButton].tap()
+        app.buttons[elementStringIDs.sourcecodeButton].tap()
+
+        htmlContentTextView = app.textViews[elementStringIDs.htmlTextField]
+        let tex: String = htmlContentTextView.value as! String
+        let expected = "<hr>"
+        XCTAssertEqual(expected, text)
+    }
+
+    func testMoreTag() {
```
catehstn: same for this one.

Example/AztecUITests/AztecUITests.swift
```diff
@@ -0,0 +1,263 @@
...
+        app.scrollViews.otherElements.buttons[elementStringIDs.moreButton].tap()
+        app.buttons[elementStringIDs.sourcecodeButton].tap()
+
+        htmlContentTextView = app.textViews[elementStringIDs.htmlTextField]
+        let text: String = htmlContentTextView.value as! String
+        let expected = "<!--more-->"
+        XCTAssertEqual(expected, text)
+    }
+
+    func testHeadingOneText() {
```
catehstn: I think headings also need tests that show that the heading style ends after tapping enter.

catehstn: and that it isn't applied to text prior to the headings. and that two different heading types can be next to each other. probably enough there that it's worth making a separate file of tests for headings.

This would be fine as a separate PR.

astralbodies:
tenor

Example/AztecUITests/AztecUITests.swift
rachelmcr: Per our Slack conversation, we'll revisit/expand the tests for lists once lists are fully implemented and changes in visual mode affect the HTML (https://github.com/wordpress-mobile/WordPress-Aztec-iOS/issues/150). I'll leave the list tests as-is for now.

Example/AztecUITests/AztecUITests.swift
rachelmcr: Agreed. My plan for this PR was to implement the most basic tests for adding the correct style to text, but we should add the header tests you've outlined as well.

We should also have similar tests for blockquotes and lists (to test that the style ends when expected and is only applied to the expected text), as well as combining styles appropriately.

I've made a note to revisit these in a future PR.

Example/AztecUITests/XCTest+Extensions.swift
rachelmcr: Although we can use the `autocorrectionType` property to change the UITextView behavior itself, it doesn't look like we can change that property in the tests only (at least in a reasonably simple/straightforward way, as far as I can tell).

I'll go ahead with pasting the text for now and will revisit if I work out a way to disable autocorrect instead.

rachelmcr:
I updated the tests based on the feedback so far. A few notes:

- After researching different methods for dealing with autocorrect/autocomplete, I switched to typing in numbers instead of letters/words. Numbers aren't autocorrected (yay!) so we avoid autocorrect without having to switch to pasting in the text. (I kept to one number at a time because with multiple numbers they were sometimes entered in the wrong order when running the tests.)
- I also split up entering and selecting text in the helper functions. This should set us up well for testing a variety of scenarios where we focus on the rich text view, type some text, interact with the toolbar, and then continue typing (as seen in the added horizontal ruler and more tag tests).
- The next step will be adding more complex tests that include both unformatted and formatted text in the rich text view, to make sure that the formatting is enabled/disabled as expected while interacting with the toolbar buttons.

@catehstn please let me know if this addresses your earlier feedback and if you see any other issues with these changes. Thanks!

Example/AztecUITests/XCTest+Extensions.swift
catehstn: Yeah I had a look into this and basically we check for the launch flag, and in a debug-testing mode, we turn it off.
@diegoreymendez what do you think? Numbers are OK here, but for proper coverage I think we have to find a way to disable autocorrect.

diegoreymendez:
@rachelmcr - Considering both reviewers gave approval: is this ready for merging?

I'm particularly interested in getting this included before the closed beta next week.

hoverduck:
FWIW I also reviewed and give the 👍 from a general testing perspective.  I haven't been able to get the project to build to actually run the tests yet, but that's an unrelated problem.  /shrug

Issue #414, opened 2017-04-11
Formatters: remove need for zero-width spaces.


diegoreymendez:
<h3>Description:</h3>

Currently, our formatters require zero-width spaces so that they can apply styles (such as header, blockquote, list) to an empty range of text.

The problem with adding these characters is that we must remove them too, once they're no longer necessary.  This has not been the case so far.

We need to remove the need to have those characters inserted to begin with.

<h3>Details:</h3>

Some research has shown that modifying the typing attributes has an impact on how the text view lays out its contents.  This is the direction that should enable us to draw blockquotes, lists, etc without really adding any content.

The problem with this approach is that `NSLayoutManager` does not have access to the `typingAttributes` of its parent `UITextView`.

There are some hints online that these attributes are actually accessible through a private property.  Search for `typingAttributes` in [these NextStep sources](http://svn.gna.org/svn/gnustep/libs/gui/trunk/Headers/AppKit/NSLayoutManager.h).

<h3>Proposal:</h3>

- Make sure styles are applied to typingAttributes whenever the line is empty (if there's a newline character, the line is NOT empty).
- Consider drawing the bullet items for lists in `TextView` instead of `LayoutManager`.
- Make sure whenever a formatter enables or disables a style, the line where the caret is placed is redrawn (we probably need to redraw al involved lines when there's a selection instead of a caret).
- Remove zero-width space markers everywhere.

### Child issues:
- [x] Nuke ZeroWidth Space from HeaderFormatter - #419. 
- [x] Nuke ZeroWidth Space from PreFormatter - #420.
- [x] Nuke ZeroWidth Space from TextList - #421.
- [x] Nuke ZeroWidth Space from BlockquoteFormatter - #422.

SergioEstevao:
I tried to remove the need of this zero-width spaces two times already, and the issues was not the access to the attributes, but to be able to make the NSLayouManager to trigger any draw method.

What I found in my tests, is that until you write something the layoumanager is not triggered in any way to draw. If we find a method that triggers drawing then we can draw the markers.

diegoreymendez:
That's good info @SergioEstevao - I'll try to investigate in that direction.

diegoreymendez:
One alternative could be to draw the custom background elements in the `TextView` instead of in the layout manager, since the TextView already has access to the typing attributes, and doesn't need a glyph range to trigger its own draw methods.

SergioEstevao:
[...]

PR #425, opened 2017-04-17
Lists: Nuking Zero Width Spaces

### LINKED ISSUE #421: Remove zero-width spaces for lists.
We need to remove the need to use zero-width spaces in lists.

Child of #414

jleandroperez:
### Description:
In this PR we're nuking Zero Width Spaces from the TextLists Formatter.

**All of the Testing Scenarios** are backed up by a unit test. Ain't that cool?

Needs Review: @diegoreymendez @SergioEstevao 
**Diego:** Thanks for the infinite help on this one!!

Closes #421
Ref.: #414

### Details:
- Yes. We're removing the Zero Width Spaces from the TextLists's inner working. Whenever a bullet must be rendered, and there's no text, we'll be inserting a `\n` character.  

  Hook points for this behavior are **insertText** and the formatter's toggle themselves.
- And yes. We're overriding **typingAttributes**, because the `didSet` observer of both, **selectedRange** and **selectedTextRange** is not being always triggered (ie. user touch).
- **TextListFormatter** got a new helper to verify if lists of any kind are present in a collection of attributes.
- Removed a workaround from **LayoutManager**, which was being used to render the *extra line fragment* (Bullets when there was no text).

--

### Testing:
Detailing below all of the test cases.

#### Scenario I: List after Backspace
1.	Empty Document
2.	Toggle List
3. 	Type something
4. 	Backspace

Verify that the list remains onscreen.

#### Scenario II: Deleting after List
1.	Empty Document
2.	Toggle List
3. 	Arrow Down
4. 	Backspace

Verify that the List is gone.

#### Scenario III: Newline Not Inserted
1.	Empty Document
2.	Toggle List
3. 	Arrow Down
4.	Enter

Verify that the Newline gets inserted.

#### Scenario IV: New List Item when at the edge of the document
1. 	Empty Document
2. 	Toggle List
3. 	Type something
4.	Arrow Down
5.	Backspace
6.	Enter

Verify that a new bullet gets added.

#### Scenario V: Typing Attributes on new line
1. Empty Document
2. Toggle List
3. Arrow Down

Verify that the Typing Attributes do not contain a Text List: indentation should be normal. Typing any character should not display a bullet.

#### Scenario VI: Empty Lines
1. Empty Document
2. Toggle List
3. Insert a newline

Verify that the bullet in the first line gets removed.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -497,7 +497,7 @@ open class TextView: UITextView {
     // MARK: - Formatting
 
     func toggle(formatter: AttributeFormatter, atRange range: NSRange) {
-        let applicationRange = storage.toggle(formatter: formatter, at: range)        
+        let applicationRange = storage.toggle(formatter: formatter, at: range)
         if applicationRange.length == 0 {
             typingAttributes = formatter.toggle(in: typingAttributes)
         } else {
@@ -568,20 +568,43 @@ open class TextView: UITextView {
     /// - Parameter range: The NSRange to edit.
     ///
     open func toggleOrderedList(range: NSRange) {
+        insertText("\n")
+        selectedRange = NSRange(location: 0, length: 0)
+
         let formatter = TextListFormatter(style: .ordered, placeholderAttributes: typingAttributes)
         toggle(formatter: formatter, atRange: range)
-        forceRedrawCursorAfterDelay()
+
+//        forceRedrawCursorAfterDelay()
     }
 
+    override open var typingAttributes: [String : Any] {
+        get {
+            NSLog("### typingAttributes Getter")
+            if selectedRange.location == storage.length && selectedRange.length == 0 {
+                var updated = super.typingAttributes
+                updated.removeValue(forKey: NSParagraphStyleAttributeName)
+
+                return updated
+            }
+
+            return super.typingAttributes
+        }
+        set {
+            super.typingAttributes = newValue
+        }
+    }
 
     /// Adds or removes a unordered list style from the specified range.
     ///
     /// - Parameter range: The NSRange to edit.
     ///
     open func toggleUnorderedList(range: NSRange) {
+        insertText("\n")
+        selectedRange = NSRange(location: 0, length: 0)
+
         let formatter = TextListFormatter(style: .unordered, placeholderAttributes: typingAttributes)
         toggle(formatter: formatter, atRange: range)
-        forceRedrawCursorAfterDelay()
+//        forceRedrawCursorAfterDelay()
     }
 
 
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -168,7 +168,7 @@ open class TextView: UITextView {
         storage.undoManager = undoManager
         commonInit()
     }
-    
+
     required public init?(coder aDecoder: NSCoder) {
 
         defaultFont = UIFont.systemFont(ofSize: 14)
@@ -573,8 +573,7 @@ open class TextView: UITextView {
 
         let formatter = TextListFormatter(style: .ordered, placeholderAttributes: typingAttributes)
         toggle(formatter: formatter, atRange: range)
-
-//        forceRedrawCursorAfterDelay()
+        forceRedrawCursorAfterDelay()
     }
 
     override open var typingAttributes: [String : Any] {
@@ -604,7 +603,7 @@ open class TextView: UITextView {
 
         let formatter = TextListFormatter(style: .unordered, placeholderAttributes: typingAttributes)
         toggle(formatter: formatter, atRange: range)
-//        forceRedrawCursorAfterDelay()
+        forceRedrawCursorAfterDelay()
     }
 
 
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -787,11 +787,7 @@ open class TextView: UITextView {
         }
 
         let previousRange = NSRange(location: selectedRange.location - 1, length: 1)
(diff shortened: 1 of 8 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -724,6 +767,56 @@ open class TextView: UITextView {
...
+            return attributes
+        }
+
+        let previousRange = NSRange(location: selectedRange.location - 1, length: 1)
+        let previousString = storage.safeSubstring(at: previousRange) ?? String(.newline)
+        guard previousString == String(.newline) else {
+            return attributes
+        }
+
+        var updatedAttributes = attributes
```
diegoreymendez: From this line below, the problem is that we're not only removing the `textList` style, but we're actually resetting the full paragraph configuration to the default, which is not really desired.  For instance: we may not want to remove the blockquote style.

To avoid resetting the whole paragraph, you can use this code instead (or an equivalent code):

```swift
let orderedListFormatter = TextListFormatter(style: .ordered)
let unorderedListFormatter = TextListFormatter(style: .unordered)

if orderedListFormatter.present(in: attributes) {
    return orderedListFormatter.remove(from: attributes)
} else if unorderedListFormatter.present(in: attributes) {
    return unorderedListFormatter.remove(from: attributes)
} else {
    return attributes
}
```

PS: don't worry too much about the code being perfect.  This code will change ENORMOUSLY once we change to using a mechanism closer to what Apple uses for nested lists, but this should be a good starting point.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -724,6 +767,56 @@ open class TextView: UITextView {
...
+    ///     A. The caret is at the very end of the document (+ there is no selected text)
+    ///     B. There's a list!
+    ///     C. The previous character is a '\n'
+    ///
+    /// - Parameter attributes: Typing Attributes.
+    ///
+    /// - Returns: Updated Typing Attributes.
+    ///
+    private func ensureRemovalOfListAttribute(from attributes: [String: Any]) -> [String: Any] {
+        guard selectedRange.location == storage.length && selectedRange.length == 0 else {
```
diegoreymendez: For reading simplicity I'd consider removing `selectedRange.length == 0`.  If the caret position is indeed at EOF, then the length will always be zero.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -724,6 +767,56 @@ open class TextView: UITextView {
     }
 
 
+    /// Removes the List Attributes from a collection of attributes, whenever:
+    ///
+    ///     A. The caret is at the very end of the document (+ there is no selected text)
```
diegoreymendez: If the selected range location starts at EOF, then `(+ there is no selected text)` is always true.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -724,6 +767,56 @@ open class TextView: UITextView {
     }
 
 
+    /// Removes the List Attributes from a collection of attributes, whenever:
```
diegoreymendez: I'd consider adding extra documentation to explain why this is necessary.  Specifically something like:

```swift
// This is necessary because when the caret is at EOF, and the previous `\n` character has
// a textList style, that style will remain in the `typingAttributes`.  We'll only allow the style
// to remain if there are contents in the current line with the textList style (in which case
// this condition won't ever trigger because we'll either no longer be at EOF, or the previous
// character won't be `\n`).

File: Aztec/Classes/TextKit/TextView.swift [hunk 5]
```diff
@@ -724,6 +767,56 @@ open class TextView: UITextView {
...
+
+
+    /// Inserts an empty line, whenever we're at the end of the document, and there's no selected text.
+    ///
+    private func ensureInsertionOfNewlineOnEmptyDocuments() {
+        guard selectedRange.location == storage.length && selectedRange.length == 0 else {
+            return
+        }
+
+        // Note: We *really* need to use super, so that we prevent recursive loops.
```
diegoreymendez: I removed all references to `super` in this method, and the code seems to work fine.  Is it possible this was a problem when this code was called directly from the `typingAttributes` getter instead?

Aztec/Classes/TextKit/TextView.swift
jleandroperez: Good call, thank you!

Aztec/Classes/TextKit/TextView.swift
jleandroperez: Updated, thank you!!

Aztec/Classes/TextKit/TextView.swift
jleandroperez: Absolutely right, updating as well, thanks!!

Aztec/Classes/TextKit/TextView.swift
jleandroperez: 100% good call, thanks for the snippet!!

Aztec/Classes/TextKit/TextView.swift
jleandroperez: For the record:

Removed the `super.typingAttributes` (it was probably crashing due to testing code!), but left the `super.insertText`, since, without that one, we do get a loop.

Re: extra docs, as discussed over Slack, we've got a unit test that would trigger the loop, so we're cool.

Thanks a lot Diego!

jleandroperez:
Woooooo thanks SO much for the support Diego!! this one is more yours than mine!!!

SergioEstevao:
@jleandroperez late to the party, but this PR was excellent!!!

jleandroperez:
Thanks a lot @SergioEstevao!!! 🔥 🔥 🔥 🔥 🔥 🔥

PR #426, opened 2017-04-17
Improving newlines handling

diegoreymendez:
### Description:

This PR completes the removal of our visual-only attribute.  Visual-only newlines are now replaced by the paragraph-separator unicode character.

This PR also adds a lot of range-mapping methods and properly differentiate between UTF16 and Swift String `NSRange` objects.

### How to test:

**Test 1:**

Run the unit tests.  A lot of new tests have been added.
Run the UI tests.

**Test 2:**

Load the sample content and make sure it looks right.  Edit text, switch to HTML and back and make sure the text changes still apply.

**Test 3:**

Do some editing in the empty editor demo.


diegoreymendez:
We have several methods using `String(.newline)` to identify the end-of-line.  This is bound to cause issues with this PR since we're sometimes using `String(.paragraphSeparator)` instead.

One such example:

```
    /// Check if the location passed is the beggining of a new line.
    ///
    /// - Parameter location: the position to check
    /// - Returns: true if beggining of a new line false otherwise
    ///
    func isStartOfNewLine(atLocation location: Int) -> Bool {
        var isStartOfLine = length == 0 || location == 0
        if length > 0 && location > 0 {
            let previousRange = NSRange(location: location - 1, length: 1)
            let previousString = attributedSubstring(from: previousRange).string
            isStartOfLine = previousString == String(.newline)
        }
        return isStartOfLine
    }
```

Before moving ahead, I'll make sure to evaluate the current code and come up with a more flexible solution for performing these checks.

diegoreymendez:
@jleandroperez - Ready for another round.

diegoreymendez:
I'm going ahead and merging, since I need this to unlock a separate PR I'm working on, and we're nearing our deadline for the closed beta.

Any other issues found please let me know and I'll take care.

PR #429, opened 2017-04-19
PreFormatter: Nukes Zero Width Spacing character

### LINKED ISSUE #414: Formatters: remove need for zero-width spaces.

### LINKED ISSUE #420: Remove zero-width spaces for PRE style.
We need to remove the need to use zero-width spaces in the PRE style.

Child of #414

jleandroperez:
### Details:
This is the (last) PR in the series "Let's stop using the Zero Width Spacing" character. We're replicating the exact same mechanisms (+ battle of unit tests) we've already got for Lists and Blockquotes.

Needs Review: @diegoreymendez 
Thaaanks!!!

Closes #420
Closes #414

### Testing:
Verifying this PR is gonna be a bit tricky, since we don't really have (yet) a way to toggle `Pre` blocks. Parsing from HTML is supported, but visual edition hasn't been shipped yet.

In this PR i've added a `togglePre` method, analog to `toggleList` and `toggleBlockquote`. In order to verify this PR, please:

- Make sure the unit tests pass
- Whenever the test scenario indicates "Toggle Pre", please:
 A. Switch to HTML
 B. Insert `<pre>Payload</pre>`
 C. Switch back!

Apologies about this inconvenience. We don't really have a `Pre` gridicon we could wire, right now, for this PR. Formal support coming in later, in another PR!

--

### Scenario A: Missing Pre after Backspace
1. Launch the empty editor
2. Toggle the Pre format
3. Enter any random text
4. Hit backspace

Verify that the pre style is still present.

### Scenario B: Nuking Pre after Backspace
1. Launch the empty editor
2. Insert `Pre` Tag
3. Enter any random text
4. Select the end of the document
5. Hit backspace

Verify that the Pre gets effectively removed.

### Scenario C: Newline after Pre
1. Launch the empty editor
2. Insert `Pre` Tag

Verify that a newline is inserted below the pre.

### Scenario D: Nuking bottom `\n` and adding newlines
1. Launch the empty editor
2. Toggle the Pre format
3. Enter any random text
4. Select the end of the document
5. Hit backspace
6. Hit `\n` and enter any random text

Verify that the new Pre line gets it's proper style, even when the bottom `\n` was initially nuked.

### Scenario E: Typing Attributes below Pre
1. Launch the empty editor
2. Insert `Pre` Tag
3. Press the arrow down

Verify that the Pre gets removed from the typing attributes (and that the text indentation looks normal).

### Scenario F: Autoremoval after hitting return on newline
1. Launch the empty editor
2. Insert `Pre` Tag
3. Hit return

Verify that the blockquote gets removed.

### DIFF
--- a/Aztec/Classes/Formatters/AttributeFormatter.swift
+++ b/Aztec/Classes/Formatters/AttributeFormatter.swift
@@ -64,8 +64,6 @@ protocol AttributeFormatter {
     func applicationRange(for range: NSRange, in text: NSAttributedString) -> NSRange
 
     func worksInEmptyRange() -> Bool
-
-    func needsEmptyLinePlaceholder() -> Bool
 }
 
 
@@ -120,15 +118,9 @@ extension AttributeFormatter {
     ///
     @discardableResult
     func applyAttributes(to text: NSMutableAttributedString, at range: NSRange) -> NSRange {
-        var rangeToApply = applicationRange(for: range, in: text)
+        let rangeToApply = applicationRange(for: range, in: text)
 
-        if needsEmptyLinePlaceholder() && worksInEmptyRange() && ( rangeToApply.length == 0 || text.length == 0)   {
-            let placeholder = placeholderForEmptyLine(using: placeholderAttributes)
-            text.insert(placeholder, at: rangeToApply.location)
-            rangeToApply = NSMakeRange(rangeToApply.location, placeholder.length)
-        }
-
-        text.enumerateAttributes(in: rangeToApply, options: []) { (attributes, range, stop) in
+        text.enumerateAttributes(in: rangeToApply, options: []) { (attributes, range, _) in
             let currentAttributes = text.attributes(at: range.location, effectiveRange: nil)
             let attributes = apply(to: currentAttributes)
             text.addAttributes(attributes, range: range)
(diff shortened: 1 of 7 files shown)


jleandroperez:
WOOOOOOOO **Thank you** Diego!!!

Issue #165, opened 2016-12-06
Attachments: Video Support


jleandroperez:
Let's add Video Support for Attachments!

SergioEstevao:
I will start implementing this assuming the use of the ```<video>``` tag at the moment.
In a second phase we will need to look at implementing through  shortcodes: ```[wpvideo]``` for video press and ```[video]``` for regular wp.org.

jleandroperez:
Updating the milestone to Beta 1!

PR #431, opened 2017-04-20
Refactor TextAttachment to be MediaAttachment.

SergioEstevao:
Refs #165 

This PR refactor TextAttachment to be MediaAttachment. It then creates a ImageAttachment to replace the uses of the previous TextAttachment and it also implements VideoAttachment as a subclass of MediaAttachment.

This change allow sharing of a lot of layout and delegate code between the Image and Video implementations and make simples in the future do changes to both systems.

How to test:
 * Run the Demo app
 * Check that the image and video of the demo show up correctly
 * Test the interactions with the image ( add image, remove image, tap for details)
 * Test the interactions with videos ( tap to play, remove)


diegoreymendez:
A few notes on issue I've spotted with media:

1. Backspacing on the video is not working (switch to HTML and back and the video is still there).
2. Editing the image URL does not refresh it.

Considering that you're the person that has worked the most on it, @SergioEstevao, can you please consider if these need to be addressed in this PR, or create new issues if necessary?

SergioEstevao:
@diegoreymendez those issues where already there
1. We are still not removing video from the HTML, that will be done in a follow up PR
2. This look a bug that was already there, you can file a new issue for it.

diegoreymendez:
I see no open issues for either point 1 or 2.  Can you create those, please?

I'd prefer the owner of each feature to be responsible for making sure the right issues exist for it to keep track of progress, otherwise I risk becoming a bottleneck and / or missing things.

SergioEstevao:
* Issue 1 tracked here: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/433
 * Issue 2 tracked here: #434

PR #458, opened 2017-04-28
Fixes two bugs.

### LINKED ISSUE #457: Backspaces can create line breaks
A backspace in the latest beta can create a line break sometimes.

Here's a video showing [the issue](https://cloudup.com/cM7f2kEiO88).

/cc @rachelmcr.

diegoreymendez:
Fixes #457 and another bug with ranges that went unnoticed so far, but could've caused crashes.

To test:

Just run the unit tests, since the case that was breaking is being caught by the provided test.


diegoreymendez:
The PR seems to have broken since I had to change the source branch.

Issue #450, opened 2017-04-25
Lists can't have empty lines.


diegoreymendez:
Lists can't currently have empty lines.  This is caused by a more fundamental problem described [here](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/447).

Don't start this issue until #447 is completed, since there's a strong dependency on it.

diegoreymendez:
This was fixed in beta 4.

PR #463, opened 2017-05-04
Fixes some issues with lists.

### LINKED ISSUE #456: Lists losing formatting
As reported by @rachelmcr:

> Here’s a case where a user switches list styles after creating the list:
> 
> 1. Create an ordered list with the steps above.
> 2. Highlight the list.
> 3. Tap the unordered list button to switch to a bulleted list.
> 
> Everything looks right at first, but switching to HTML view shows that everything was condensed into a single list item and the line breaks are gone.
> 
> Here’s another case that doesn’t go so well, where the user types out the list and then adds the list styles:
> 
> 1. Type out the list as regular text with line breaks.
> 2. Highlight the list.
> 3. Tap a list button to add the list styles.
>
> Everything looks right in the visual view, but switching to HTML shows that there’s a single list item with line breaks between the other items. This becomes apparent when you preview the post.

PS: make sure to write automated tests to prevent regressions in this feature.

diegoreymendez:
### Description:

This PR fixes #456.

Lists were losing some of their formatting when combined with multi-line selection and newlines in general.

This PR solves those issues.

### Testing:

**Test 1:**

1. Launch the empty editor.
2. Add an ordered list with 3 lines of text or more.
3. Select part of the list.
4. Convert to unordered list.
5. Switch to HTML and back.  Make sure everything is OK.

**Additional testing:**

Play with lists and try to break them in visual mode.

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -50,6 +50,16 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     ///
     fileprivate func convert(_ node: Node, inheritingAttributes attributes: [String:Any]) -> NSAttributedString {
 
+        let string = convertContents(of: node, inheritingAttributes: attributes)
+
+        guard node.needsClosingParagraphSeparator() else {
+            return string
+        }
+
+        return appendParagraphSeparator(to: string, inheritingAttributes: attributes)
+    }
+
+    private func convertContents(of node: Node, inheritingAttributes attributes: [String:Any]) -> NSAttributedString {
         switch node {
         case let textNode as TextNode:
             return convertTextNode(textNode, inheritingAttributes: attributes)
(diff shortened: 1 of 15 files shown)


SergioEstevao:
Found some issues...

1. If you open the demo controller with content and go to the middle paragraph of the  unordered list and switch it to ordered  list, the app crashes. ( but if you remove the list, add again, and then switch it works)
2.On any list if you go the end of any of the lines of the list press Enter and the switch to HTML and back you will see that the last character from the line move to the new list line.

diegoreymendez:
@SergioEstevao - This is ready for round 2!

SergioEstevao:
@diegoreymendez looking better, but still seeing one issue, if you add new empty list line in the middle of a list and then switch to html and back the new line is removed.

diegoreymendez:
@SergioEstevao - That one will be tracked [separately](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/450).

The good news is it wasn't introduced by this PR.  The bad news is we'll need to refactor newlines again to make that work.  Unfortunately it would be too big a change to include in this PR.

diegoreymendez:
Thank you, @SergioEstevao !

PR #469, opened 2017-05-08
TextView: Updates Style Removal Logic

### LINKED ISSUE #466: Loosing Header Style after Return
### Steps:
1. Launch the Sample Demo (w/text)
2. With the caret at the initial position (0, 0), hit return

As a result, the h1 style will be lost.

jleandroperez:
Fixes #466

### To test:
1. Launch the Sample Demo (w/text)
2. With the caret at the initial position (0, 0), hit return
3. Verify that the h1 style is not lost anymore

Needs Review: @diegoreymendez 
Thanks in advance!


jleandroperez:
Thanks for the review @diegoreymendez!!

PR #470, opened 2017-05-08
Support for Nested Lists

### LINKED ISSUE #448: Add Nested Lists Support
We should support nested lists in Aztec.

SergioEstevao:
Fixes #448

This PR adds support for nested lists only.
It's still not 100% but I wanted @diegoreymendez and @jleandroperez  have a look on the code until now, and play around to see if they see any major issues.

I will continue to work in fixing the failing unit tests.

To test:
 - Run the sample app and check if the content shows correctly
 - Interact around the nested lists and see if all is working correctly
 - Interact on regular lists and see if all things are still working.

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -275,8 +275,8 @@ class HMTLNodeToNSAttributedString: SafeConverter {
     }
 
     public let elementToFormattersMap: [StandardElementType: AttributeFormatter] = [
-        .ol: TextListFormatter(style: .ordered),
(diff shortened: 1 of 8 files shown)


SergioEstevao:
So I fixed the unit tests by adding an option on the formatters to say if we are increasing the depth of the list or just replacing the current level.

I tested it a bit and it looks it's working pretty well on the visual side. On the DOM it looks that when you change the list type on a nested list if gets a bit confused and moves the nested list out of the ```<li>``` element. @diegoreymendez do you mind to give it a look?

SergioEstevao:
Discussed with @diegoreymendez online and decided that we can tackle the DOM issues on a later PR after a newline refactor that is on the way.

PR #471, opened 2017-05-10
Newlines refactoring.

### LINKED ISSUE #447: DOM Newline Editing Improvements:
### Description:

Right now we are having some problems with the DOM editing logic, due to limitations in our data structure and logic.

The Visual-to-DOM range mapping we're using is not good enough since it potentially reduces several Visual locations to a single DOM location.

As an example: lists can suffer from several different caret positions being mapped to a single position in DOM coordinates.

The locations marked with a pipe in the example below, all (mistakenly) mapped to the same DOM location.

```
- Item 1|
- |
- |
- |Item 2
```

### Proposal:

This description will be improved moving forward, but the basis of the proposal is to remove Visual -> DOM range mapping, and let each node (or some object in charge of it) handle editions to its own visual newlines.

We're going to need to have both a pure representation of a node's contents and a visual representation of it's contents:

For example:

```html
<blockquote>Hello!</blockquote>
```

Would have its pure HTML text content representation of `Hello!`, and it's visual representation as `Hello!\n`.

diegoreymendez:
Fixes #447 

TBD.  Not ready for review.


diegoreymendez:
Closing for now as there was a shift in direction.

PR #472, opened 2017-05-10
HTML: Pretty Print

### LINKED ISSUE #239: When switching to HTML mode, add an HTML beautifier step.
When switching to HTML mode, add an HTML beautifier step to make the HTML content easier to follow and edit.

jleandroperez:
### Details:
- Nuked legacy **OutHTMLConverter**, and implemented a new version, not based anymore on libxml.
- Support for HTML prettification in place
- New Untit Tests!

Closes #239

### Prettifier rules:
- Blocklevel element's Opening Tag get a preceeding `\n` character
- Blocklevel element's Closing Tag get a preceeding `\n` if they contain at least one Blocklevel Child.
- Blocklevel element's Closing Tag get a trailing `\n` if their right sibling is **not** a Blocklevel Element.

### Testing:
- Verify the Unit Tests run correctly
- Verify the Demo Document gets properly prettified, when in HTML Mode
- Play with an empty document, try switching from Rich / HTML. Verify everything looks nice!

cc @diegoreymendez


Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
```diff
@@ -1,55 +1,215 @@
...
+        return "<!--" + node.comment + "-->"
+    }
 
-            xmlDocSetRootElement(xmlDocPtr, xmlNodePtr)
-            htmlNodeDumpFormatOutput(outputBuffer, xmlDocPtr, xmlNodePtr, "UTF-8", 0)
+    /// Serializes an ElementNode into it's HTML String Representation
+    ///
+    private func convert(element node: ElementNode, level: Int) -> String {
+        // Prefixes + Posfixes
+        let indentForOpeningTag = requiresOpeningTagPrefix(node) ? indentationString(for: level) : ""
```
diegoreymendez: We're calling `requiresClosingTagPrefix()` and `requiresOpeningTagPrefix(node)` repeatedly on the same node.

Any chance we can store these calls in a single variable at the beginning of this method instead?

Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
```diff
@@ -1,55 +1,215 @@
...
+        let indentForClosingTag = requiresClosingTagPrefix(node) ? indentationString(for: level) : ""
+        let prefixForOpeningTag = requiresOpeningTagPrefix(node) ? String(.newline) : ""
+        let prefixForClosingTag = requiresClosingTagPrefix(node) ? String(.newline) : ""
+        let posfixForClosingTag = requiresClosingTagPosfix(node) ? String(.newline) : ""
 
-            let buffer = xmlBufContent(outputBuffer.pointee.buffer)!
+        // Serialize Attributes
```
diegoreymendez: This method handles the detail of:

1. Defining the indentation string, prefix strings, postfix strings.
2. Converting the attributes.
3. Creating the opening tag
4. Getting the children representations
5. Creating the closing tag

Using comments to name chunks of code is not the best tool for the job.  A method would achieve the same result, and do it better since it would also make the chunk usable from other portions of the code.

Can we split this method up more atomically?

Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
```diff
@@ -1,55 +1,215 @@
...
+        typealias Node              = Libxml2.Node
+        typealias TextNode          = Libxml2.TextNode
+        typealias CommentNode       = Libxml2.CommentNode
+        typealias RootNode          = Libxml2.RootNode
+
+        /// Indentation String to be applied
         ///
-        /// - Parameters:
-        ///     - rawNode: the Libxml2 Node to convert.
+        var indentationString = "  "
```
diegoreymendez: The converters were designed to be transient, and created whenever a specific conversion is needed.  Having these settings be variables makes it feel like you may want to store a single converter for separate conversions, which is not ideal (because for example this class is not thread safe).

Can we change these to constants and configure them through the required initializer instead?

PS: changing this after a conversion has started is risky as well.

Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
```diff
@@ -1,55 +1,215 @@
...
+
+        /// Indicates whether we want Pretty Print or not
         ///
-        /// - Returns: a String object representing the specified HTML data.
+        var prettyPrint = false
```
diegoreymendez: The converters were designed to be transient, and created whenever a specific conversion is needed.  Having these settings be variables makes it feel like you may want to store a single converter for separate conversions, which is not ideal (because for example this class is not thread safe).

Can we change these to constants and configure them through the required initializer instead?

PS: changing this after a conversion has started is risky as well.

Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
jleandroperez: Good call, updating. Thanks for the suggestion!!

Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
jleandroperez: Good call, updating. Thanks for the suggestion!!

Aztec/Classes/Libxml2/Converters/Out/OutHTMLConverter.swift
jleandroperez: Huh sorry about missing that! updating as well!

jleandroperez:
@diegoreymendez all comments addressed, ready for another look!

jleandroperez:
Thanks Diego!!

PR #475, opened 2017-05-15
Add basic usage documentation

rmccue:
Since Aztec is written in Swift, it's a bit different using it in Obj-C land than a traditional Obj-C view. In particular, you have to load in the compatibility headers, which are actually quite a pain to find in the official documentation.

To improve this experience a bit, it'd be nice to add basic usage documentation to the readme. That's what this PR does. :)

Potentially, adding an example Obj-C project (or even just one view) might be useful as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,6 @@ pod "WordPress-Aztec-iOS"
 
 ## Usage
 
-### Swift
-
 After installing Aztec, import the module and use the `Aztec.TextView` view:
 
 ```swift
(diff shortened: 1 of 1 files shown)


rmccue:
(As a side note: looks like the delegates aren't available in Obj-C?)

diegoreymendez:
Thank you for the PR, @rmccue !

Objective-C, it's really not supported yet.

During the initial planning for Aztec we made the decision of building Aztec as a Swift-only library.  This decision is mainly a strategical one: the priority is on having the library ready and feature-complete.

We're not ruling out Objective-C support, but at the time of this reply, there's literally been zero time spent on making that happen, and implying there's some degree of support in the documentation would be a bad idea.

We don't rule it out for the future, though.

diegoreymendez:
PS: we could still use the "Swift" section if you could adjust the PR.

rmccue:
Thanks @diegoreymendez, as I dug deeper I suspected this may be the case. We don't have any particular preference with our project, so happy to use Swift, but wanted to give Obj-C a go first (I know it a little better).

Will update the PR, and also note that Obj-C is not officially supported.

PR #477, opened 2017-05-15
Unknown HTML Edition

### LINKED ISSUE #411: Unknown HTML: Editing Support
<h3>Description:</h3>

We need to implement a mechanism to allow users to edit unknown HTML snippets right from visual mode.

We are going to base this feature on the same feature from Aztec Android, with some slight UI/UX differences.

<h3>Details:</h3>

[Here's a video](https://cloudup.com/c341L7hjWJZ) showing the feature in Aztec Android.

We should do something similar, although to simplify and standardize the experience, the popup will be a full-screen VC for iPhone, and will be a popup for the iPad.

On the iPhone the experience should be similar to that of entering HTML mode, but it will only show the Unknown HTML snippet, and will not include the editor toolbar.

On the iPad the experience should be the same, with the expection that the VC will not be full-screen (but more like a popup on top of the editor).  The editor popup should let you scroll up and down if there's more content to show.

jleandroperez:
### Details:
In this PR we're adding support for Unknown HTML Edition.

cc @diegoreymendez 
Closes #411

### To test:
- Verify that the Unit Tests run correctly
- Launch the Demo and tap over the `[TABLE]` at the bottom
- Verify that the Unknown HTML Editor works smoothly

#### Known Issues:
- The new HTML should be valid. We'll address this later.


jleandroperez:
Thanks for the review sir!!

PR #482, opened 2017-05-24
Fix media unavailable issues

### LINKED ISSUE #434: When changing an Image url to an invalid URL the image doesn't update.
### Expected behavior
 * Tap on image
 * Choose edit details
 * Change the image URL to an invalid image URL
 * Save the image details
 * See the image change to the placeholder on the visual editor

### Actual behavior
 * Tap on image
 * Choose edit details
 * Change the image URL to an invalid image URL
 * Save the image details
 *The image stays the same, no error is show

 * Tap on image
 * Choose edit details
 * Change the image URL to an invalid image URL
 * Save the image details

SergioEstevao:
Fixes #434 

This PR address two issues:
 1. The #434 issue when changing a image src to a non available/incorrect source the image is not updated.
 1. When the image or video don't have a src or poster attribute

Both issues are sorted by using a placeholder image to represent them when no media image is available
 
To test:

Scenario 1:
 - Start the demo app
 - Open the demo with content
 - Scroll to the image on the demo
 - Tap on it
 - Change the source of the image to an invalid image
 - Make sure the placeholder image shows in place
 - Repeat the process but with a valid image
 - Check that the image shows up.

Scenario 2:
 - Start the demo app
 - Open the demo with content
 - Switch to HTML mode
 - Remove src from the image tag and poster from the video shortcode
 - Switch to visual mode
 - Check if the image and video show as a placeholder.

### DIFF
--- a/Aztec/Classes/TextKit/TextStorage.swift
+++ b/Aztec/Classes/TextKit/TextStorage.swift
@@ -1061,6 +1061,12 @@ open class TextStorage: NSTextStorage {
 //
 extension TextStorage: MediaAttachmentDelegate {
 
+    func mediaAttachmentPlaceholderImageFor(attachment: MediaAttachment) -> UIImage {
+        assert(attachmentsDelegate != nil)
+        return attachmentsDelegate.storage(self, missingImageFor: attachment)
+    }
+
+
     func mediaAttachment(
         _ mediaAttachment: MediaAttachment,
         imageForURL url: URL,
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -832,7 +832,7 @@ extension EditorDemoController : Aztec.FormatBarDelegate {
 
 extension EditorDemoController: TextViewAttachmentDelegate {
 
-    func textView(_ textView: TextView, imageAt url: URL, onSuccess success: @escaping (UIImage) -> Void, onFailure failure: @escaping (Void) -> Void) -> UIImage {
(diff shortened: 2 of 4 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -849,9 +849,28 @@ extension EditorDemoController: TextViewAttachmentDelegate {
...
+
+    func textView(_ textView: TextView, placeholderForAttachment attachment: NSTextAttachment) -> UIImage {
+        return placeholderImage(forAttachment: attachment)
+    }
+
+    func placeholderImage(forAttachment attachment: NSTextAttachment) -> UIImage {
+        let placeholderImage: UIImage
+        switch attachment {
+        case _ as ImageAttachment:
+            placeholderImage = Gridicon.iconOfType(.image, withSize: CGSize(width:32, height:32))
```
diegoreymendez: Since the 32x32 size seems to be standard, why not create a constant for the size.  Something like `defaultAttachmentSize` would do.

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -1061,6 +1061,12 @@ open class TextStorage: NSTextStorage {
 //
 extension TextStorage: MediaAttachmentDelegate {
 
+    func mediaAttachmentPlaceholderImageFor(attachment: MediaAttachment) -> UIImage {
```
diegoreymendez: Thoughts on calling this `placeholderImage(for attachment: MediaAttachment)`?

SergioEstevao:

Issue #492, opened 2017-06-07
[Master Issues] Remove DOM tree.


diegoreymendez:
Migrate to only using NSAttributedString.

More information will be added here after defining the work plan.

### Tasks:

- [x] Unhook the dom syncrhonization
- [x] Adapt ParagraphStyle to hold all paragraph styles in a single array
- [x] Adapt ParagraphStyle so that each style has HTML attribute info
- [x] Write NSAttributedString -> Node output converter.
  - [x] Unit Tests
  - [x] Nuke internal enums
  - [x] Newline Support
- [x] Remove EditContext from the code.
- [x] Remove old logic and unused unit tests.

diegoreymendez:
Closing this for now.  We'll be treating any remaining work in more atomic issues.

PR #496, opened 2017-06-08
Introducing HTMLRepresentation

diegoreymendez:
Fixes part of #492.

Adds an HTMLRepresentation protocol, and two classes implementing it so we can attach HTML representation information to `NSAttributedString`.

**How to test:**

Simlpy run the unit tests in the following two files:

- HTMLAttributeRepresentationTests.
- HTMLElementRepresentationTests.

### DIFF
--- a/Aztec/Classes/NSAttributedString/HTML/HTMLAttributeRepresentation.swift
+++ b/Aztec/Classes/NSAttributedString/HTML/HTMLAttributeRepresentation.swift
@@ -1,0 +1,30 @@
+
+class HTMLAttributeRepresentation: HTMLRepresentation {
+
+    typealias Attribute = Libxml2.Attribute
+    typealias StringAttribute = Libxml2.StringAttribute
+
+    /// The element that owns this attribute.
+    ///
+    var element: HTMLElementRepresentation?
+
+    /// The attribute name.
+    ///
+    let name: String
+
+    /// The attribute's value, if present.
+    ///
+    let value: String?
+
+    init(for attribute: Attribute, in element: HTMLElementRepresentation? = nil) {
+
+        self.element = element
+        name = attribute.name
+
+        if let stringAttribute = attribute as? StringAttribute {
+            value = stringAttribute.value
+        } else {
+            value = nil
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


Aztec/Classes/NSAttributedString/HTML/HTMLAttributeRepresentation.swift
```diff
@@ -0,0 +1,30 @@
...
+    init(for attribute: Attribute, in element: HTMLElementRepresentation? = nil) {
+
+        self.element = element
+        name = attribute.name
+
+        if let stringAttribute = attribute as? StringAttribute {
+            value = stringAttribute.value
+        } else {
+            value = nil
+        }
```
jleandroperez: Hoooow about...

```
let stringAttribute = attribute as? StringAttribute {
value = stringAttribute?.value
```

(Or)

```
guard let stringAttribute = attribute as? StringAttribute else {
   return
}

value = stringAttribute.value
```

Aztec/Classes/NSAttributedString/HTML/HTMLRepresentation.swift
```diff
@@ -0,0 +1,4 @@
+import Foundation
+
+protocol HTMLRepresentation {
+}
```
jleandroperez: (Still wondering if we want this guy around). If we do, maybe just move it to `HTMLElementRepresentator.swift` as well?

Aztec/Classes/NSAttributedString/HTML/HTMLAttributeRepresentation.swift
diegoreymendez: I can't really use the second mechanism as the initializer needs to assign a value to `value` in all cases (the code can't just `return`).  And even if I set a default value there, adding a guard to an `init` method means you can't initialize anything else below it, if the need arises - it's a bit restrictive.

The first snippet is nicer so I'm going to switch to it.

Aztec/Classes/NSAttributedString/HTML/HTMLRepresentation.swift
diegoreymendez: Regarding if it's necessary - it will be.  Otherwise you can't have properties or arrays that can contain both `HTMLElementRepresentation` and `HTMLAttributeRepresentation` objects.

Since a style can be represented by either an element or an attribute, unless there's a better alternative we can implement right now, I'd still go forward with this.

Regarding moving the protocol, I'm not sure how that makes sense since the protocol is not related to `HTMLElementRepresentator` in any way, other than the latter implementing the protocol.

diegoreymendez:
Thank you, @jleandroperez !

PR #499, opened 2017-06-08
Out Converter: Nuking Internal Structures

jleandroperez:
### Details:
In this PR we're nuking NSAttributedStringToNodes's internal structures. In order to test this, please, run a smoke test using the empty editor, and verify nothing (too obvious) breaks.

More changes upcoming in another PR!

Ref. #492

cc @diegoreymendez 
Thanks in advance!!


jleandroperez:
As discussed with @diegoreymendez, merging right away so we can submit a second PR on this.

PR #500, opened 2017-06-08
Make sure the output is done from each paragraph on the attributed string

SergioEstevao:
This PR improves/fixes the output by actually checking the styles on all paragraphs

To test:
 - Start the demo app
 - Switch to html mode and back
 - See that all styles are still there


jleandroperez:
Closing this one since it was addressed via #499.

Thanks Sergio!

PR #502, opened 2017-06-09
HTML Generation: Merging to Left!

jleandroperez:
### Details:
In this PR we're implementing Merge-to-Left: as we parse different Paragraphs, we're now merging N and N-1 nodes, as long as `canMergeChildren`'s `Node` (new helper) returns true.

Pendings:
- Newlines Logic
- Merge Explicit Unit Tests
- (And of course!) revisit the Merge Logic

Ref #492

### To test:
1. Verify the Unit Tests and make sure *NSAttributedStringToNodesTests* are green.
2. Try inserting a List with multiple items, and verify the `LI` items get grouped together under the same `OL / UL`

cc @diegoreymendez 
Thanks in advance!!

### DIFF
--- a/Aztec/Classes/Converters/NSAttributedStringToNodes.swift
+++ b/Aztec/Classes/Converters/NSAttributedStringToNodes.swift
@@ -159,27 +159,19 @@ private extension NSAttributedStringToNodes {
     ///
     func createLeafNodes(from attrString: NSAttributedString) -> [Node] {
         let attachment = attrString.attribute(NSAttachmentAttributeName, at: 0, effectiveRange: nil) as? NSTextAttachment
-        var leafs = [Node]()
 
         switch attachment {
         case let lineAttachment as LineAttachment:
-            let node = lineAttachmentToNode(lineAttachment)
-            leafs.append(node)
+            return lineAttachmentToNodes(lineAttachment)
         case let commentAttachment as CommentAttachment:
-            let node = commentAttachmentToNode(commentAttachment)
-            leafs.append(node)
+            return commentAttachmentToNodes(commentAttachment)
         case let htmlAttachment as HTMLAttachment:
-            let nodes = htmlAttachmentToNode(htmlAttachment)
-            leafs.append(contentsOf: nodes)
+            return htmlAttachmentToNode(htmlAttachment)
         case let imageAttachment as ImageAttachment:
-            let node = imageAttachmentToNode(imageAttachment)
-            leafs.append(node)
+            return imageAttachmentToNodes(imageAttachment)
         default:
-            let nodes = textToNode(attrString.string)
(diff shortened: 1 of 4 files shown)


Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -59,17 +79,96 @@ class NSAttributedStringToNodes: Converter {
...
+            return false
+        }
+
+        target.children += source.children
+
+        return true
+    }
+
+
+    /// Returns the "Right Most" Blocklevel Node from a collection fo nodes.
```
diegoreymendez: Small grammar note: rightmost and leftmost are single words.

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -59,17 +79,96 @@ class NSAttributedStringToNodes: Converter {
...
+    ///
+    func rightMostParagraphStyleElements(from nodes: [Node]) -> [ElementNode] {
```
diegoreymendez: If you agree with what I wrote above, this should turn into `rightmostParagraphStyleElements `.

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -59,17 +79,96 @@ class NSAttributedStringToNodes: Converter {
...
+        return paragraphStyleElements(from: nodes) { children in
+            return children.last
+        }
+    }
+
+
+    /// Returns the "Left Most" Blocklevel Node from a collection fo nodes.
+    ///
+    func leftMostParagraphStyleElements(from nodes: [Node]) -> [ElementNode] {
```
diegoreymendez: If you agree with what I wrote above, this should turn into `leftmostParagraphStyleElements`.

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
```diff
@@ -199,6 +201,27 @@ extension Libxml2 {
...
+        ///     - node: Target node for which we'll determine Merge-ability status.
+        ///
+        /// - Returns: true if both nodes can be merged, or not.
+        ///
+        func canMergeChildren(of node: ElementNode) -> Bool {
+            guard name == node.name && Set(attributes) == Set(node.attributes) else {
+                return false
+            }
+
+            guard let standardName = self.standardName else {
```
diegoreymendez: I don't think the merge should fail just because the element has no standard name.

The truth is that the code below this point will need to be changed, so if you prefer to avoid analyzing this further now, we can proceed as it is and analyze in a second PR.

jleandroperez:
Thanks Diego!!

PR #507, opened 2017-06-12
Fix: testBlockquoteToggle2 Test

jleandroperez:
### Details:
In this (super small PR) we're updating a Unit Test condition. In order to test it, please, verify `testBlockquoteToggle2` is green.

Ref #492 

### Scenario:
1. Type `Hello`
2. Enter newline
3. Toggle Blockquote
4. Type `World`

Verify that the generated HTML looks like: `Hello <blockquote>World</blockquote>`. No newline in beteween!


jleandroperez:
Thank you!

PR #508, opened 2017-06-13
492: Style hierarchy

SergioEstevao:
This PR implements a new paragraph style hierarchy.

All the attributes of a paragraph (header level, lists, blockquotes) are now in a hierarchy and they can be nested like in HTML.
The base class of the properties is the `ParagraphProperties` class. This class contains an optional `HTMLRepresentation` that can be used to store the original HTML attributes and element used.

I kept the old properties available but only has getters to allow a smooth transition of the old code, specially the list layout code.

To test:
 - Launch the demo app
 - See if all content shows up correctly
 - Apply toggle of the toolbar and check if they work correctly: Header, Lists, Blockquotes, etc..


Aztec/Classes/TextKit/Header.swift
```diff
@@ -0,0 +1,71 @@
+import Foundation
+import UIKit
+
+
+// MARK: - Text List
```
frosty: Wrong comment?

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -220,6 +244,41 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
...
+
+    func add(property: ParagraphProperty) {
+        properties.append(property)
+    }
+
+    func removeProperty(ofType type: AnyClass) {
+        var position: Int?
+        for index in (0..<properties.count).reversed() {
+            if type(of: properties[index]) == type {
+                position = index
```
diegoreymendez: Any reason not to remove the property right here?  It would make for shorter / cleaner code, and since we're breaking out of the loop immediately it shouldn't break the enumerator.

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -220,6 +244,41 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
...
+        if let positionFound = position {
+            properties.remove(at: positionFound)
+        }
+    }
+
+    func replaceProperty(ofType type: AnyClass, with newProperty: ParagraphProperty) {
+        var position: Int?
+        for index in (0..<properties.count).reversed() {
+            if type(of: properties[index]) == type {
+                position = index
```
diegoreymendez: Same comment as above.

SergioEstevao:
@diegoreymendez update the code addressing your comments.

PR #510, opened 2017-06-13
NSAttributedString >> Nodes: Updating Merge Logic

jleandroperez:
Ref. #492

### To test:
- Verify that the Unit Tests are okay
- Review the `merge(left:, right:)` logic please!
- Play with the empty editor, and verify that:
  - Last Mergeable item should never get merged
  - Last LI item is never merged

cc @diegoreymendez 

Thanks in advance!!

### DIFF
--- a/Aztec/Classes/Converters/NSAttributedStringToNodes.swift
+++ b/Aztec/Classes/Converters/NSAttributedStringToNodes.swift
@@ -95,7 +95,9 @@ private extension NSAttributedStringToNodes {
     /// - The remaining elements will get merged
     ///
     func merge(left: [ElementNode], right: [ElementNode]) -> Bool {
-        guard let mergeableNodes = findMergeableNodes(left: left, right: right)?.dropLast() else {
+        guard let mergeableNodes = findMergeableNodes(left: left, right: right)?.dropLast(),
+            !mergeableNodes.isEmpty
+        else {
             return false
         }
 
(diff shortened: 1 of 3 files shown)


Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -89,8 +89,17 @@ private extension NSAttributedStringToNodes {
...
     /// Attempts to merge the Right array of Element Nodes (Paragraph Level) into the Left array of Nodes.
     ///
+    /// - We expect two collections of Mergeable Elements: Paragraph Level, with matching Names + Attributes
+    /// - Last LI item is never merged
+    /// - Last 'Mergeable' element is never merged (ie. <h1>Hello\nWorld</h1> >> <h1>Hello</h1><h1>World</h1>
+    /// - The remaining elements will get merged
+    ///
     func merge(left: [ElementNode], right: [ElementNode]) -> Bool {
-        guard let (target, source) = findLowestMergeableNodes(left: left, right: right) else {
+        guard let mergeableNodes = findMergeableNodes(left: left, right: right)?.dropLast() else {
```
diegoreymendez: **Possible logical improvement:**

I think this guard should be changed to:

```swift
guard let mergeableNodes = findMergeableNodes(left: left, right: right)?.dropLast(),
     mergeableNodes.count > 0 else {
```

This is because when `findMergeableNodes()` returns 1 element, we'll not be merging anything after this point.

jleandroperez:
@diegoreymendez ready for another look, thanks!!

jleandroperez:
Woooo thanks Diego!!

PR #511, opened 2017-06-13
Fixing Image Conversion Tests

jleandroperez:
Ref. #492

### To test:
Run the Unit Tests, and verify that nothing breaks when comparing `<img src...` strings!

cc @diegoreymendez 
Thanks!!


jleandroperez:
Thanks!!!

PR #516, opened 2017-06-13
Fixing Last Standing BREAKING Unit Tests!!

jleandroperez:
Ref. #492

### To test:
- Run the unit tests. Everything *everything* should be green
- Test Inserting a List, and verify the HTML looks good

cc @diegoreymendez @SergioEstevao 
Thanks!!!


jleandroperez:
Thanks Sergio!!

PR #519, opened 2017-06-14
Output the extra attributes on a video.

SergioEstevao:
This PR fixes the only unit test that is breaking.

To test:
 - Run the app
 - See if the demo content shows up correctly
 - Run the unit test and see if all pass.


frosty:
I don't know anything about the underlying code, but I can confirm tests all pass.

Issue #521, opened 2017-06-14
Import HTML data into HTMLRepresentation elements.


diegoreymendez:
When loading the initial HTML, we need to store the HTML representation information into our `HTMLRepresentation` instances, and link those to our `NSAttributedString` styles.

diegoreymendez:
Done.

PR #523, opened 2017-06-14
Supporting Unsupported HTML

jleandroperez:
### Details:
Best Issue Title *Ever*. In this PR we're implementing a new NSAttributedString attribute, responsible of holding references to the HTML Entities, present in the original HTML, that aren't directly editable, but should be maintained.

Unit Tests coming up in a different PR.

Ref. #521 

### To test:
1. Launch the Empty Editor
2. Switch to HTML mode
3. Enter `<span class="i am"><span class="evil>YEAH!</span></span>`
4. Switch back to Rich Mode
5. Switch back to HTML mode

Verify that the SPAN entities are there.

cc @diegoreymendez 
Thanks in advance!!

### DIFF
--- a/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
+++ b/Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
@@ -351,7 +351,7 @@ private extension HTMLNodeToNSAttributedString {
         if let elementFormatter = formatter(for: elementRepresentation) {
             finalAttributes = elementFormatter.apply(to: finalAttributes, andStore: elementRepresentation)
         } else {
-            finalAttributes = store(elementRepresentation: elementRepresentation, in: finalAttributes)
+            finalAttributes = self.attributes(storing: elementRepresentation, in: finalAttributes)
         }
 
         for attributeRepresentation in elementRepresentation.attributes {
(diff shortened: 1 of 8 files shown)


Aztec/Classes/NSAttributedString/Styles/UnsupportedHTML.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+
+// MARK: - UnsupportedHTML NSAttributedString Attribute Name
+//
+let UnsupportedHTMLAttributeName = "UnsupportedHTMLAttributeName"
```
diegoreymendez: Since this is part of the `UnsupportedHTML` logic... why not make it a static constant named `UnsupportedHTML.attributeName`?

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -382,6 +383,25 @@ private extension HTMLNodeToNSAttributedString {
...
+
+    /// Stores the specified HTMLElementRepresentation in a collection of NSAttributedString Attributes.
+    ///
+    /// - Parameters:
+    ///     - elementRepresentation: Instance of HTMLElementRepresentation to be stored.
+    ///     - attributes: Attributes where we should store the HTML Representation.
+    ///
+    /// - Returns: A collection of NSAttributedString Attributes, including the specified HTMLElementRepresentation.
+    ///
+    private func store(elementRepresentation: HTMLElementRepresentation, in attributes: [String: Any]) -> [String: Any] {
```
diegoreymendez: I'd consider renaming this to:

```swift
private func attributes(storing elementRepresentation: HTMLElementRepresentation, in attributes: [String: Any]) -> [String: Any] {
```

jleandroperez:
Thanks for the review sir!!!

PR #527, opened 2017-06-15
FormatBar: Update list and heading pickers

frosty:
This PR improves the appearance and behaviour of the list and heading pickers in the format bar.

* We now present options for Lists and well as Headers
* Both are based on their respecting formatting types, with some extensions to provide names and icons.
* Menus now contain icons, and reflect the state of the currently selected format.
* Some improvements to the behaviour when switching between lists:
  * If you tap an item (e.g. Header), and then tap it again, the menu will be hidden.
  * If you tap an item (e.g. Header), and then another item that shows a picker (e.g. List), the picker will switch to the other picker (previously it would be hidden).

screen shot 2017-06-15 at 14 06 06

screen shot 2017-06-15 at 14 06 18

To test:

* Check the interactions described above
* Check switching between heading types, and list types
* Test hiding and then re-showing the various menus

Needs review: @jleandroperez


frosty:
Thanks @SergioEstevao!

PR #532, opened 2017-06-16
FormatBar: Add popovers for pickers on iPad

frosty:
Follows on from #527. This PR displays the Header and List pickers as popovers on iPad. We're doing it at any size class, because it looks fine even when the view is compact, and even if the app is narrow the keyboard is still full width.

It's not my favourite implementation ever, but I feel like popovers are only a stop gap. I'd prefer to use custom views in the future and not actually 'present' another view controller. Let me know what you think.

To test:

* Run the demo app on iPad, and try using the heading and list menus. They should reflect the current state, and dismiss if you tap away from them or interact with the format bar. 
* They should work correctly no matter what size the app is at, and they should dismiss if you rotate the device.

Needs review: @SergioEstevao or @jleandroperez would one of you mind taking a look?

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -199,13 +199,7 @@ class EditorDemoController: UIViewController {
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
 
-        if optionsViewController != nil {
-            if presentedViewController == optionsViewController {
-                dismiss(animated: true, completion: nil)
-            }
-
-            optionsViewController = nil
-        }
+        dismissOptionsViewControllerIfNecessary()
     }
 
     // MARK: - Title and Title placeholder position methods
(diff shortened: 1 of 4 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -591,25 +608,62 @@ extension EditorDemoController : Aztec.FormatBarDelegate {
...
+            } else {
+                present(optionsViewController, animated: true, completion: selectRow)
+            }
+
+        } else {
+            self.addChildViewController(optionsViewController)
+            changeRichTextInputView(to: optionsViewController.view)
+            optionsViewController.didMove(toParentViewController: self)
+            selectRow()
         }
```
jleandroperez: Can we split this method into a couple smaller? Couple ideas below!

A. Show Options could call one of these two in the end:
```
- displayOptionsOnPad(optionsViewController:...
- displayOptionsOnPhone
```

B. potentially a method just dedicated to instantiating the Options VC?

frosty: Thanks for the suggestion! I've just pushed a tweak – could you take one more look please?

jleandroperez:
Merging right away, so that we can test this in the WPiOS + Aztec integration.
Thanks for taking good care of this @frosty, beautiful work!!

:shipit:

PR #544, opened 2017-06-21
FormatBar: A couple of RTL improvements

frosty:
This PR makes a few improvements to right-to-left support for the format bar. Previously, the overflow toggle was hidden, so it was impossible to show the extra items in the bar. Now it should be visible and usable:

formatbar-rtl

There is still an issue where the items in the bar are left aligned instead of right aligned, but I haven't been able to get to the bottom of that yet. At least, with this PR, the entire bar becomes usable. Hopefully we can address the alignment in a future PR.

Needs review: @jleandroperez

### DIFF
--- a/Example/Example/MoreAttachmentRenderer.swift
+++ b/Example/Example/MoreAttachmentRenderer.swift
@@ -30,7 +30,11 @@ extension MoreAttachmentRenderer: TextViewAttachmentImageProvider {
         UIGraphicsBeginImageContextWithOptions(size, false, 0)
 
         let label = attachment.text.uppercased()
-        let attributes = [NSForegroundColorAttributeName: textColor]
+
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.baseWritingDirection = .leftToRight
+        let attributes: [String: Any] = [NSForegroundColorAttributeName: textColor, NSParagraphStyleAttributeName: paragraphStyle]
+
         let colorMessage = NSAttributedString(string: label, attributes: attributes)
 
         let textRect = colorMessage.boundingRect(with: size, options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil)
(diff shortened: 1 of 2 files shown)


frosty:
@SergioEstevao I fixed the MORE text! Had to hardcode it to `leftToRight` to make it appear – which shouldn't be an issue for RTL, as it'll always be centered anyway.

Example/Example/MoreAttachmentRenderer.swift
```diff
@@ -30,7 +30,11 @@ extension MoreAttachmentRenderer: TextViewAttachmentImageProvider {
         UIGraphicsBeginImageContextWithOptions(size, false, 0)
 
         let label = attachment.text.uppercased()
-        let attributes = [NSForegroundColorAttributeName: textColor]
+
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.baseWritingDirection = .leftToRight
```
SergioEstevao: this changes depending of the current local properties? If the more string is localised this can force the writing to be always left to right...

frosty: Does it matter in this case, if we always center the text between the dividers?

diegoreymendez: I know auto-layout is a bit of a headache, but it certainly feels like we're missing something if we need to force this setting here.

That said, workarounds can be acceptable if we document them explaining why they're needed.  Maybe even adding a link to this PR or the discussion.

The reason being: the next developer working on this code may very well remove this line without knowing what's affected by it.

SergioEstevao: @frosty humm, I'm just not sure is regarding if on an RTL language the .leftToRight forces the text to be written MORE instead of EROM...

SergioEstevao:
I'm merging this one because it's good enough for now, and we can address further RTL issues on another PR.

PR #557, opened 2017-06-22
Shift + enter no longer breaks paragraphs.

### LINKED ISSUE #553: Shift + enter should not interrupt paragraphs.
Shift+enter should not interrupt paragraphs, and instead produce a new line separated from the previous one by a line separator character (U+2028).

In lists it should produce a new line that's part of the same bullet as the previous line.

diegoreymendez:
Fixes #553 

To test:

* Launch the Demo with data, and make sure the lists look fine.
* Launch the empty Demo, and insert a list.
* Pressing shift + enter should no longer break paragraphs, and produce `<br>` elements in the HTML.


diegoreymendez:
Thank you @SergioEstevao !

PR #559, opened 2017-06-22
Remove gridicons references from the main Aztec pod.

SergioEstevao:
Only keep references that are done on the demo project.

There is only things that I see as issues if we remove Gridicons from the pod:

 - We don't get a default Missing Image on the TextView -> but this could be set for however is using the pod
 - We don't get a play icon on the the Video attachment, I was considering implementing a default play button icon using some CoreGraphics code ( a triangle with round corners)

If we agree with this solution I can go ahead and implement the play icon and remove the Gridicon dependency.

To test:
 - Build the pod
 - Check if all tests build and pass
 - Build the example
 - Run the example and check if nothing breaks.


diegoreymendez:
While having no dependencies is very convenient, I don't want to do it in detriment of not having default images.

The only scenario in which I'd consider removing those defaults is if we could force the client app to set those defaults.

Unfortunately, I don't think it's very feasible since our TextView could be instantiated in a number of different ways (in some cases with the initializer not taking any input parameters from the client app).

SergioEstevao:
To be honest at the moment if you don't set an attachment delegate the app will assert crash when trying to render an image. And that delegate is also not initiated by default.

The default image could be one of the pre-conditions also.

diegoreymendez:
@SergioEstevao - I'm ok with having the client app provide the default images through a delegate, but if that's the idea I'd need us to do the full conversion.

That is:
- You'd have to add the delegates and assertions in this PR,
- Once it's merged you'd have to integrate it into WPiOS to make sure we provide those defaults.

If that sounds fine then let me know when the changes are up.

SergioEstevao:
@diegoreymendez and/or @jleandroperez ready for another look.

AztecTests/Formatters/BlockquoteFormatterTests.swift
```diff
@@ -101,7 +100,7 @@ class BlockquoteFormatterTests: XCTestCase {
 
 private extension BlockquoteFormatterTests {
     var testTextView: TextView {
-        let view = TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.image))
+        let view = TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewStubAttachmentDelegate.swift
```diff
@@ -17,11 +17,11 @@ class TextViewStubAttachmentDelegate: TextViewAttachmentDelegate {
         let placeholderImage: UIImage
         switch attachment {
         case _ as ImageAttachment:
-            placeholderImage = Gridicon.iconOfType(.image, withSize: imageSize)
+            placeholderImage = UIImage()
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewStubAttachmentDelegate.swift
```diff
@@ -17,11 +17,11 @@ class TextViewStubAttachmentDelegate: TextViewAttachmentDelegate {
...
         case _ as VideoAttachment:
-            placeholderImage = Gridicon.iconOfType(.video, withSize: imageSize)
+            placeholderImage = UIImage()
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewStubAttachmentDelegate.swift
```diff
@@ -17,11 +17,11 @@ class TextViewStubAttachmentDelegate: TextViewAttachmentDelegate {
...
         default:
-            placeholderImage = Gridicon.iconOfType(.attachment, withSize: imageSize)
+            placeholderImage = UIImage()
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewTests.swift
```diff
@@ -25,13 +24,13 @@ class TextViewTests: XCTestCase {
     // MARK: - TextView construction
 
     func createEmptyTextView() -> Aztec.TextView {
-        let richTextView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.attachment))
+        let richTextView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewTests.swift
```diff
@@ -25,13 +24,13 @@ class TextViewTests: XCTestCase {
...
         richTextView.textAttachmentDelegate = attachmentDelegate
         return richTextView
     }
 
     func createTextView(withHTML html: String) -> Aztec.TextView {
-        let richTextView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.attachment))
+        let richTextView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewTests.swift
```diff
@@ -40,7 +39,7 @@ class TextViewTests: XCTestCase {
 
     func createTextViewWithContent() -> Aztec.TextView {
         let paragraph = "Lorem ipsum dolar sit amet.\n"
-        let richTextView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.attachment))
+        let richTextView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewTests.swift
```diff
@@ -58,7 +57,7 @@ class TextViewTests: XCTestCase {
 
     func testTextViewReferencesStorage() {
 
-        let textView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.attachment))
+        let textView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewTests.swift
```diff
@@ -79,7 +78,7 @@ class TextViewTests: XCTestCase {
     // MARK: - Test Index Wrangling
 
     func testMaxIndex() {
-        let textView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.attachment))
+        let textView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/TextViewTests.swift
```diff
@@ -92,7 +91,7 @@ class TextViewTests: XCTestCase {
     }
 
     func testAdjustedIndex() {
-        let textView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: Gridicon.iconOfType(.attachment))
+        let textView = Aztec.TextView(defaultFont: UIFont.systemFont(ofSize: 14), defaultMissingImage: UIImage())
```
diegoreymendez: We're removing the default here, but not assigning a new one.

AztecTests/Formatters/BlockquoteFormatterTests.swift
SergioEstevao: Test file

AztecTests/TextViewStubAttachmentDelegate.swift
SergioEstevao: Test file.

AztecTests/TextViewStubAttachmentDelegate.swift
SergioEstevao: Test File.

AztecTests/TextViewTests.swift
SergioEstevao: Test file.

AztecTests/TextViewTests.swift
SergioEstevao: Test file.

AztecTests/TextViewTests.swift
SergioEstevao: Test File

AztecTests/TextViewTests.swift
SergioEstevao: Test file.

AztecTests/TextViewTests.swift
SergioEstevao: Test file.

AztecTests/TextViewTests.swift
SergioEstevao: Test file.

PR #560, opened 2017-06-22
Obliterating Fragmentation: Mark I

jleandroperez:
### Details:
In this initial PR, we're updating the NSAttributedStringToNodes helper's (priavte API's), so that we can address the Fragmentation Issue in a simple way.

Plus: since i was around, added support to serialize Pre elements (+ unit tests!!)

### Testing:
1. Run the unit tests
2. Run a smoke test over the Out Converter

cc @diegoreymendez


jleandroperez:
Closed in favor of #567

PR #562, opened 2017-06-22
Fix image drawing on paragraphs with list and/or block quotes

### LINKED ISSUE #545: Image are draw out of screen when inserted inside a blockquote or list
### Expected behavior

 - Start the demo app
 - Write a paragraph
 - Set the paragraph has blockquote and list
 - Add an image to it
 - The image should be draw inside the screen taking in account the block quotes and list margin

### Actual behavior
 - The image is draw a bit outside of the margins

screen shot 2017-06-21 at 14 49 39

SergioEstevao:
Fixes #545 

To test:
 - Start the empty demo app
 - Insert one image on the content
 - On the same line insert another image
 - Check that both image show up 
 - add a list style
 - Check that both image show up inside the textview bounds
 -  add a block quote style
 - Check that both image show up inside the textview bounds

### DIFF
--- a/Aztec/Classes/TextKit/MediaAttachment.swift
+++ b/Aztec/Classes/TextKit/MediaAttachment.swift
@@ -295,8 +295,12 @@ open class MediaAttachment: NSTextAttachment
             return CGRect.zero
         }
 
-        let padding = textContainer?.lineFragmentPadding ?? 0
-        let width = floor((lineFrag.width - position.x ) - ( padding * 2))
+        var padding = (textContainer?.lineFragmentPadding ?? 0)
+        if let storage = textContainer?.layoutManager?.textStorage,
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -295,8 +295,12 @@ open class MediaAttachment: NSTextAttachment
             return CGRect.zero
         }
 
-        let padding = textContainer?.lineFragmentPadding ?? 0
-        let width = floor((lineFrag.width - position.x ) - ( padding * 2))
+        var padding = (textContainer?.lineFragmentPadding ?? 0)
```
diegoreymendez: There's no need to add parentheses here.

PR #566, opened 2017-06-23
Add keyboard shortcuts.

SergioEstevao:
This PR adds keyboard shortcuts to the editor

To test:
 - Run the test app on the iPad
 - On the editor tap on the CMD key
 - See if shorchuts show in screen
 - Try the shortcuts and see if they are working correctly.

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -411,6 +411,26 @@ class EditorDemoController: UIViewController {
         toolbar.selectItemsMatchingIdentifiers(identifiers)
     }
 
+    override var keyCommands: [UIKeyCommand] {
+        if richTextView.isFirstResponder {
+            return [ UIKeyCommand(input:"B", modifierFlags: .command, action:#selector(toggleBold), discoverabilityTitle:NSLocalizedString("Bold", comment: "Discoverability title for bold formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"I", modifierFlags: .command, action:#selector(toggleItalic), discoverabilityTitle:NSLocalizedString("Italic", comment: "Discoverability title for italic formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"D", modifierFlags: [.command, .alternate], action:#selector(toggleStrikethrough), discoverabilityTitle: NSLocalizedString("Strikethrough", comment:"Discoverability title for strikethrough formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"U", modifierFlags: .command, action:#selector(EditorDemoController.toggleUnderline(_:)), discoverabilityTitle: NSLocalizedString("Underline", comment:"Discoverability title for underline formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"Q", modifierFlags:[.command,.alternate], action: #selector(toggleBlockquote), discoverabilityTitle: NSLocalizedString("Block Quote", comment: "Discoverability title for block quote keyboard shortcut.")),
+                     UIKeyCommand(input:"K", modifierFlags:.command, action:#selector(toggleLink), discoverabilityTitle: NSLocalizedString("Insert Link", comment: "Discoverability title for insert link keyboard shortcut.")),
+                     UIKeyCommand(input:"M", modifierFlags:[.command,.alternate], action:#selector(showImagePicker), discoverabilityTitle: NSLocalizedString("Insert Media", comment: "Discoverability title for insert media keyboard shortcut.")),
+                     UIKeyCommand(input:"U", modifierFlags:[.command, .alternate], action:#selector(toggleUnorderedList), discoverabilityTitle:NSLocalizedString("Bullet List", comment: "Discoverability title for bullet list keyboard shortcut.")),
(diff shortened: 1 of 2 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -411,6 +411,26 @@ class EditorDemoController: UIViewController {
         toolbar.selectItemsMatchingIdentifiers(identifiers)
     }
 
+    override var keyCommands: [UIKeyCommand] {
+        if richTextView.isFirstResponder {
+            return [ UIKeyCommand(input:"B", modifierFlags: .command, action:#selector(toggleBold), discoverabilityTitle:NSLocalizedString("Bold", comment: "Discoverability title for bold formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"I", modifierFlags: .command, action:#selector(toggleItalic), discoverabilityTitle:NSLocalizedString("Italic", comment: "Discoverability title for italic formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"D", modifierFlags: [.command, .alternate], action:#selector(toggleStrikethrough), discoverabilityTitle: NSLocalizedString("Strikethrough", comment:"Discoverability title for strikethrough formatting keyboard shortcut.")),
+                     UIKeyCommand(input:"U", modifierFlags: .command, action:#selector(EditorDemoController.toggleUnderline(_:)), discoverabilityTitle: NSLocalizedString("Underline", comment:"Discoverability title for underline formatting keyboard shortcut.")),
```
frosty: How come this one is namespaced `EditorDemoController`?

Example/Example/EditorDemoController.swift
```diff
@@ -411,6 +411,26 @@ class EditorDemoController: UIViewController {
...
+                     UIKeyCommand(input:"M", modifierFlags:[.command,.alternate], action:#selector(showImagePicker), discoverabilityTitle: NSLocalizedString("Insert Media", comment: "Discoverability title for insert media keyboard shortcut.")),
+                     UIKeyCommand(input:"U", modifierFlags:[.command, .alternate], action:#selector(toggleUnorderedList), discoverabilityTitle:NSLocalizedString("Bullet List", comment: "Discoverability title for bullet list keyboard shortcut.")),
+                     UIKeyCommand(input:"O", modifierFlags:[.command, .alternate], action:#selector(toggleOrderedList), discoverabilityTitle:NSLocalizedString("Numbered List", comment:"Discoverability title for numbered list keyboard shortcut.")),
+                     UIKeyCommand(input:"H", modifierFlags:[.control, .shift], action:#selector(toggleEditingMode), discoverabilityTitle:NSLocalizedString("Toggle HTML Source ", comment: "Discoverability title for HTML keyboard shortcut."))
```
frosty: In WPiOS, we currently use `Cmd` + Shift + H for HTML, not `Ctrl`. Is there a reason this is different?

Example/Example/EditorDemoController.swift
SergioEstevao: Ask Swift :). It was getting confused with a method with the same name in super... This was the way to solve it :).

Example/Example/EditorDemoController.swift
SergioEstevao: Ops, I hacked this in because in the simulator that is the shortcut for the Home Key, will change it now :)

Example/Example/EditorDemoController.swift
frosty: Okay 👍

Example/Example/EditorDemoController.swift
```diff
@@ -411,6 +411,26 @@ class EditorDemoController: UIViewController {
...
+            ]
+        } else if htmlTextView.isFirstResponder {
+            return [UIKeyCommand(input:"H", modifierFlags:[.control, .shift], action:#selector(toggleEditingMode), discoverabilityTitle:NSLocalizedString("Toggle HTML Source ", comment: "Discoverability title for HTML keyboard shortcut."))
```
frosty: Don't forget this one too :)

SergioEstevao: It's done.

frosty:
Looking good, Sergio – two more comments:

* We should probably switch from using Cmd + Opt + D for strikethrough, as that brings up the Dock on iOS 11.
* If you activate HTML mode with the keyboard shortcut, and the format bar isn't expanded, the _only_ way to get out of HTML mode is the keyboard shortcut. Perhaps if you trigger it and the keyboard is not expanded, we should expand the keyboard. Alternatively, perhaps we should leave the toggle button enabled when in HTML mode.

SergioEstevao:
@frosty 

Cmd + Opt + S for strikethrough?

So can't they use the toolbar to switch back again?

diegoreymendez:
> If you activate HTML mode with the keyboard shortcut, and the format bar isn't expanded, the only way to get out of HTML mode is the keyboard shortcut.

That's more like a problem with the toolbar, though.

IMHO the HTML button should not be in the toolbar as it should be possible to switch regardless of if the caret is placed in the text view or not.

Either that, or the toolbar should at least be always visible, even if the caret is not in place.

Ping @iamthomasbishop ^

SergioEstevao:
So for strikethrough with went for Cmd+S to be inline with Bold, Italic, Underline.

I also did some changes to make the toolbar expand when switching to HTML mode.

Mind to have another look @frosty ?

Issue #485, opened 2017-05-31
The default paragraph style is sometimes lost.


diegoreymendez:
While editing, the default paragraph style is sometimes lost, as can be seen in the smaller text near the bottom of the following screenshot:

![img_0029](https://cloud.githubusercontent.com/assets/1836005/26638115/b4837536-45f7-11e7-9154-123c31f6624f.png)

Thanks to @catehstn for reporting.

jleandroperez:
Closing this one as cannot repro. Probably solved via #758.

Thanks!

diegoreymendez:
Agreed on #758.  Tagging this.

PR #569, opened 2017-06-26
Make sure attributes are applied to attachment insertions/replacements.

SergioEstevao:
Refs #485 

This PR makes sure the typing attributes are applied when inserting attachments (media, more, hr) to the content.

To test:
 - Start the demo app
 - Insert a image on the start of the content
 - Write just after image and see if the text has the correct style.


SergioEstevao:
@diegoreymendez good suggestion, I updated the code please have a look.

PR #572, opened 2017-06-28
Implement undo and redo when removing attachments by code.

SergioEstevao:
This PR implements the capability of undo redo when removing an attachment by pressing on it and select the option remove.

To test:
 - Start the demo app
 - open the demo with content
 - scroll to the coyote image
 - Tap on it
 - Tap again
 - Select Remove Media
 - Check if image was removed
 - Tap Undo 
 - Check if image comes back again.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -972,7 +972,7 @@ open class TextView: UITextView {
         return storage.attachment(withId: id)
     }
 
-    /// Removes the attachment that match the attachment identifier provided from the storage
+    /// Removes the attachment that matches the attachment identifier provided from the storage
     ///
     /// - Parameter attachmentID: the unique id of the attachment
     ///
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -972,18 +972,28 @@ open class TextView: UITextView {
         return storage.attachment(withId: id)
     }
 
-    /// Removes the attachments that match the attachament identifier provided from the storage
+    /// Removes the attachment that match the attachment identifier provided from the storage
     ///
     /// - Parameter attachmentID: the unique id of the attachment
     ///
     open func remove(attachmentID: String) {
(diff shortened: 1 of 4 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -972,18 +972,28 @@ open class TextView: UITextView {
...
+    /// Removes the attachment that match the attachment identifier provided from the storage
     ///
     /// - Parameter attachmentID: the unique id of the attachment
     ///
     open func remove(attachmentID: String) {
-        storage.remove(attachmentID: attachmentID)
+        guard let range = storage.rangeFor(attachmentID: attachmentID) else {
+            return
+        }
+        let originalText = storage.attributedSubstring(from: range)
```
diegoreymendez: From line 983 to line 990 we're repeating code that's in several other places in `TextView.swift`

Why not implement in `TextView.swift` method

```
func replaceCharacters(in: range, with attrString: NSAttributedString)
```

 to take care of unified undo support?

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -972,18 +972,28 @@ open class TextView: UITextView {
         return storage.attachment(withId: id)
     }
 
-    /// Removes the attachments that match the attachament identifier provided from the storage
+    /// Removes the attachment that match the attachment identifier provided from the storage
     ///
     /// - Parameter attachmentID: the unique id of the attachment
     ///
     open func remove(attachmentID: String) {
-        storage.remove(attachmentID: attachmentID)
```
diegoreymendez: This method is no longer called from anywhere in the code.  We should remove it from `TextStorage.swift`.

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: Good suggestion may I do this in another PR just dedicated to that refactor?

diegoreymendez: Sure.

PR #573, opened 2017-06-28
Add an extra check for alternative emoji font.

SergioEstevao:
Refs #485 

This PR adds another check for alternative emoji font when not using the standard system font (San Francisco UI)

To test:
 - Make sure unit tests are ok
 - Run the demo app
 - Insert an emoji anywhere in the text
 - Make sure that characters written after have the correct style

### DIFF
--- a/AztecTests/TextViewTests.swift
+++ b/AztecTests/TextViewTests.swift
@@ -38,10 +38,10 @@ class TextViewTests: XCTestCase {
         return richTextView
     }
 
-    let anotherFont = UIFont(name:"HelveticaNeue", size: 14)!
+    let nonStandardSystemFont = UIFont(name:"HelveticaNeue", size: 14)!
 
-    func createEmptyTextViewWithAnotherFont() -> Aztec.TextView {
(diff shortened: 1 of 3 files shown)


AztecTests/TextViewTests.swift
```diff
@@ -38,6 +38,15 @@ class TextViewTests: XCTestCase {
         return richTextView
     }
 
+    let anotherFont = UIFont(name:"HelveticaNeue", size: 14)!
```
diegoreymendez: "anotherFont" feels very generic and is not very clear for maintainability.

Can we add a comment explaining why this is used, plus probably rename this to something like `nonStandardSystemFont`?

AztecTests/TextViewTests.swift
```diff
@@ -38,6 +38,15 @@ class TextViewTests: XCTestCase {
...
+
+    func createEmptyTextViewWithAnotherFont() -> Aztec.TextView {
```
diegoreymendez: Rename this method following the new name for `anotherFont` above.

SergioEstevao:
@diegoreymendez mind to have another look?

PR #576, opened 2017-06-29
Add resources to pod spec file.

### LINKED ISSUE #577: New assets icons on the project are not available for users of the pod
### Expected behavior
 - When using the pod on another project the assets should be available and load correctly

### Actual behavior
 - When using the pod on another project the app crashes when trying to load the resources of the Aztec pod. Default Image placeholder and play icon.

 - On the main WPiOS app
 - Open the pod file and refer the develop Aztec branch
 - Run pod install
 - Start the app
 - Start a new post
 - Add a video to the post
 - See the app crashing when trying to load the default play icon for video.

SergioEstevao:
Closes #577 

This issue resulted from the Removal of Gridicons done here: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/559

The resources where not added to the pod, so this made application that used the pod crash when trying to access a resource.

To test:
 - The easiest way to test this is to try this version directly on the main WPiOS app.

Note: @diegoreymendez because we are using CocoaPods should convert the Example app to use CocoaPods instead of Carthage? This will make it easier to find this kind of issue.

### DIFF
--- a/WordPress-Aztec-iOS.podspec
+++ b/WordPress-Aztec-iOS.podspec
@@ -31,6 +31,7 @@ Pod::Spec.new do |s|
 
   s.module_name = "Aztec"
   s.source_files = 'Aztec/Classes/**/*'
+  s.resources = 'Aztec/Assets/**/*'
 
   # For more info about these, see: https://medium.com/swift-and-ios-writing/using-a-c-library-inside-a-swift-framework-d041d7b701d9#.wohyiwj5e
   # For this to work on local/development pods and outside projects we added two paths one for each scenario. See here: https://github.com/CocoaPods/CocoaPods/issues/5375


diegoreymendez:
> because we are using CocoaPods should convert the Example app to use CocoaPods instead of Carthage? This will make it easier to find this kind of issue.

WordPress is using the library through CocoaPods, but that doesn't mean this library is only intended for use in CocoaPods.

I haven't tested it yet, since WPiOS uses CocoaPods, but the idea is Aztec should be usable through Carthage as well.

SergioEstevao:
@diegoreymendez maybe two example projects? one with pods another with Carthage? 

I just suggest this because it will make it easier and earlier to detect this kind problem.

diegoreymendez:
If we had problems with CocoaPods often I could agree, but these issues are very sporadic.

The downside of having two example projects and having to keep them in synch would affect us far more often.

PR #588, opened 2017-07-06
Minor Cleanup!

jleandroperez:
### Details:
In this PR we're removing a couple extensions that aren't being used anymore, along with a few minor style updates.

### To test:
Verify that the unit tests pass!

cc @diegoreymendez


jleandroperez:
Thanks Diego!!

Issue #285, opened 2017-02-16
Support CSS Styles


diegoreymendez:
Support parsing and rendering of certain CSS styles in Aztec.

This'll have to be implemented gradually, so this issue servers the purpose of starting the discussion.

PR #591, opened 2017-07-06
WIP: adding support for CSS styles, and improving HTMLRepresentation.

diegoreymendez:
### Description:

Reengineers HTML representation to prepare for CSS Styles.

This is some base work to be able to implement #285.

### Details:

Changes:

- `HTMLRepresentation` is now an enum specifying the different types or representations we support.
- `HTMLElementRepresentation` has been modified and doesn't inherit from `HTMLRepresentation` anymore.
- `HTMLAttributeRepresentation` have been removed.
- `HTMLNodesToNSAttributedString` has several changes in method signatures.

### Testing:

**Test 1:**

Run the editor without content, and edit styles freely.  Make sure when converting to HTML everything is properly persisted.

**Test 2:**

Run the editor with content, and edit styles freely.  Make sure when converting to HTML everything is properly persisted.

**IMPORTANT:** There are some issues with editing that are not related to this PR.  Whenever you find an issue make sure to check it against `develop` as well, before reporting in this PR.


Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -575,7 +673,7 @@ private extension NSAttributedStringToNodes {
         }
 
         return unsupported.elements.reversed().flatMap({ element in
-            return element.toNode()
+            return element
         })
```
jleandroperez: I think we can safely remove the `flatMap`, since the `elements` collection can't contain nil values, and the closure isn't really applying any kind of modification.

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -487,83 +535,133 @@ private extension NSAttributedStringToNodes {
...
-        let node = representation?.toNode() ?? ElementNode(type: .u)
+        let element: ElementNode
+
+        if let representation = attributes[UnderlineFormatter.htmlRepresentationKey] as? HTMLRepresentation,
+            case let .element(representationElement) = representation {
+
+            element = representationElement.toElementNode()
+        } else {
+            element = ElementNode(type: .u)
+        }
```
jleandroperez: We can simplify by just....


```
if let (...) {
    return representationElement.toElementNode()
}

return ElementNode(type: .u)
```

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -487,83 +535,133 @@ private extension NSAttributedStringToNodes {
...
+        if let representation = attributes[StrikethroughFormatter.htmlRepresentationKey] as? HTMLRepresentation,
+            case let .element(representationElement) = representation {
+
+            element = representationElement.toElementNode()
+        } else {
+            element = ElementNode(type: .strike)
+        }
 
-        return [node]
+        return element
```
jleandroperez: (Same comment as below!) can be simplified by just returning `toElementNode()` or `ElementNode(type: .strike)`, no need to use the var!

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
diegoreymendez: These are equivalent, though.  It's not like `else` has suddenly become evil!

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
diegoreymendez: Big miss on my side.

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
jleandroperez: Just a suggestion, feel free to disregard!. Not evil, but with the suggestion's pattern you end up with much more compact code.

jleandroperez: Remember Diego: the goal of this PR is to have an equal number of (-) and (+) changes. Please consider my suggestion, since you're just +7 away from the target!

diegoreymendez:
@jleandroperez - Good observations.  Ready for round 2.

diegoreymendez:
I'm merging since this is going into Beta 9, which still has a week to go.

In any case feel free to add comments if there's anything @SergioEstevao.

PR #596, opened 2017-07-10
TextView: Cleanup

jleandroperez:
### Details:
- Nuked a dead helper
- Updated private extension method(s) visibility
- Updated several method docs here and there
- No logic was altered at all!

cc @diegoreymendez 

In order to test: please, verify the unit tests work smoothly!


jleandroperez:
Thank you Diego!

PR #597, opened 2017-07-10
TextView: Nuking paragraph styles after hitting backspace

### LINKED ISSUE #443: Blockquote + Lists: Moving Style Upwards upon Deletion
### Steps:
1. Enter several empty new lines
2. Toggle [Blockquote, Ordered List, Unordered List]
3. Hit Backspace

As a result, the caret will be moved one line upwards. The Paragraph Format will be carried over as well.

Expected: the paragraph format should be deleted first, then, the caret moved upwards.

jleandroperez:
Fixes #443

To test:
1. Enter several empty new lines
2. Toggle a Paragraph Style (Blockquote OR Ordered List OR Unordered List)
3. Hit Backspace

Verify that the Paragraph Style is no longer carried over to the previous line.

cc @diegoreymendez

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -413,11 +413,11 @@ open class TextView: UITextView {
             deletedString = storage.attributedSubstring(from: deletionRange)
         }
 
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thanks Diego!

PR #599, opened 2017-07-10
Fixes an issue with shift + ENTER and adds a unit test.

### LINKED ISSUE #594: Shift + Enter is ending lists.
When the caret is in a list at end-of-file, and the user presses Shift + ENTER, the list is automatically finalized.

Before attempting a fix, please add unit tests to reproduce this scenario.

diegoreymendez:
Fixes #594.

It seems the previous fix was aimed at beta.7 and got lost at some point.  This PR brings the fix to beta.8.

Also adds a unit test to make sure there's no future regression with this issue.


diegoreymendez:
Closing, will re-create.

PR #601, opened 2017-07-11
Remove outdated documentation.

SergioEstevao:
This PR removes some outdated documentation from the repo.

To test:
 - Just check that files are removed from the project in Xcode and from the file system.


diegoreymendez:
I'm not sure if we should remove these or update the documentation.  Thoughts, @SergioEstevao and @jleandroperez?

SergioEstevao:
I think we should remove this, and then create new updated documentation. 

And I think the docs should be done on the Markdown format in order to be readable directly on Github.

jleandroperez:
Maybe we only remove *EditingLogic*, since it's the one no longer valid?

diegoreymendez:
@jleandroperez - Is `UnsupportedTags.rtf` up to date?

SergioEstevao:
If we keep UnsupportedTags I can convert that to a MD document, probably even add then to the README section for now.

When the documentation grows bigger we can split in multiple files.

jleandroperez:
@diegoreymendez `UnsupportedTags.rtf` is still valid, but a bit incomplete, since there are a ton of other unsupported tags, such as:

- TABLE
- SELECT / OPTION
- (And so on!!)

PR #604, opened 2017-07-11
Issue/603 add extra image attributes

### LINKED ISSUE #252: Allow editing of the extra image attributes
Here is a list of image attributes that we should support:

- Caption
- Title
- Alt Text

Ideally we should also have way to assign any other extra arguments to img node in order to support attributes that come from the HTML and user/developers specific attributes.

### LINKED ISSUE #603: Uploaded images missing attributes
### Expected behavior
When I create a post and upload an image to it, I expect that image to be displayed on the post list and on reader, both on the apps and on the web.

### Actual behavior
The post doesn't display an image in the following places:
- Reader iOS
- Post List iOS
- Post List Calypso

The reason behind this seems to be that the uploaded image is missing the `width` attribute. As a reference here's how an image uploaded via the Hybrid editor looks like:
`<a href="https://elibudandpets.files.wordpress.com/2017/07/img_6963.jpg"></a>`
And here's one uploaded with Aztec:
``

Create a new post, upload an image, check the post on Reader and the site's Post List.

SergioEstevao:
Fixes #252 

To test:
 - Make sure the demo content is opening.
 - Make sure the unit tests are running correctly.


Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -787,6 +787,14 @@ private extension NSAttributedStringToNodes {
             element.updateAttribute(named: attribute.name, value: attribute.value)
         }
 
+        for (key,value) in attachment.extraAttributes {
+            var finalValue = value
+            if key == "class", let baseValue = element.stringValueForAttribute(named: "class"){
+                finalValue = baseValue + " " + finalValue
+            }
+            element.updateAttribute(named: key, value: Attribute.Value.string(finalValue))
```
diegoreymendez: Replace:

```
Attribute.Value.string(finalValue)
```

with just:

```
.string(finalValue)
```

Aztec/Classes/Formatters/Implementations/ImageFormatter.swift
```diff
@@ -13,30 +13,51 @@ class ImageFormatter: StandardAttributeFormatter {
     override func apply(to attributes: [String : Any], andStore representation: HTMLRepresentation?) -> [String: Any] {
 
         if let representation = representation {
-            switch(representation) {
+            switch representation {
             case .element(let element):
+                var extraAttributes = [String:String]()
+                for attributeRepresentation in element.attributes {
```
diegoreymendez: For simplicity `attributeRepresentation` should be just `attribute` here.

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -15,6 +15,10 @@ protocol MediaAttachmentDelegate: class {
 ///
 open class MediaAttachment: NSTextAttachment
 {
+    /// Attributes accessible by the user, for general purposes.
+    ///
+    open var extraAttributes = [String: String]()
```
diegoreymendez: Are we expecting this property to be overridden?

IMHO this should not be `open`.

AztecTests/TextViewTests.swift
```diff
@@ -1516,4 +1516,24 @@ class TextViewTests: XCTestCase {
 
         XCTAssertEqual(html, "")
     }
+
+    func testParseImageWithExtraAttributes() {
```
diegoreymendez: To keep in line with the other tests, you should document the test cases explaining at least in high-level what is going on inside the test.

SergioEstevao:

diegoreymendez:
I moved [the referenced issue](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/603) to [the WPiOS repository](https://github.com/wordpress-mobile/WordPress-iOS/issues/7537), since it's really not related to Aztec but to WPiOS.

I'm a bit confused by this PR fixing the reported issue, though.  This PR doesn't seem to add a "width" attribute to images.

Can you clarify a bit @SergioEstevao?

diegoreymendez:
@SergioEstevao - The problem is that the description of this PR states that it fixes the "width" parameter issue, which is not the case.  In fact the issue is not really a concern for Aztec (at least not directly: Aztec doesn't enforce a width param).

Can you create an issue explaining what needed to be done in Aztec and reference it in this PR?

As I understand it should be an explanation that we need to offer a mechanism to access and write the HTML attributes that are associated with images and / or videos.

diegoreymendez:
Actually @SergioEstevao, you can just update the description of this PR to state:

Fixes #252

designsimply:
I was unable to see options to edit image attributes such as caption, title, alt text, or class attribute using the following testing steps:

1. Open a post for editing.
1. Insert an image.
1. Tap the image, tap again for options, select "Details."
1. Try to find a way to change the image caption, title, and alt text.

<strong>Result:</strong> 

img_1374  img_1373
<sup>Tested with alittletestblog.com running WordPress 4.9-alpha-41197 via WordPress Beta Tester 1.1.2 and Jetpack Bleeding Edge | 5.2-beta-12424-fdf636d-master via Jetpack Beta Tester 2.0.3 hosted at WP Engine. Note: I also tested with a wpcom site.</sup>

Am I looking in the right place? (It's entirely possible I'm not looking in the right place.)

SergioEstevao:
@designsimply you are looking at the right place :). This PR only added the option to edit and add extra attributes to the image attachments on Aztec in terms of code capability. 
We still need to update the main app edit screen and add the alt, title, and caption options in the UI.

PR #608, opened 2017-07-12
Fixes some obscure issues when editing HTML outside of block-level elements.

diegoreymendez:
Fixes some obscure issues when editing HTML outside of block-level elements.

**To test:**

1. Using the HTML input:

```text
<br><br><h1>Test</h1>
```

2. Switch to visual mode and back.

Expected output: same as the input (no `<br>` nodes should be removed).


diegoreymendez:
Thank you @jleandroperez !

diegoreymendez:
Fixes #609

PR #610, opened 2017-07-13
Draw block quotes inside lists

### LINKED ISSUE #449: Add mixed lists and blockquote support.
We should support mixing lists and blockquotes.

SergioEstevao:
Fixes #449 

This PR adds support for drawing block quotes inside lists.

To test:
 - Run the demo app with sample content
 - Scroll to the place where lists are mixed with block quotes and see if the presentation is done correctly
 - Open the empty demo app
 - write something
 - Tap on list formatting
 - Tap on block quote formatting
 - Check if presentation is correct.

### DIFF
--- a/Aztec/Classes/TextKit/ParagraphStyle.swift
+++ b/Aztec/Classes/TextKit/ParagraphStyle.swift
@@ -121,7 +121,7 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
 
     open override var headIndent: CGFloat {
         get {
-            let extra: CGFloat = (CGFloat(lists.count) * Metrics.listTextIndentation)
+            let extra: CGFloat = (CGFloat(lists.count + blockquotes.count) * Metrics.listTextIndentation)
 
             return baseHeadIndent + extra
         }
@@ -133,7 +133,7 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
 
     open override var firstLineHeadIndent: CGFloat {
         get {
-            let extra: CGFloat = (CGFloat(lists.count) * Metrics.listTextIndentation)
+            let extra: CGFloat = (CGFloat(lists.count + blockquotes.count) * Metrics.listTextIndentation)
 
             return baseFirstLineHeadIndent + extra
         }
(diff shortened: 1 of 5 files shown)


SergioEstevao:
@jleandroperez ready for review.

jleandroperez:
Couple notes!

A. Hitting *Tab* causes the bullet to indent, but the blockquote is not moving
B. We should probably cut the "nested" style after hitting enter. IE:
  1. Add a List
  2. Toggle Blockquote
  3. Enter a line of text
  4. Hit Enter
  In this scenario we should probably remove the blockquote from the next bullet (We're currently carrying it over).

I'll keep on testing, hopefully nothing else will show up!

jleandroperez:
Another note!: We should probably add a little bit of padding between the Bullet and the Blockquote, since they look super close,

SergioEstevao:
> Another note!: We should probably add a little bit of padding between the Bullet and the Blockquote, since they look super close,

Vertically or horizontally?

SergioEstevao:
> A. Hitting Tab causes the bullet to indent, but the blockquote is not moving

Is this scenario where block quote comes first or lists come first?

jleandroperez:
@SergioEstevao Compiling all of the comments, and adding extra details on each one of them.

We can always address the rest in different issues, absolutely cool with it!  But we should probably address, at least, Comment C here, since it's related to the drawing snippets.

Thanks sir, nice job!!!

### Comment A: Nested Lists
Hitting Tab causes the bullet to indent, but the blockquote is not moving. Steps:

1. Toggle Ordered List
2. Toggle Blockquote
3. Hit Tab

As a result the Ordered List will get shifted rightwards, but the blockquote will remain in place.

### Comment B: Newline
We should probably cut the "nested" style after hitting enter. IE:

1. Add a List
2. Toggle Blockquote
3. Enter a line of text
4. Hit Enter

In this scenario we should probably remove the blockquote from the next bullet (We're currently carrying it over).

### Comment C: Spacing
Adding an image below, comparing Aztec and Calypso. Spacing should proooobably be horizontal (between the bullet and the blockquote).

image

SergioEstevao:
@lantean I agree with except for C all the others should be separated bugs.

A) is the result of list being added to the end of properties when indenting instead of following the last list

B) It's a general logic bug/issue we need to find how the other editor behave on these scenarios.

SergioEstevao:
@jleandroperez I improved the layout, give it another look please.

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -183,6 +183,39 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
...
+
+        return CGFloat(depth) * Metrics.listTextIndentation
+    }
+
+    /// The amount of indent for the list of the paragraph if any.
+    ///
+    public var listIndent: CGFloat {
+        let listAndBlockquotes = self.properties.filter({ (property) -> Bool in
+            return property is Blockquote || property is TextList
+        })
```
jleandroperez: Few notes!

- No need for `self`
- No need to wrap up the last closure within braces
- No need to define the closure's return type

```
let listAndBlockquotes = properties.filter { property in
    return property is Blockquote || property is TextList
}
```

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -183,6 +183,39 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
         }
     }
 
+    /// The amount of indent for the blockquote of the paragraph if any.
+    ///
+    public var blockquoteIndent: CGFloat {
+        let blockquoteIndex = self.properties.filter({ (property) -> Bool in
```
jleandroperez: - No need for `self`
- No need to define the return type

```
let blockquoteIndex = properties.filter({ property in
```

jleandroperez:
@SergioEstevao the new indentation feels a bit too much. Can we shrink that by 25% (or so) perhaps?

simulator screen shot jul 20 2017 10 10 55 am

Awesome changes BTW!!

SergioEstevao:
@jleandroperez addressed your comments, can you have another look.

PR #612, opened 2017-07-13
TextView: Nukes dead method

jleandroperez:
### Details:
(Just by chance) found a dead method. Keeping the house clean!

cc @diegoreymendez


jleandroperez:
Thanks!!

PR #618, opened 2017-07-14
Issue-568. fixing bug for underline link.

### LINKED ISSUE #568: Links are not underlined in the Editor
### Expected behavior
Links are underlined and easy to catch

### Actual behavior
Links are not underlined. Problem to  see a link, when a person is colorblind or the screen is harder to see (bright sunlight, low brightness of the screen, ...) 

Create a link in the editor

Iphone 5s

arisun:
Fixes #568 
introduced underline in link.
To test:
check for underline link in the editor.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -918,31 +918,12 @@ open class TextView: UITextView {
         let formatter = LinkFormatter()
         formatter.attributeValue = url
         let attributes = formatter.apply(to: typingAttributes)
-        let underLineStr = addUnderlineInLink(title: title, attributes: attributes)
 
-        storage.replaceCharacters(in: range, with: underLineStr)
+        storage.replaceCharacters(in: range, with: NSAttributedString(string: title, attributes: attributes))
 
         selectedRange = NSRange(location: finalRange.location + finalRange.length, length: 0)
 
         delegate?.textViewDidChange?(self)
-    }
-
-    /// Create underline link to the designated string with existing attributes.
-    ///
-    /// - Parameters:
-    ///     - title: the text for the link.
-    ///     - attributes: Existing atributes for formatting.
-    ///
-
-    open func addUnderlineInLink(title: String?, attributes:[String: Any]?) -> NSMutableAttributedString{
-
-        guard let textString = title, let attrbutes = attributes else{
-            return NSMutableAttributedString(string:"")
-        }
-        let attrStrMutable = NSMutableAttributedString(string: textString, attributes: [NSUnderlineStyleAttributeName: NSUnderlineStyle.styleSingle.rawValue])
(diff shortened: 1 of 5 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -909,23 +909,42 @@ open class TextView: UITextView {
...
 
         undoManager?.registerUndo(withTarget: self, handler: { [weak self] target in
             self?.undoTextReplacement(of: originalText, finalRange: finalRange)
         })
 
         let formatter = LinkFormatter()
-        formatter.attributeValue = url        
+        formatter.attributeValue = url
         let attributes = formatter.apply(to: typingAttributes)
+        let underLineStr = addUnderlineInLink(title: title, attributes: attributes)
```
SergioEstevao: @arisun instead of adding the underline attributes here you should do it inside the LinkFormatter class above.

The formatter classes are our central place to apply attributes and they are used in all place that the link attributes are needed. 

The way you implemented the underline attribute is only applied when the user create a link using the UI not when a link is read directly from the HTML.

arisun: got it. thanks for the guidance.

diegoreymendez: As a side note, the `LinkFormatter` class should also have additional code to be able to remove the underline from links too.

SergioEstevao:
@arisun I had a look at your changes while the new code is perfect for adding the new attributes. It introduces a problem we are not able to distinguish between a underline that is added because of explicit style and an underline added because of a link.

This causes issues with our toolbar style selection.

I investigated a bit further and found an alternative solution using the NSLayoutManager you can see it in action here: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/621

I think that approach will be better alternative and give us better control of link styling.

Thoughts?

cc @diegoreymendez

diegoreymendez:
@SergioEstevao - I don't think we should be drawing underlines ourselves.  I'd rather try to work this out using logic (since the drawing is already resolved by the underline style).

I've played a bit with Pages.app, and it seems like:

1. Links are underlined by default.
2. You can remove the underline from a link, so it's indeed a style (not sure this makes sense for our HTML backed links though.  In fact I think it doesn't.
3. Removing a word doesn't go back to what it was before adding the link.  It's based on the style right before the word (well sort of).

I'll review this PR to comment specifically on what I think needs to be done.  But I'd still like to go in the direction of using regular styles @SergioEstevao.

SergioEstevao:
@arisun I found even a simpler solution that solves all our issues and it's configurable for third parties.

There is a system property in UITextView that allows to configure the presentation of links.

Check the PR here: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/623

diegoreymendez:
I'm closing this PR as it seems the solution in #623 is the standard one in terms of `UITextView`.

Issue #417, opened 2017-04-12
Nuke Force Redraw Cursor Workaround


jleandroperez:
### Details:
This workaround was added as a way to fix #144. Let's investigate if the original issue can be solved in a more elegant way, and remove the `forceRedrawCursorAfterDelay` method, if possible.

PR #622, opened 2017-07-18
Fix: Missing Cursor after Inserting Blockquote

### LINKED ISSUE #440: Blockquote: Missing Cursor
### Steps:
1. Toggle Blockquote

As a result, the cursor will be invisible (since it'll be displayed below the blockquote's background).

jleandroperez:
Fixes #440

### To test:
1. Launch the empty editor
2. Insert a Blockquote
3. Verify that the Cursor is visible

cc @diegoreymendez 
Thanks in advance!

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -887,9 +887,13 @@ open class TextView: UITextView {
         let delay = 0.05
         DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
             let pristine = self.selectedRange
-            let updated = NSMakeRange(max(pristine.location - 1, 0), 0)
+            var location = max(pristine.location - 1, 0)
+            if pristine.location == location {
+                location = min(pristine.location + 1, self.storage.length)
+            }
+
             let beforeTypingAttributes = self.typingAttributes
-            self.selectedRange = updated
+            self.selectedRange = NSMakeRange(location, 0)
             self.selectedRange = pristine
             self.typingAttributes = beforeTypingAttributes
         }
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -883,6 +883,9 @@ open class TextView: UITextView {
     /// Force the SDK to Redraw the cursor, asynchronously, after a delay. This method was meant as a workaround
     /// for Issue #144: the Caret might end up redrawn below the Blockquote's custom background.
     ///
+    /// Workaround: By changing the selectedRange back and forth, we're forcing UITextView to effectively re-render
+    /// the caret.
+    ///
     func forceRedrawCursorAfterDelay() {
         let delay = 0.05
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -887,9 +887,13 @@ open class TextView: UITextView {
         let delay = 0.05
         DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
             let pristine = self.selectedRange
-            let updated = NSMakeRange(max(pristine.location - 1, 0), 0)
+            var location = max(pristine.location - 1, 0)
+            if pristine.location == location {
+                location = min(pristine.location + 1, self.storage.length)
```
diegoreymendez: Can we document this method, and anything that's not ovbious in it a bit more?

I'm realizing that reviewing this PR is difficult because it's not immediately obvious what the logic is.

jleandroperez:
@diegoreymendez comment added, ready for another look!

Aztec/Classes/TextKit/TextView.swift
diegoreymendez: The method documentation is appreciated, however the reasoning behind the logic in this method is still not easy to understand:

1. Why are we moving location to `-1`?
2. What's the logical meaning of `pristine.location == location`?  Can we store that in a boolean with a more meaningful name?  (for example: `selectedRangeStartsAtPositionZero`, which is what I believe we're checking for)
3. Why are we moving the location forward if `pristine.location == location`?  This is really not clear to me.

diegoreymendez: Also, `pristine` should be stored before the block is executed, otherwise you risk using an outdated selection range.

jleandroperez: Ahhhhh i see! Goal there is to move the **selectedRange.location** to +1 or -1, in order to force the cursor to redraw. Logic is as follows:

If we're at the end of the document... we can't move +1, or we'd trigger an Out of Bounds exception, in which case we'll proceed with -1.

(After seeing this logic a couple days later, i think it can be just simplified). Will be updating + documenting why, thanks Diego!

jleandroperez: Actually, nope. We're storing `pristine` during the block execution **on purpose**. Reasoning behind that is:

- There's a distant, very remote chance in which the selectedRange changes between the block scheduling and acutal execution
- If the selectedRange does change, though remote, you'll see the cursor jumping around

The goal of this hack (it's indeed a **hack**) is to force TextKit to redraw the cursor, while retaining the exact position it had at the moment of execution. Initial cursor position is irrelevant, in this particular scenario.

Thanks for the feedback Diego!

jleandroperez:
@diegoreymendez it should (hopefully) be crystal clear now. However, i'd like to note: i'm not happy with this hack, since day one. That's why we have #417 in place, perhaps there's a more suitable alternative.

Thanks Diego!!

jleandroperez:
Thanks Diego!!

PR #628, opened 2017-07-19
Copy + Paste: Stripping Attributes that don't support NSCoding

### LINKED ISSUE #627: Copy + Paste: Strip HTMLRepresentation Elements
Due to Crash #626, we'll be releasing Aztec Beta 8.1 with a hotfix: HTMLRepresentation will be stripped from the Pasteboard.

This will be later re-enabled with support for the new entities.

jleandroperez:
### Details:
In this brief PR we're disabling the Copy of attribute sthat can't be archived, as a temporary workaround for issue #627.

Please note that two required initializers have been added to VideoAttachment and ImageAttachment, respectively. This was done because it was the source of a secondary Copy-Y crash, triggered with the exact same steps.

Fixes #627

### To test:
1. Launch the Demo Editor with sample text
2. Select the entire document
3. Hit Copy
4. Go to the Empty Editor
5. Hit Paste

Aztec shouldn't crash in any step.


jleandroperez:
@diegoreymendez @SergioEstevao ready for another look

Sergio: Thanks so much for catching the **Cut** crash as well!!

jleandroperez:
Thanks a lot @SergioEstevao !!

PR #639, opened 2017-07-21
Only strip non-standard attributes.

### LINKED ISSUE #638: When copy pasting inside the editor all style attributes are lost
### Expected behavior

 When copy pasting inside the editor, I was expecting the attributes set on the editor to be kept

### Actual behaviour

All the style attributes (including font and paragraph style) are lost

On the demo app:

- Select a portion of text
- Tap Copy or Cut
- Paste on another section
- See that all style attributes are lost.

SergioEstevao:
Fixes #638

The current strip method is not working correctly in the sense that it strips all attributes and not only the specified in the kinds parameter.
I tried to fix that method but it looks there is no proper way to find if type is the type that belongs to the array.

This call `return self.isObject(value, kindOf: kind)`

for this method

```
private func isObject<T>(_ object: Any, kindOf type: T) -> Bool {
        return type(of: object) is T
    }
```

is testing that the value is of type Any.Type and because all classes are, they are all stripped.

In replacement I added a method that simple checks the name of the attributes is prefixed by `NS`

This method is a temporary fix, until we actually encode properly all our custom attribute classes.

To test:
 - On the demo app
 - Select a segment of text
 - Copy or Cut it
 - Paste across another part of the document
 - Check that styles and attachment are copied through

### DIFF
--- a/Aztec/Classes/Extensions/NSAttributedString+Strip.swift
+++ b/Aztec/Classes/Extensions/NSAttributedString+Strip.swift
@@ -33,4 +33,24 @@ extension NSAttributedString {
     private func isObject<T>(_ object: Any, kindOf type: T) -> Bool {
         return type(of: object) is T
     }
+
+    /// Removes attributes that not the standard attribute type of NSAttributedString
+    ///
+    func stripNonStandardAttributes() -> NSAttributedString {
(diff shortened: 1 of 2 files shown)


SergioEstevao:
Closing this in favour of #642

PR #644, opened 2017-07-24
Fix: Stop encoding non-latin characters

### LINKED ISSUE #632: Stop encoding non-latin characters in HTML Mode
### Description:

We're currently encoding many non-latin characters when switching to HTML mode.

### Details:

This issue was created by unifying #298 (Chinese) and #631 (Arabic).

This is an example provided by @elibud showing how Arabic characters are encoded when switching to HTML mode:

arabicandhtmlswitch

1. On XCode set the scheme to use pseudo RTL language
2. On the simulator add the Arabic keyboard
3. Create a new post using Aztec, with the Arabic keyboard type some text.
4. Switch to HTML

### Tested on simulator, iOS 10.3, Aztec 1.0.0-beta.8

jleandroperez:
Fixes #632

### To test:
1. Run the Unit Tests!
2. Verify that Japanese / Non Latin characters don't get encoded anymore

cc @SergioEstevao 
Thanks in advance sir!!

### DIFF
--- a/Aztec/Classes/Extensions/String+HTML.swift
+++ b/Aztec/Classes/Extensions/String+HTML.swift
@@ -19,11 +19,25 @@ extension String {
 
     /// Escapes the following HTML entities: [&, <, >, ', "]
     ///
-    private func escapeHtmlNamedEntities() -> String {
-        return replacingOccurrences(of: "&", with: "&amp;")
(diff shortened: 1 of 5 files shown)


jleandroperez:
Obrigado @SergioEstevao !!!

PR #646, opened 2017-07-24
Fixing Blockquote within Nested List Indentation

### LINKED ISSUE #633: Hitting Tab causes the bullet to indent, but the blockquote is not moving.
### Expected behavior

When hitting tab on a paragraph with a list and a block quote the list should indent together with the block quote

### Actual behavior

The list is indented one level down but the the block quote stays one level up

in terms of HTML we start with this:

`<ol><li><blockquote>Test<blockquote></li><ol>`

and we want to finish with this:

`<ol><li><ol><li><blockquote>Test<blockquote></li><ol></li><ol>`

instead we are getting this:

`<ol><li><blockquote><ol><li>Test</li><ol><blockquote></li><ol>`

### Steps to reproduce the behaviour

 - Toggle Ordered List
 - Toggle Blockquote
 - Hit Tab

As a result the Ordered List will get shifted rightwards, but the blockquote will remain in place.

jleandroperez:
### Details:
In this PR we're adding logic so that **Nested Text Lists** get grouped together. Instead of producing:

```
OL > LI > Blockquote > OL > LI
```

We should produce:

```
OL > LI > OL > LI > Blockquote
```

Fixes #633 

### To test:
1. Verify that the Unit Tests are happy
2. Launch the Empty Editor
2.1. Toggle Ordered List
2.2. Toggle Blockquote
2.3. Hit Tab
3. Verify that the produced HTML looks as expected.


jleandroperez:
@SergioEstevao implementing the behavior you've suggested in the original issue (because i kinda like it, it's exactly what i'd expect!).

Please, let me know what you think about it. Thanks in advance sir!!

cc @diegoreymendez

jleandroperez:
Thanks for the review sir!

Issue #578, opened 2017-06-29
Attributes in "li" nodes may be lost.


diegoreymendez:
### Description:

Attributes in `<li>` nodes are lost when exporting HTML.

### Details & background:

List details are currently stored in instances of class `TextList`, inside `ParagraphStyle.properties` (`ParagraphStyle` is a customizable replacement for standard `NSParagraphStyle` in `NSAttributedString`).

The problem we have is that, while we're currently saving the HTML representation of `<ul>` and `<ol>` nodes in the `TextList` instances, we're not doing the same for `<li>` nodes.

We need to figure out an adequate location and mechanism to store this information.

arisun:
I would like contribute on this project. Can I start with this issue ?

diegoreymendez:
Sure @arisun !

I'll add some details to the issue, though.

diegoreymendez:
Done.  Let me know if there's anything I can help you with.

diegoreymendez:
@arisun - This may not be the easiest first issue to tackle, so if you prefer me to give you some help picking one, just let me know.

arisun:
Thanks @diegoreymendez . It will be very helpful if you could provide some help selecting the  first one.

arisun:
Not able to compile the project. showing "Command failed due to signal: Segmentation fault: 11" on Aztec framework.

diegoreymendez:
@arisun - Could you join the WordPress.org Slack server?  (instructions here: https://make.wordpress.org/chat/)

The channel is #mobile and my handle is @diegoreymendez.

It'd make it so much easier to interact, and it would be good to avoid chat noise in this issue.  Thanks!

diegoreymendez:
As reported by @koke here:

When Aztec opens a gallery from Gutenberg and saves it's stripping the `class` attribute from each list item, breaking the gallery block.

It's turning this:

```
<!-- wp:gallery -->
<ul class="wp-block-gallery alignnone columns-1 is-cropped">
    <li class="blocks-gallery-item">
        <figure><img src="https://sandbox.koke.me/wp-content/uploads/2018/05/fullsizeoutput_52f7.jpeg" alt="" data-id="96" /></figure>
    </li>
</ul>
<!-- /wp:gallery -->
```

Into this:

```
<!-- wp:gallery -->
<ul class="wp-block-gallery alignnone columns-1 is-cropped">
  <li>
[...]

PR #666, opened 2017-07-31
Data Loss Fix

### LINKED ISSUE #658: Unsupported HTML: Data Loss
### Details:
We've found a sample case of UnsupportedHTML effectively breaking the document serialization. Input HTML and generated output below. 

- The div tag gets applied to the entire Attributed String's legnth
- The BR tag ends up in the same UnsupportedHTML container
- As a side effect, the string `Aztec, don't forgive me!` ends up being a child node of the `br` node
- Upon serialization, since **BR**'s cannot have children, the text won't be printed out. Also known as Data Loss.

### Input HTML (Pristine):

```
"<div><br>Aztec, don't forget me!</div>"
```

### Output HTML (Broken):

```
<div><br></div>
```

jleandroperez:
## Details:
In this PR we're addressing an issue in our AttributedString to Nodes converter, which resulted in BR entities in having TextNodes as children:

- UnsupportedHTML now contains an array of HTMLElementRepresentation. We were, previously, storing an ElementNode with an entire subtree, which was the source of multiple Tree Generation issues.
- Multiple UnitTests added
- And multiple UnitTests updated!

Fixes #658

### To test:
Run the unit tests and make sure everything is happy!

cc @diegoreymendez


Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -43,6 +42,31 @@ class NSAttributedStringToNodes: Converter {
...
+    /// present in the Enclosing Range. Otherwise, we might loose data!
+    ///
+    /// - Parameter:
+    ///     - attrString: Reference to the document to be converted.
+    ///     - paragraphRange: Defines the Paragraph Range to be converted into Nodes.
+    ///     - enclosingRange: Defines the Paragraph's Enclosing Range (containing newline characters).
+    ///
+    /// - Returns: Array of Node instances.
+    ///
+    private func createNodes(fromString attrString: NSAttributedString, paragraphRange: NSRange, enclosingRange: NSRange) -> [Node] {
```
diegoreymendez: Consider renaming `fromString attrString:` to `fromParagraph paragraph:` since it focuses more on the logical significance on the string, and less on its type.

AztecTests/Converters/HTMLNodeToNSAttributedStringTests.swift
```diff
@@ -24,32 +24,80 @@ class HTMLNodeToNSAttributedStringTests: XCTestCase {
...
         XCTAssertEqual(restoredSpanAttribute1?.value.toString(), "first")
     }
+
+
+    /// Verifies that BR elements contained within div tags do not cause Data Loss.
+    /// Ref. #658
+    ///
+    func testLineBreakTagWithinUnsupportedHTMLDoesNotCauseDataLoss() {
+        let inHtml = "<div><br>Aztec, don't forget me!</div>"
+        let expectedHtml = "<div><br></div><div>Aztec, don't forget me!</div>"
```
diegoreymendez: The best scenario would be `<div><br>Aztec, don't forget me!</div>`.
We could eventually accept `<div></div><div>Aztec, don't forget me!</div>`

But this is not a valid result as it duplicates the amount of newlines.

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -249,17 +249,20 @@ private extension HTMLNodeToNSAttributedString {
             return attributes
         }
 
-        let representation = HTMLRepresentation(for: .element(HTMLElementRepresentation(element)))
+        let elementRepresentation = HTMLElementRepresentation(element)
+        let representation = HTMLRepresentation(for: .element(elementRepresentation))
         var finalAttributes = attributes
 
+        let ignoredAttributes = [StandardElementType.li.rawValue]
```
diegoreymendez: Can we roll this change back please?

This should be tackled and discussed as part of #578, and doesn't really belong to the scope of this PR.

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -249,17 +249,20 @@ private extension HTMLNodeToNSAttributedString {
...
+
         if let elementFormatter = formatter(for: element) {
             finalAttributes = elementFormatter.apply(to: finalAttributes, andStore: representation)
-        } else  if element.name == StandardElementType.li.rawValue {
```
diegoreymendez: Can we roll this change back please? (including next 2 lines).

This should be tackled and discussed as part of #578, and doesn't really belong to the scope of this PR.

Aztec/Classes/Converters/HTMLNodeToNSAttributedString.swift
```diff
@@ -303,12 +306,17 @@ private extension HTMLNodeToNSAttributedString {
...
-        let unsupportedHTML = attributes[UnsupportedHTMLAttributeName] as? UnsupportedHTML ?? UnsupportedHTML()
-        unsupportedHTML.append(element: element)
-
+    private func attributes(storing representation: HTMLElementRepresentation, in attributes: [String: Any]) -> [String: Any] {
+        let unsupportedHTML = attributes[UnsupportedHTMLAttributeName] as? UnsupportedHTML
+        var representations = unsupportedHTML?.representations ?? []
+        representations.append(representation)
+
+        // Note:
+        // We'll *ALWAYS* store a copy of the UnsupportedHTML instance. Reason is: reusing the old instance
```
diegoreymendez: The location of these comments seems slightly off... what about moving it one line below so that it's attached to the line that creates the new instance?

jleandroperez: I've done it this way... because the way you've mentioned implies leaving the `var updated` floating above, which looks awful!

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
jleandroperez: Detail to it is: the received string does **not** represent a paragraph, but the **entire** string, and the nodes will be created using the specified ranges. What do you think?

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -395,14 +435,42 @@ private extension NSAttributedStringToNodes {
     ///
     /// - Returns: ElementNode representing the specified Paragraph.
     ///
-    func createParagraphNodes(from attrString: NSAttributedString) -> [ElementNode] {
-        guard let paragraphStyle = attrString.attribute(NSParagraphStyleAttributeName, at: 0, effectiveRange: nil) as? ParagraphStyle else {
+    func createParagraphNodes(fromParagraph attrString: NSAttributedString) -> [ElementNode] {
```
diegoreymendez: Maybe `attrString` -> `paragraph`?

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -705,7 +775,7 @@ private extension NSAttributedStringToNodes {
     ///
     /// - Returns: Leaf Nodes contained within the specified collection of attributes
     ///
-    func createLeafNodes(from attrString: NSAttributedString) -> [Node] {
+    func createLeafNodes(fromString attrString: NSAttributedString) -> [Node] {
```
diegoreymendez: Any reason why we did this name change?

Adding `fromString` would make sense if there was another method with a similar signature, but the variable is clearly an attributed string given its type, and there's no other method with the same signature.

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -553,7 +621,7 @@ private extension NSAttributedStringToNodes {
     ///
     /// - Returns: Style Nodes contained within the specified collection of attributes
     ///
-    func createStyleNodes(from attributes: [String: Any]) -> [ElementNode] {
+    func createStyleNodes(fromAttributes attributes: [String: Any]) -> [ElementNode] {
```
diegoreymendez: Can we roll this back please?

There's no chance of this method colliding with another one, and `from attributes` makes it clear what we're expecting here.

PS: If the problem is that `[String : Any]` not being clearly identified as "Attributes" I'd suggest adding to this class:

```typealias Attributes = [String : Any]```

So this method would become:

```
func createStyleNodes(from attributes: Attributes) -> [ElementNode] {
```

File: Aztec/Classes/Converters/NSAttributedStringToNodes.swift [hunk 9]
```diff
@@ -395,14 +435,42 @@ private extension NSAttributedStringToNodes {
...
+
+
+    /// Extracts the ElementNodes contained within a Paragraph's AttributedString.
+    ///
+    /// - Parameters:
+    ///     - attributes: Paragraph's Attributes from which we intend to extract the ElementNode
+    ///
+    /// - Returns: ElementNode representing the specified Paragraph.
+    ///
+    func createParagraphNodes(fromAttributes attributes: [String: Any]) -> [ElementNode] {
```
diegoreymendez: As commented below, the type should be pretty clear from the variable type, so there should not be a need to specify it in the parameter name.

Can we make this `from attributes`?

Aztec/Classes/Converters/NSAttributedStringToNodes.swift
```diff
@@ -59,18 +83,34 @@ class NSAttributedStringToNodes: Converter {
...
     }
+
+
+    /// Converts a collection of Attributes into their Node(s) representation.
+    ///
+    /// - Parameter attributes: Attributes to be mapped into nodes
+    ///
+    /// - Returns: Array of Node instances.
+    ///
+    private func createNodes(fromAttributes attributes: [String: Any]) -> [Node] {
```
diegoreymendez: Same as below.

jleandroperez:
Closing this PR, since we've split it's changes in multiple smaller PR's. Thanks @diegoreymendez!!

PR #671, opened 2017-08-01
UnsupportedHTML: Multiple Updates

jleandroperez:
## Details:
In this PR we're updating the UnsupportedHTML property:

- UnsupportedHTML is now immutable. Win: subranges are enforced to require a new copy, whenever a new UnsupportedHTML entity shows up.
- UnsupportedHTML now contains an array of HTMLElementRepresentation. We were, previously, storing an ElementNode with an entire subtree, which was the source of multiple Tree Generation issues.
- Multiple UnitTests added + fixed!

Ref #658

### To test:
Run the unit tests and make sure everything is happy!

cc @diegoreymendez


jleandroperez:
Thank you!!

PR #672, opened 2017-08-01
Fixing div partitioning

jleandroperez:
### Details:
The DIV element is considered non mergeable. As a result, Aztec currently has the following behavior:

Input: `<div><p>Hello</p><p>World</p></div>`
Output: `<div><p>Hello</p></div><div><p>World</p></div>`

In this PR we're marking the **div** element as mergeable, and thus, preventing such partitioning.

Ref #658

### To test:
Run the unit tests!

cc @diegoreymendez


jleandroperez:
Thanks Diego!!

PR #676, opened 2017-08-01
Media: Preventing Infinite Loop (Just the retry, we love Cupertino though)

### LINKED ISSUE #675: Media: Infinite Retry Loop
### Expected behavior
Upon N failed attempts, we shouldn't retry downloading a remote asset.

### Actual behavior
Aztec enters a loop in which download is retried an infinite number of times.

Enter the following HTML:

```
<img src="https://someinvalid.url/with-an-invalid-resource">
```

Toggle between rich and HTML mode, and hit back. Eventually, the simulator will crash.

jleandroperez:
### Details:
In this PR we're implementing a limit to the Image Download Retry Mechanism. After 3 failed attempts, we shall no longer retry (we were previously retrying an infinite number of times).

Fixes #675

### To test:
1. Add an invalid image (``
2. Verify the image disappears shortly after
3. Add a valid image (``
4. Verify the image remains onscreen

cc @diegoreymendez  @SergioEstevao 

P.s: YES Brent's blog was the first image that showed up.

Thanks in advance!

### DIFF
--- a/Aztec/Classes/TextKit/MediaAttachment.swift
+++ b/Aztec/Classes/TextKit/MediaAttachment.swift
@@ -49,7 +49,7 @@ open class MediaAttachment: NSTextAttachment {
     
     /// Attachment URL
     ///
-    open var url: URL? {
+    public var url: URL? {
         didSet {
             retryCount = 0
         }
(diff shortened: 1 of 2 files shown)


diegoreymendez:
@jleandroperez - Working as described, but for invalid images we should be showing a broken / missing image icon.

If we don't have that (can't remember right now) we should at least show the generic image icon we use as a placeholder.

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -49,9 +49,24 @@ open class MediaAttachment: NSTextAttachment {
     
     /// Attachment URL
     ///
-    open var url: URL?
+    open var url: URL? {
```
diegoreymendez: Do we want this to be overridden?

Allowing this value to be overriden will most likely break the retry count, as the setter will never be executed.

jleandroperez:
Thank you Diego!!

Issue #682, opened 2017-08-09
Media: Placeholder is Rendered after successfull Upload


jleandroperez:
### Details:
This issue is [WPiOS's #7648](https://github.com/wordpress-mobile/WordPress-iOS/issues/7648) sibling. Whenever the user uploads an image:

- The asset is immediately displayed
- As soon as the upload completes, the Media URL changes
- The placeholder is displayed
- And the newly uploaded asset gets downloaded.

### Objective(s):
Let's update our mechanism(s) so that:
- The image is never redownloaded after a successful upload.
- Alternatively: we do redownload the asset, but in this case, we don't display a new placeholder.

jleandroperez:
Closing this one since it was addressed via #686.

PR #686, opened 2017-08-11
MediaAttachment: Download Mechanism Cleanup

jleandroperez:
### Details:
In this PR we're updating the MediaAttachment download mechanism.

cc @diegoreymendez 
Ref. #682

### To test:
- Verify the unit tests are still happy
- Run a smoke test over the Sample App:
  - Try adding new assets (Video + Image).
  - Verify the overlay didn't break!

### DIFF
--- a/Aztec/Classes/TextKit/MediaAttachment.swift
+++ b/Aztec/Classes/TextKit/MediaAttachment.swift
@@ -27,7 +27,7 @@ open class MediaAttachment: NSTextAttachment {
 
     /// Attributes accessible by the user, for general purposes.
     ///
-    public var extraAttributes = [String: String]()
+    open var extraAttributes = [String: String]()
 
     /// Identifier used to match this attachment with a custom UIView subclass
     ///
@@ -80,14 +80,6 @@ open class MediaAttachment: NSTextAttachment {
         }
     }
 
-    /// Clears all overlay information that is applied to the attachment
-    ///
-    open func clearAllOverlays() {
-        progress = nil
-        message = nil
-        overlayImage = nil
-    }
-
     /// Image to be displayed: Contains the actual Asset + the overlays (if any), embedded
     ///
     internal var glyphImage: UIImage?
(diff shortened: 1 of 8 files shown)


Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -140,6 +136,17 @@ open class MediaAttachment: NSTextAttachment {
     }
 
 
+    // MARK: - Open Helpers
```
diegoreymendez: This category name is very generic and of no use.  Why not switch it to something more specific like "Overlay support"?

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -113,33 +101,53 @@ open class MediaAttachment: NSTextAttachment {
     required public init(identifier: String, url: URL? = nil) {
         self.identifier = identifier
         self.url = url
+        self.needsNewAsset = true
```
diegoreymendez: Init this at the declaration, since all initializers use the same value.

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -28,23 +28,23 @@ open class MediaAttachment: NSTextAttachment {
...
     private(set) open var identifier: String
     
     /// Attachment URL
     ///
-    public var url: URL? {
-        didSet {
-            retryCount = 0
-        }
-    }
+    fileprivate(set) open var url: URL?
```
diegoreymendez: Any specific scenario in which we think it may make sense to override this?  I'd consider removing the `open` modifier.

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -329,27 +346,30 @@ private extension MediaAttachment {
...
+            self.invalidateLayout(in: textContainer)
+
+        }, onFailure: { [weak self] _ in
+
+            guard let `self` = self else {
+                return
+            }
+
+            self.isFetchingImage = false
+            self.invalidateLayout(in: textContainer)
```
diegoreymendez: Since we're not changing the image, I'd consider removing this line.

Aztec/Classes/TextKit/MediaAttachment.swift
jleandroperez: Nope, you're actually right. Updating accordingly!

jleandroperez:
Wooooo thanks @diegoreymendez !!

PR #688, opened 2017-08-14
Links over images

### LINKED ISSUE #641: Impossible to link images to
### Expected behavior

We should be able to select an image and then tap on the link option and link it to any url

### Actual behavior

When we select the image and then tap on the link button, the image is removed.

This is cause because we are only reading plain text, and applying plain text when editing links.
This make any attachment to be lost.

 - Open the demo app
 - Scroll to an image
 - Select the image
 - Tap on the link button
 - Add an URL
 - Tap ok
 - See image be removed

SergioEstevao:
Fixes #641 

To test:
 - Start the demo app.
 - Select a range of text where an image or video exists.
 - Tap on the link button on the toolbar
 - Check that the link text option doesn't show
 - Change the url or write a new url
 - Tap on Insert Link
 - Switch to HTML mode and see if link is applied around the image/video

### DIFF
--- a/Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
+++ b/Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
@@ -49,7 +49,7 @@
             </BuildableReference>
          </TestableReference>
          <TestableReference
-            skipped = "YES">
+            skipped = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "CC400F161E9EC04200859AB4"
--- a/Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
+++ b/Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
@@ -49,7 +49,7 @@
             </BuildableReference>
          </TestableReference>
          <TestableReference
-            skipped = "NO">
+            skipped = "YES">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "CC400F161E9EC04200859AB4"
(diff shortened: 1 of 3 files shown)


diegoreymendez:
Even though this is working, I really think this PR needs to add an "Add link" / "Edit link" option whenever you tap on an image (depending on if the image has a link or not).

In its current state, it feels like editing links for images it not very easy.

Thoughts, @SergioEstevao?

Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
```diff
@@ -49,7 +49,7 @@
             </BuildableReference>
          </TestableReference>
          <TestableReference
-            skipped = "NO">
+            skipped = "YES">
```
diegoreymendez: Any idea why this changed or what it does?

diegoreymendez:
Even though this is working, I really think this PR needs to add an "Add link" / "Edit link" option whenever you tap on an image (depending on if the image has a link or not).

In its current state, it feels like editing links for images it not very easy.

Thoughts, @SergioEstevao?

Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
SergioEstevao: It was to disable the UITests.

SergioEstevao:
@diegoreymendez the initial idea of this fix was to make it possible to add links to sections of text with image without removing the images. 

That was the problem reported in #641 that I think this PR sorts.

As a side effect it also allows to add links to images, I agree with you that this is not the ideal solution for images but that will need to be address on a separate PR. For me the link option on images should be part of the edit image options.

diegoreymendez:
> For me the link option on images should be part of the edit image options.

Could be.

Example/AztecExample.xcodeproj/xcshareddata/xcschemes/AztecExample.xcscheme
diegoreymendez: Can you revert this change please?

The UI tests have been disabled in a separate PR, and this change doesn't really seem to prevent them from running anyway.

PR #689, opened 2017-08-14
Carthage Script: Nuking logfile

jleandroperez:
### Details:
This (brief) PR nukes the Carthage Logfile, which isn't really needed.

### To test:
- Nuke `Example/Carthage/*`
- Run the sample app from within Xcode
- Verify that the script still runs!

cc @diegoreymendez


jleandroperez:
Thanks Diego!

PR #694, opened 2017-08-16
Fixes several memory leaks and retain cycles in Aztec.

diegoreymendez:
Fixes several memory leaks and retain cycles in Aztec.

Please feel free to discuss the fixes as much as necessary.

**How to test:**

1. Run `develop` with the Leaks instrument.
2. Type into a post a few words, using blockquotes, lists, styles, etc.
3. Switch to HTML mode and back repeatedly.
4. Notice the big amount of leaks and retain cycles reported.
5. Repeat the same process in my branch and notice many of them are gone.


diegoreymendez:
Merged with an Admin override after 23 minutes of waiting for Travis to finish.

jleandroperez:
Nice touches!

PR #696, opened 2017-08-18
Update FormatBar to use Strings as identifiers

frosty:
This PR is part of the format bar updates required for https://github.com/wordpress-mobile/WordPress-iOS/issues/7703.

In particular, it updates the format bar to use plain `String`s for bar item identifiers instead of `FormattingIdentifier`s specifically. This allows us to support a wider range of options for format bar
buttons – the buttons required for media insertion don't have associated formatting identifiers, so instead we can just provide strings.

The format bar itself never really made use of the fact that the identifiers were `FormattingIdentifier`s, only the code outside of the bar did. We can still use `FormattingIdentifier`s to manage the formatting items, but the bar doesn't need to care that we're doing so.

I'll have a PR coming up for WPiOS so that it'll work correctly with this update.

To test:

* Run the demo app and check the bar still works as intended. In particular, check that formatting of text works, and that the items become enabled correctly to show the format of the currently selected text.

Needs review: @SergioEstevao

### DIFF
--- a/Aztec/Classes/GUI/FormatBar/FormatBarDelegate.swift
+++ b/Aztec/Classes/GUI/FormatBar/FormatBarDelegate.swift
@@ -7,10 +7,10 @@ public enum FormatBarOverflowState {
 }
 
 public protocol FormatBarDelegate : NSObjectProtocol {
-    /// Prompts the delegate that the bar item with the specified identifier was tapped,
+    /// Prompts the delegate that the specified bar item was tapped,
     /// and it should take appropriate action.
     ///
-    func handleActionForIdentifier(_ identifier: String, barItem: FormatBarItem)
+    func handleAction(for barItem: FormatBarItem)
 
     /// Informs the delegate that a touch down event was received on the format bar.
     ///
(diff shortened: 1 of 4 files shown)


diegoreymendez:
Adding myself as reviewer too.

Aztec/Classes/GUI/FormatBar/FormatBarDelegate.swift
```diff
@@ -10,7 +10,7 @@ public protocol FormatBarDelegate : NSObjectProtocol {
     /// Prompts the delegate that the bar item with the specified identifier was tapped,
     /// and it should take appropriate action.
     ///
-    func handleActionForIdentifier(_ identifier: FormattingIdentifier, barItem: FormatBarItem)
+    func handleActionForIdentifier(_ identifier: String, barItem: FormatBarItem)
```
diegoreymendez: Even though this wasn't introduced by this PR specifically, it could be a good opportunity to do some cleanup.

If the `FormatBarItem` instance contains its own identifier, there's no need to specify it as an additional parameter.

diegoreymendez: PS: also the name could use some Swift 3 love... something like this:

```
handleAction(for barItem: FormatBarItem)
```

frosty:
@diegoreymendez Addressed your comments! Thanks!

diegoreymendez:
@frosty - I just opened #698 in this repo for reference and informational purposes.

This PR fixes it.

Issue #699, opened 2017-08-18
FormatBar: Allow customization of leading and trailing items


frosty:
In order to add an 'insert' button to the right end of the format bar for media insertion in WPiOS, we should add the ability to set a custom 'trailing' button in the bar. While we're at it, we'll improve the way the media `+` button is added, and change that to a custom 'leading' button.

Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/7703

PR #700, opened 2017-08-21
FormatBar: Customise left and right items

frosty:
Implements #699

Our designs for the new media picker toolbar in WordPress for iOS require us to set a custom button on the trailing end of the format bar:

This PR adds a new `trailingItem` property that can be used to set a custom item on the right end of the bar. It also changes the way the leftmost item is set to use a similar `leadingItem` property instead of a subarray of bar items.

To test:

* Ensure that the toolbar behaves as it did before.
* In the demo app, you can add a custom trailing button in the `createToolbar` method of `EditorDemoController.swift`, by adding the following code just before the `return`:

```
        let trailingItem = UIButton(type: .custom)
        trailingItem.setTitle("Testing", for: .normal)
        trailingItem.setTitleColor(.black, for: .normal)
        trailingItem.titleLabel?.font = UIFont.systemFont(ofSize: 13.0, weight: UIFontWeightMedium)
        toolbar.trailingItem = trailingItem

        toolbar.trailingItemHandler = { item in
            print("Trailing item tapped")
        }
```
* Check that the trailing item looks correct and sticks to the right edge of the toolbar. It should look like this:

screen shot 2017-08-21 at 12 05 02

Needs review: @SergioEstevao

### DIFF
--- a/Aztec/Classes/GUI/FormatBar/FormatBar.swift
+++ b/Aztec/Classes/GUI/FormatBar/FormatBar.swift
@@ -26,15 +26,32 @@ open class FormatBar: UIView {
     fileprivate let topDivider = UIView()
     fileprivate let bottomDivider = UIView()
 
+    /// A leading item that will displayed in front of any default items.
+    /// A divider will be drawn between the leading item and any default items.
+    /// If set to a FormatBarItem, the appearance will match the default items in the bar.
+    ///
+    open var leadingItem: UIButton? = nil {
+        didSet {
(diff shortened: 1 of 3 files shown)


frosty:
The related WPiOS branch is available here, if you want to see it in use: https://github.com/wordpress-mobile/WordPress-iOS/tree/feature/new-media-picker-toolbar

frosty:
Thanks @SergioEstevao!

PR #706, opened 2017-08-23
HeaderFormatter: Restoring the default font size upon removal

### LINKED ISSUE #677: Default Font Restoration: Invalid Size
### Expected behavior
After entering a H1 snippet and hitting newline, the default font's size should get restored.

### Actual behavior
A smaller font size is restored.

1. Launch the empty editor
2. Toggle H1
3. Enter `Hello` + hit newline
4. Enter `World`

As a result, the font size used in `World` will be smaller than expected. Switch to HTML mode, and back to Rich Mode: note that the `World` word will be slightly bigger.

jleandroperez:
### Details:
In this PR we're updating the HeaderFormatter so that the default Font Size is preserved whenever a Header Style is applied / updated, and restored whenever the style gets effectively removed.

Fixes #677

### To test:
1. Verify the unit tests are happy
2. Launch the empty editor
3. Type `Pristine` + newline
4. Toggle H1
5. Type `Hello` + newline + `World`
6. Verify that the `World` word gets rendered in the exact same size as `Pristine`

cc @diegoreymendez  + @SergioEstevao 
Thanks in advance!

### DIFF
--- a/Aztec.xcodeproj/project.pbxproj
+++ b/Aztec.xcodeproj/project.pbxproj
@@ -60,6 +60,7 @@
 		B5B96DAB1E01B2F300791315 /* UIPasteboard+Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5B96DAA1E01B2F300791315 /* UIPasteboard+Helpers.swift */; };
 		B5BC4FEE1DA2C17800614582 /* NSAttributedString+Lists.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5BC4FED1DA2C17800614582 /* NSAttributedString+Lists.swift */; };
 		B5BC4FF21DA2D17000614582 /* NSAttributedStringListsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5BC4FF11DA2D17000614582 /* NSAttributedStringListsTests.swift */; };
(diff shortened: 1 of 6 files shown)


SergioEstevao:
@jleandroperez while this sort the issue you refer I think this is part of an bigger issue.

At the moment if you have two paragraphs like this:
```
<h1>Header 1 </h1>
<h2>Header 2 </h2>
```
I was expecting that when I back delete just before Header 2 the text font on Header 2 should switch to Header 1 font size. We actually get a single paragraph but with both fonts size applied.... Funny enough if you switch to HTML you see a single <h1> so the resulting HTML is correct but the font sizes no.

Do you want to try to fix it here? or do a separate PR?

diegoreymendez:
> ...so the resulting HTML is correct but the font sizes no.

I think that should be H2 after backspacing, though.

SergioEstevao:
@diegoreymendez I'm expecting to be `h1` and I think it's correct just tested it in Calypso and it does convert the second line to be the same style has the first.

jleandroperez:
@diegoreymendez @SergioEstevao gentlemen, i've [filed a new issue here](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/708) to deal with the behavior that Sergio spotted.

I'd vote we deal with both as stand alone issues, since the logic to address each one of them should differ. Thank you!

jleandroperez:
Thanks everyone!

PR #716, opened 2017-09-05
String+EndOfLine: Renaming Helper

jleandroperez:
### Details:
Our **isEndOfParagraph** helper, actually, should be named **isEndOfLine**, since it checks if a given position is the beginning AND end of a line (condition which doesn't imply End of Paragraph!).

### To test:
Verify that the unit tests run correctly.

cc @diegoreymendez 
Thanks!


jleandroperez:
Thanks Diego!

PR #717, opened 2017-09-06
Not Loosing Styles!

### LINKED ISSUE #715: Nested Blockquote: Loosing Style
### Expected behavior
Blockquote Style is not lost after hitting backspace

### Actual behavior
Blockquote Style gets lost

1. Launch the Empty Editor
2. Insert a List
3. Insert a Blockquote in the first bullet
4. Type `Hello` **Shift + Enter** `There`
5. Hit backspace 6 times

As a result, the caret should be after the `Hello` word, in the first line. Instead: the Blockquote Style gets removed.

iPhone 7. iOS 10.3 / Aztec develop (09.05.2017)

jleandroperez:
### Details:
In this PR we're adding several Paragraph Analysis helpers, and switching the mechanism that removes Styles, from relying on `Lines` to `Paragraphs` instead.

Fixes #715
Needs Review: @diegoreymendez 

Thanks in advance!

### To test:
1. Launch the Empty Editor
2. Insert a List
3. Insert a Blockquote in the first bullet
4. Type Hello Shift + Enter There
5. Hit backspace 6 times

Verify that the Blockquote Style does not get removed from the `Hello` word.

Please note that after hitting backspace 6 times, the caret will be right by the `Hello` word, but it won't be visible. To be addressed in another PR!


jleandroperez:
Merging after waiting for way over an hour, for Travis to wrap up it's build. 

Thanks Diego!

PR #719, opened 2017-09-07
Base modifications for CSS Support

diegoreymendez:
### Description:

Several changes to support CSS styling.  Brings #285 forward.

### Details:

This PR moves us towards CSS support by:

- Adding class `CSSParser`.
- Integrating the CSS parsing step as part of the HTML parsing.
- Adding support for copying and pasting CSS attributes.

Also not directly related:

- Renames `CSSProperty` to `CSSAttribute`, for a more precise naming.
- Reorganizes the element -> formatters map, so that the same formatters can be reused for attributes if necessary.
- Introduces the `Coding` protocol and the `NSCodingProxy` class so we can implement `NSCoding` in `enum`s and `struct`s.  This will bring our code closer to the new Coding features in Swift 4.

### How to test:

**Test 1:**

1. Write some HTML with attributes (and CSS attributes).
2. Switch to visual mode.
3. Copy the text.
4. Paste it.
5 Switch to HTML mode and make sure no data was lost.

**Test 2:**

1. Open the sample editor with content.
2. Copy and paste all of the content.
3. Make sure HTML attributes are not lost in the process.
4. Make sure the app doesn't crash (of course compare against develop).


diegoreymendez:
@jleandroperez - Awesome catch.  This is ready for another round.

Issue #725, opened 2017-09-20
iOS 11 losing styles after typing 1 character


diegoreymendez:
When using Aztec in iOS 11, styles are being lost after typing just 1 character.

How to test:

1. Open the empty editor.
2. Type a regular first line.
3. Press enter
4. Enable any style.
5. Start typing.

Expected behavior: the full line should have the selected style.
Current behavior: after the first character is entered, the style is removed.

diegoreymendez:
Reported as a `UITextView` issue to Apple (Apple ref 34546954).

I've also created [an example minimal project](https://cloudup.com/c8rCpGokxGS) demonstrating this issue is not caused by our code.

jleandroperez:
Closed via #726

PR #726, opened 2017-09-20
Issue/725 ios11 losing styles

diegoreymendez:
Workaround for #725 

This issue seems to be cause by a UIKit bug introduced in iOS 11.  Read the original issue for more information.

To test:
1. Open the empty editor.
2. Type "asd" on the first line and press enter.
3. Enable bold, start typing.

Make sure the style is not lost.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -396,6 +396,8 @@ open class TextView: UITextView {
         // to `super.insertText(text)`.  Our workaround is to simply save the typing attributes
         // and restore them after that call.
         //
+        // Issue: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/725
+        //
         // Diego: I reproduced this issue in a very simple project (completely unrelated to Aztec)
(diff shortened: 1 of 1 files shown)


SergioEstevao:
While this solves the initial issue, it introduces another problem.

With this fix, when you have a line that has bold (or another character style) and tap Enter, the bold continues to the next line/paragraph. The characters style should be reseted when a new line starts.

jleandroperez:
Fix confirmed! @SergioEstevao the deleted code is actually a snippet that was duplicated (see a couple lines above the removed block!).

Bold is not getting reset after pressing Enter, but that also happens in develop, so it's not related.

Let's :shipit: this fellow!

diegoreymendez:
PS: as an additional note, I don't consider the bold style reset as an issue.  Even if it differs a bit from Calypso, it's the standard behavior that iOS provides (and quite honestly what I'd expect to happen).

If we get reports on it we can consider changing the behavior.

catehstn:
Agree with @diegoreymendez - I noticed it the other day, but realized it was the behavior I expect and that WP has just trained me otherwise.

Issue #728, opened 2017-09-25
Layout issues on iPhoneX


SergioEstevao:
### Expected behaviour

The editor should show content correctly on iPhone X and the text toolbar should take account the new margins.

### Actual behaviour

Some of the editor content is overlapped by the notch on landscape mode.

### Steps to reproduce the behavior

- Start the demo app on iPhoneX
- Open the editor with content
- See something like this:

<img width="964" alt="screen shot 2017-09-25 at 12 52 57" src="https://user-images.githubusercontent.com/651601/30807422-a638b454-a1f1-11e7-8fa7-4c426e912cfb.png">

jleandroperez:
Closed via #732 by @SergioEstevao.

Tested in develop / iOS 11 / iPhone X. Looking smooth!

PR #732, opened 2017-09-25
Fix margins on editor.

SergioEstevao:
Refs #728 

This PR just fixes the layout issues for the editor (html and visual) and title views work on iPhoneX, it not fixes the toolbar issues that will be addressed on a separate PR

To test:
 - Start the demo app
 - open the demo editor
 - check the layout of the several text components.
 - test on iPhoneX and other devices on portrait and landscape.

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -337,8 +337,8 @@ class EditorDemoController: UIViewController {
     fileprivate func refreshInsets(forKeyboardFrame keyboardFrame: CGRect) {
         let referenceView: UIScrollView = editingMode == .richText ? richTextView : htmlTextView
 
-        let scrollInsets = UIEdgeInsets(top: referenceView.scrollIndicatorInsets.top, left: 0, bottom: view.frame.maxY - keyboardFrame.minY, right: 0)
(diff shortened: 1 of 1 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -207,9 +207,9 @@ class EditorDemoController: UIViewController {
     
     func updateTitleHeight() {
         let referenceView: UIScrollView = editingMode == .richText ? richTextView : htmlTextView
-
-        let sizeThatShouldFitTheContent = titleTextField.sizeThatFits(CGSize(width:view.frame.width - ( 2 * Constants.margin), height: CGFloat.greatestFiniteMagnitude))
+        let layoutInsets = view.layoutMargins
```
frosty: Nitpick but could we call this `layoutMargins` instead of `layoutInsets`? Even though they are insets...

SergioEstevao:
@frosty ready for another look.

SergioEstevao:
@frosty ready for another look.

frosty:
Weird, rebuilt and it seems okay now. I'm sure I had latest, but hey... let's :shipit:

PR #736, opened 2017-09-27
Fixing Font upon text deletion

### LINKED ISSUE #708: HeaderFormatter: Back Delete should merge Header Styles
1. Enter two paragraphs (h1 + h2):
  ```
  <h1>Header 1 </h1>
  <h2>Header 2 </h2>
  ```
2. Back delete just before `Header 2`

As a result, the text font on Header 2 should switch to Header 1 font size. Aztec currently ends up rendering a single paragraph but with both fonts size applied.

Props to @SergioEstevao for spotting this!

jleandroperez:
### Details:
This PR ensures Font Consistency whenever
Fixes #708

### To test:
1. Launch Aztec
2. Toggle H1
3. Enter a line of text
4. Toggle H2
5. Enter another line of text
6. Place the cursor at the beginning of the second line
7. Hit backspace

As a result, you should have a single line of text. **Font size** should be the same for the entire string.
Verify that the HTML generated looks as expected + verify that the unit tests are green!

cc @diegoreymendez

### DIFF
--- a/Aztec/Classes/TextKit/TextStorage.swift
+++ b/Aztec/Classes/TextKit/TextStorage.swift
@@ -260,7 +260,7 @@ open class TextStorage: NSTextStorage {
     override open func setAttributes(_ attrs: [String: Any]?, range: NSRange) {
         beginEditing()
 
-        let fixedAttributes = ensureFontAttributeIsValid(beforeApplying: attrs ?? [:], at: range)
+        let fixedAttributes = ensureMatchingFontAndParagraphHeaderStyles(beforeApplying: attrs ?? [:], at: range)
 
         textStore.setAttributes(fixedAttributes, range: range)
         edited(.editedAttributes, range: range, changeInLength: 0)
@@ -370,7 +370,7 @@ open class TextStorage: NSTextStorage {
 //
 private extension TextStorage {
 
-    /// Ensures that the TextFont matches the new HeaderLevel that's about to be applied.
+    /// Ensures the font style is consistent with the paragraph header style that's about to be applied.
     ///
     /// - Parameters:
     ///   - attrs: NSAttributedString attributes that are about to be applied.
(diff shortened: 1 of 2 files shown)


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -363,6 +366,48 @@ open class TextStorage: NSTextStorage {
 }
 
 
+// MARK: - Header Font Attribute Fixes
+//
+private extension TextStorage {
+
+    /// Ensures that the TextFont matches the new HeaderLevel that's about to be applied.
```
diegoreymendez: A header style may not be applied at all, so the description of this method is a bit off.

Something along the lines of `Ensures the font style is consistent with the paragraph header style` would be closer to the truth.

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -363,6 +366,48 @@ open class TextStorage: NSTextStorage {
...
+    ///
+    /// - Parameters:
+    ///   - attrs: NSAttributedString attributes that are about to be applied.
+    ///   - range: Range that's about to be affected by the new Attributes collection.
+    ///
+    /// - Returns: Collection of attributes with the Font Attribute corrected, if needed.
+    ///
+    func ensureFontAttributeIsValid(beforeApplying attrs: [String: Any], at range: NSRange) -> [String: Any] {
```
diegoreymendez: In line with my previous comment, maybe rename this to something `ensureMatchingFontAndParagraphHeaderStyles()`?

The problem with `ensureFontAttributeIsValid` is that it doesn't really describe what's happening.

jleandroperez:
Thanks Diego!

jleandroperez:
Merging after ~30' of waiting for the CI to wrap up.

PR #751, opened 2017-10-05
Fix/iphone x toolbar

frosty:
Refs #728 and https://github.com/wordpress-mobile/WordPress-iOS/issues/7821

This PR adjusts the FormatBar so it better handles the iPhone X's screen and safe insets. In particular:

* The edges of the toolbar's buttons' container is now constrained to the safe area guides on either side, so in landscape we don't flow underneath the notch.
* The bar now adapts its height if the software keyboard is hidden (due to a hardware keyboard being connected) so it no longer sits underneath the home indicator.

iphone-x-format-bar-before-after

To test:

* Build and run the sample app on iPhone X.
* In the demo editor, check the toolbar's appearance in the following scenarios:
   * Portrait, hardware keyboard connected (you can use Cmd+K to toggle it) so toolbar sits at bottom of screen
   * Landscape, hardware keyboard connected
   * Landscape, hardware keyboard disconnected

Needs review: @SergioEstevao

### DIFF
--- a/Aztec/Classes/GUI/FormatBar/FormatBar.swift
+++ b/Aztec/Classes/GUI/FormatBar/FormatBar.swift
@@ -156,7 +156,7 @@ open class FormatBar: UIView {
         let availableWidth = visibleWidth
         guard availableWidth > 0 else { return [] }
 
-        let visibleItemCount = Int(floor(availableWidth / Constants.stackButtonWidth))
+        let visibleItemCount = Int(floor(availableWidth / Constants.defaultButtonWidth))
 
         let allItems = items
         let leadingItemCount = leadingItem != nil ? 1 : 0
@@ -177,7 +177,7 @@ open class FormatBar: UIView {
             trailingItem.sizeToFit()
             return trailingItem.bounds.size.width + Constants.trailingButtonMargin
         } else {
-            return Constants.stackButtonWidth
+            return Constants.defaultButtonWidth
         }
     }
 
@@ -328,7 +328,7 @@ open class FormatBar: UIView {
     open override func safeAreaInsetsDidChange() {
         super.safeAreaInsetsDidChange()
 
-        let newHeight = Constants.stackButtonWidth + safeAreaInsets.bottom
+        let newHeight = Constants.defaultBarHeight + safeAreaInsets.bottom
         if newHeight != heightConstraint?.constant {
             heightConstraint?.constant = newHeight
             layoutIfNeeded()
@@ -347,7 +347,7 @@ open class FormatBar: UIView {
     }
 
     open override var intrinsicContentSize: CGSize {
-        var height = Constants.formatBarHeight
+        var height = Constants.defaultBarHeight
         if let heightConstraint = self.heightConstraint {
             height = heightConstraint.constant
         }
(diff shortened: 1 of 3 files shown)


Aztec/Classes/GUI/FormatBar/FormatBar.swift
```diff
@@ -576,6 +630,16 @@ private extension FormatBar {
     /// Sets up the Constraints
     ///
     func configureConstraints() {
+        var leadingAnchor = self.leadingAnchor
+        var trailingAnchor = self.trailingAnchor
+        var bottomAnchor = self.bottomAnchor
+
+        if #available(iOS 11.0, *) {
+            leadingAnchor = safeAreaLayoutGuide.leadingAnchor
```
SergioEstevao: ins't the safeAreaLayoutGuide the same as layoutMarginsGuide on iOS11? any difference on the value?

Aztec/Classes/GUI/FormatBar/FormatBarItem.swift
```diff
@@ -101,13 +101,14 @@ open class FormatBarItem: UIButton {
     // MARK: - Lifecycle
 
     public convenience init(image: UIImage, identifier: String? = nil) {
-        let defaultFrame = CGRect(x: 0, y: 0, width: 44, height: 44)
+        let defaultFrame = CGRect(x: 0, y: 0, width: FormatBar.Constants.defaultButtonWidth, height: FormatBar.Constants.defaultButtonHeight)
```
SergioEstevao: Loving that the magical numbers went away :).

SergioEstevao:
There is still a scenario that is not working correctly. On a IPhone X simulator do the following:

- Start the demo app with no hardware keyboard connected
- Enter one of the sample editor
- Tap on the editor
- See the home indicator overlapping the toolbar.

Aztec/Classes/GUI/FormatBar/FormatBar.swift
frosty: As far as I'm aware, the layout margins guide is the same as the safe area + margins.

frosty:
@SergioEstevao I'm not sure I understand the situation you're highlighting. Following your steps I get this:

simulator screen shot - iphone x - 2017-10-05 at 21 17 19

Aztec/Classes/GUI/FormatBar/FormatBar.swift
SergioEstevao: Ok so the layoutMargins has extra spacing.

frosty:
@SergioEstevao Could you re-test please? I just pushed a fix that should ensure the bar is in the correct state when it first appears.

Aztec/Classes/GUI/FormatBar/FormatBar.swift
```diff
@@ -716,10 +794,18 @@ extension FormatBar: UIScrollViewDelegate {
     }
 }
 
+// MARK: - Constants
+extension FormatBar {
+    struct Constants {
```
diegoreymendez: Just a single comment because I've seen we're aiming towards this in several PRs.

Any chance we can opt for more meaningful / descriptive names for more atomic grouping?  And maybe even drop the word "Constants" entirely from it?

Having it named `Constants` will mean it'll end up holding a lot of unrelated values, while we could be using the naming to group them logically.

As an example, instead of:

```swift
Constants.defaultBarHeight
```

Make it:

```swift
Dimensions.defaultHeight
```

frosty: I see – that makes sense (some of these are already split out into `Animations`). Do you want me to do that in this PR?

diegoreymendez: I don't want to step into the review and force a change, so feel free to move ahead.

I just happened to see the `Constants` grouping in a few PRs and wanted to chime in since I feel like this can be a missed opportunity for us.

frosty:
@SergioEstevao As discussed in Slack, I can't see an easy way to fix that remaining issue at the moment, and I've already spent a while on this. Providing this last commit passes CI, let's merge what we have for now and we can always improve on it later.

Thanks for the review!

PR #759, opened 2017-10-09
Keyboard Switch: Fixing Vertical Padding Glitch

### LINKED ISSUE #756: Keyboard Switch: Vertical Padding Glitch
1. Emtpy Document
2. Type enough text to cause vertical scrolling
3. Place the cursor at the bottom of the document
4. Press `H` / `List`'s format bar buttons

As a result, the caret will flicker for a split second. This is (probably) due to a change in the Keyboard Notifications. Let's analyze if this is caused by our handlers updating the constraints, or it's something else!

jleandroperez:
### Details:
This PR prevents an undesired vertical scrolling glitch, when switching to a custom Input View (ie. Header Level picker).

Fixes #756 
cc @diegoreymendez 

### To test:
1. Emtpy Document
2. Type enough text to cause vertical scrolling
3. Place the cursor at the bottom of the document
4. Press `H` / `List`'s format bar buttons

Verify that the vertical offset doesn't flicker at all.

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -709,9 +709,9 @@ extension EditorDemoController {
         if richTextView.inputView == to {
             return
         }
-        richTextView.resignFirstResponder()
+
         richTextView.inputView = to
-        richTextView.becomeFirstResponder()
+        richTextView.reloadInputViews()
     }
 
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thank you Diego!

PR #763, opened 2017-10-12
Send UITextViewTextDidChange notification after text changes(e.g. paste)

azone:
Fixes #

Can not receive UITextViewTextDidChangeNotificiation for some operation

To test:

Copy any text or web content, paste into TextView

### DIFF
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -6,19 +6,6 @@ import QuartzCore
 // MARK: - Aztec Layout Manager
 //
 class LayoutManager: NSLayoutManager {
-
-    /// Blockquote's Left Border Color
-    ///
-    var blockquoteBorderColor = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
-
-    /// Blockquote's Background Color
-    ///
-    var blockquoteBackgroundColor = UIColor(red: 0.91, green: 0.94, blue: 0.95, alpha: 1.0)
-
-    /// HTML Pre Background Color
-    ///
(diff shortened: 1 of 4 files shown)


Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -6,19 +6,6 @@ import QuartzCore
...
-    var blockquoteBorderColor = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
-
-    /// Blockquote's Background Color
-    ///
-    var blockquoteBackgroundColor = UIColor(red: 0.91, green: 0.94, blue: 0.95, alpha: 1.0)
-
-    /// HTML Pre Background Color
-    ///
-    var preBackgroundColor = UIColor(red: 243.0/255.0, green: 246.0/255.0, blue: 248.0/255.0, alpha: 1.0)
-
```
jleandroperez: Mind reverting this one sir? (commit `13ed3719b02397821e2186e6ff85bf7e9b03ced6`). Reason is that by having these colors as instance variables, App Devs may run multiple Aztec instances, each one with different Blockquote / Pre Colors.

By moving this to static vars, we loose the ability to tailor down specific instances.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -381,6 +381,11 @@ open class TextView: UITextView {
         pasteboard.items[0][NSAttributedString.pastesboardUTI] = data
     }
 
+    fileprivate func textViewDidChange() {
```
jleandroperez: Looks good! but perhaps a name like `notifyTextViewDidChange` would work better? (that way it's crystal clear what the method does)

PR #766, opened 2017-10-13
TextView: Attachments Recognizer no longer receiving all touches

### LINKED ISSUE #765: Magnifier Issues
Moving WordPress iOS [Issue #7931](https://github.com/wordpress-mobile/WordPress-iOS/issues/7931) over, since it's an Aztec glitch.

--

As reported by a user:

> in the iOS where there is editable text you should be able to hold a finger on the text and the magnification loupe appears just above your finger. You then drag so the insertion point moves between letters. However holding on text does nothing until you remove the finger from the screen, the magnification loupe appears for a split second after the finger is removed and the black horizontal menu options appear.

Additional info by our internal user:

> I tested this on my iPhone and was able to replicate what they are describing. Steps to reproduce:
> 1. While editing a post tap and hold onto some text
> 2. After a few second remove your finger from the screen
> 3. For a quick moment the magnifier will appear, and then disappear being replaced by the menu bar.
>
> I switched to the old editor and the magnifier does work there. So the problem seems isolated to Aztec.
>
> iOS – 11.0.2
> WordPress – 8.4
> editor – Visual 2.0

jleandroperez:
### Details:
This PR updates the Attachments Gesture Recognizer, so that touches that fall beyond an image attachment aren't even accepted.

Fixes #765

### To test:
1. Launch the empty editor
2. Insert an image
3. Verify that you're still able to press over the attachment, and access it's options
4. Enter some text
5. Verify that iOS's Magnifier shows up after a long press

Needs Review: @SergioEstevao 
Thanks in advance sir!

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -1634,9 +1634,17 @@ extension TextView: TextStorageAttachmentsDelegate {
         return true
     }
 
-    func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
+        guard let textView = textView else {
+            return false
+        }
 
-        guard let textView = self.textView else {
+        let locationInTextView = touch.location(in: textView)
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thanks for the review @SergioEstevao, **MUCH** appreciated!!

PR #772, opened 2017-10-16
Updating Attachments Delegate's API

jleandroperez:
### Details:
In this PR we're allowing for `nil` MediaAttachment(s) URL. This will allow us to remove the `placeholder://` workaround that has served us well, but is behind a [WordPress iOS Issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/7973) in which Drag + Drop images won't immediately be displayed.

### To test:
1. Try adding new images
2. Verify nothing breaks
3. Verify that the unit tests still pass

cc @SergioEstevao 
Thanks in advance sir!


jleandroperez:
Thank you sir! much appreciated!

PR #775, opened 2017-10-17
Fixing Image Paste Glitch

### LINKED ISSUE #722: Pasting content with images is failing to load the images.
Pasting content with images is failing to load the images in visual mode.

No content is lost, though.  This is just a refreshing issue.

Steps to repro:

1. Launch the editor with sample content.
2. Copy all the sample content.
3. Go to the empty editor.
4. Paste the copied sample content.

Expected behavior: the pasted images should load.
Actual behavior: they don't load.

Props to @jleandroperez

jleandroperez:
### Details:
In this PR we're:

- Preserving the UIImage scale upon un-archival
- Not overriding a MediaAttachment's **image** with a placeholder, unless we've got an actual URL to request.

Fixes #722

### To test:
1. Launch the editor with sample content.
2. Copy all the sample content.
3. Go to the empty editor.
4. Paste the copied sample content.

Verify that the pasted images do load.

### DIFF
--- a/Aztec/Classes/TextKit/MediaAttachment.swift
+++ b/Aztec/Classes/TextKit/MediaAttachment.swift
@@ -32,7 +32,7 @@ open class MediaAttachment: NSTextAttachment {
 
     /// Identifier used to match this attachment with a custom UIView subclass
     ///
-    private(set) open var identifier: String
+    open var identifier = String()
     
     /// Attachment URL
     ///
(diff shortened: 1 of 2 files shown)


Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -334,8 +334,9 @@ private extension MediaAttachment {
     /// Requests a new asset (asynchronously), and on completion, triggers a relayout cycle.
     ///
     private func updateImage(in textContainer: NSTextContainer?) {
-
-        self.image = delegate!.mediaAttachmentPlaceholderImageFor(attachment: self)
+        if mustDisplayPlaceholder {
```
diegoreymendez: Why not avoid the method and the conditional block completely and just use the guard below to display the placeholder?

The execution path when `url == nil` in this method is a bit weird.

jleandroperez:
@diegoreymendez apologies about over complicating things in this one. All we needed to do is to move the guard up there, not sure why i didn't see that at glance!.

Ready for another look!

diegoreymendez:
@jleandroperez - The images are not loading unless I switch to HTML and back.  Is that a known issue?

jleandroperez:
@diegoreymendez i've found an incredible "Initializer's Loop", in which `ImageAttachment`'s URL was getting reset.

[Fixed via this commit](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/775/commits/0821f77b14f8a796a5d5d2f87593cc4cd5dad602). Mind re-testing?. My steps are as follows:

1. Turn off Wifi
2. Launch Aztec Demo (w/content)
3. Copy all the things
4. Hit back
5. Open the Empty Document screen
6. Toggle Wifi back on
7. Paste

As a result, Bambi should show up shortly.

Thanks in advance!

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -32,7 +32,7 @@ open class MediaAttachment: NSTextAttachment {
 
     /// Identifier used to match this attachment with a custom UIView subclass
     ///
-    private(set) open var identifier: String
+    open var identifier = String()
```
diegoreymendez: This is only used by this class... and there's no public method offering any interaction with it whatsoever...

Why the change?

jleandroperez: Because i've been dealing with the idea of switching this to a `let` constant, ended up reverting... and forgot it was `private(set)`. Nice catch!

jleandroperez:
Thanks sir!

diegoreymendez:
Still approved.

PR #780, opened 2017-10-19
Issue/778 using base property

SergioEstevao:
Another alternative to how we set the alternative styles.

This is done in a way that Lists or Blockquotes can influence the original style set to the paragraph.


diegoreymendez:
@SergioEstevao - Can I close this one?

PR #783, opened 2017-10-19
Improves paragraph style customization support.

diegoreymendez:
### Description:

While we had some support for paragraph customization, this PR make it fully accessible for third party apps.

### Changes:

- No longer uses `PragraphStyle.default` as Aztec's default.
- Relies on values provided to the view on initialization for defaults.  This ensures we can have different defaults for different `TextView`s within the same App.
- Several properties in `ParagraphStyle` are now open, so that Apps can customize them.
- `ParagraphStyle.default` is no longer cached, and is instead instantiated every time.  This is because `ParagraphStyle` is mutable and having a single cached instance means it'll be probably modified by the App.
- `ParagraphStyle.default` can no longer be set.
- Simplifies `AttributedStringSerializer`'s initializer considerably.

### Testing:

1. Go to [this line](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/783/files#diff-e7e4649f5f7465f68641a838cc55412fR22) in the code.
2. Modify [any of the values](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/783/files#diff-bdaaffdf124d77d47d70df9fafaa3bbcR195) now marked as open in `ParagraphStyle`.
3. Run the app and make sure the paragraph style is properly working as the default.

### DIFF
--- a/Aztec/Classes/TextKit/ParagraphStyle.swift
+++ b/Aztec/Classes/TextKit/ParagraphStyle.swift
@@ -1,13 +1,12 @@
 import Foundation
 import UIKit
 
-
 // MARK: - ParagraphStyle
 //
 open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
 
     // MARK: - CustomReflectable
-
+    
     public var customMirror: Mirror {
         get {
             return Mirror(self, children: ["blockquotes": blockquotes,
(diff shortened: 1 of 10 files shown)


Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -1,13 +1,12 @@
...
 import UIKit
 
-
 // MARK: - ParagraphStyle
 //
 open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
 
     // MARK: - CustomReflectable
-
+
```
jleandroperez: Woof.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -202,9 +208,13 @@ open class TextView: UITextView {
 
     // MARK: - Init & deinit
 
-    public init(defaultFont: UIFont, defaultMissingImage: UIImage) {
+    public init(
+        defaultFont: UIFont,
+        defaultParagraphStyle: ParagraphStyle = ParagraphStyle.default,
+        defaultMissingImage: UIImage) {
```
jleandroperez: Since this initializer is multiline, maybe move the `{` to the next one?

diegoreymendez: AFAIK `{` go on the same line as the declaration?

jleandroperez: I've gotta check on the coding standard. When the `{ Initializer, Method }`'s signature is multiline, i always promote the `{` to a new line.

Reason is: the method parameter ends up having the exact same indentation as the method's body itself, and promoting the `{` improves readability.

```
 public init(
         someParameter1: Any,
         someParameter2: Any,
         someParameter3: Any) {        
         self.defaultFont = defaultFont
```

PR #784, opened 2017-10-24
Fix keyboard auto-correction crash issue(if the view is inherited from TextView)

azone:
Fixes keyboard auto-correction crash issue(if the view is inherited from TextView)

To test(recur):
1. Open Aztec.xcworkspace
2. Create new view inherited from TextView
3. Replace TextView to the view just created
4. Type any english words
6. Press space button in keyboard
7. If not crash, press again.

Device: iPhone 7, OS version: iOS 11

### DIFF
--- a/Aztec/Classes/TextKit/TextStorage.swift
+++ b/Aztec/Classes/TextKit/TextStorage.swift
@@ -233,6 +233,13 @@ open class TextStorage: NSTextStorage {
 
         return textStore.attributes(at: location, effectiveRange: range)
     }
+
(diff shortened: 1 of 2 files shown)


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -231,6 +233,13 @@ open class TextStorage: NSTextStorage {
 
         return textStore.attributes(at: location, effectiveRange: range)
     }
+
```
jleandroperez: These changes (TextStorage) don't seem related to the crash fix.  Is that correct?

PR #786, opened 2017-10-27
Issue/8045 aztec image links

jim-rhoades:
This is the Aztec related work needed to implement [adding links to images in posts from its detail editor](https://github.com/wordpress-mobile/WordPress-iOS/issues/8045).

One small detail to note is that in the example app, the `linkURL` is automatically set to the same thing as the file URL after the image is added in EditorDemoController.

Doing it that way, rather than automatically setting it within ImageAttachment, leaves it up to the developer integrating AztecEditor into their app to decide whether or not they want to automatically link images to themselves.

To test:
1. run the example app and tap "Empty Editor Demo"
2. tap the text view to bring up the format bar
3. tap the "+" button in the lower left corner to add an image
4. once you've added the image, tap on it to show "Tap for options"
5. tap the image again, then tap "Media Details"
6. note that the link URL is shown (change it to something else if you want)
7. tap "Done"
8. bring up the format bar again and tap the `</>` button to show that the link URL is shown within the  HTML (with an `<a href></a>` containing the URL around the img tag)
9. if you want, edit the URL within the HTML
10. tap the `</>` button again to show that the HTML is properly converted back to an NSAttributedString... and that the image displays properly
11. bring up the image media details again, and observe that the link URL is still set (because it was set after converting back from HTML)

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -716,7 +716,7 @@ private extension AttributedStringParser {
             element = ElementNode(type: .a)
         }
 
-        element.updateAttribute(named: "href", value: .string(urlString))
+        element.updateAttribute(named: HTMLLinkAttribute.Href.rawValue, value: .string(urlString))
 
         return element
     }
@@ -901,7 +901,7 @@ private extension AttributedStringParser {
 
         if let linkText = attachment.linkURL?.absoluteString {
             let hrefValue = Attribute.Value(withString: linkText)
-            let hrefAttribute = Attribute(name: "href", value: hrefValue)
+            let hrefAttribute = Attribute(name: HTMLLinkAttribute.Href.rawValue, value: hrefValue)
             let linkElement = ElementNode(type: .a, attributes: [hrefAttribute], children: [element])
 
             return linkElement
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -898,17 +898,17 @@ private extension AttributedStringParser {
             }
             element.updateAttribute(named: key, value: .string(finalValue))
         }
-		
+
 		// if the image attachment has a linkURL set,
 		// wrap the .img element node in a .a node
-		if let linkText = attachment.linkURL?.absoluteString {
-			let hrefValue = Attribute.Value(withString: linkText)
-			let hrefAttribute = Attribute(name: "href", value: hrefValue)
-			let linkElement = ElementNode(type: .a, attributes: [hrefAttribute], children: [element])
-			
-			return linkElement
-		}
-		
+        if let linkText = attachment.linkURL?.absoluteString {
+            let hrefValue = Attribute.Value(withString: linkText)
+            let hrefAttribute = Attribute(name: "href", value: hrefValue)
+            let linkElement = ElementNode(type: .a, attributes: [hrefAttribute], children: [element])
+
+            return linkElement
+        }
+
         return element
     }
 
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -902,7 +902,7 @@ private extension AttributedStringParser {
 		// if the image attachment has a linkURL set,
 		// wrap the .img element node in a .a node
 		if let linkText = attachment.linkURL?.absoluteString {
-			let hrefValue = Attribute.Value(withCSSString: linkText)
+			let hrefValue = Attribute.Value(withString: linkText)
 			let hrefAttribute = Attribute(name: "href", value: hrefValue)
 			let linkElement = ElementNode(type: .a, attributes: [hrefAttribute], children: [element])
 			
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -899,8 +899,6 @@ private extension AttributedStringParser {
             element.updateAttribute(named: key, value: .string(finalValue))
         }
 
-		// if the image attachment has a linkURL set,
-		// wrap the .img element node in a .a node
         if let linkText = attachment.linkURL?.absoluteString {
             let hrefValue = Attribute.Value(withString: linkText)
             let hrefAttribute = Attribute(name: "href", value: hrefValue)
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -898,7 +898,17 @@ private extension AttributedStringParser {
             }
             element.updateAttribute(named: key, value: .string(finalValue))
         }
-
+		
+		// if the image attachment has a linkURL set,
+		// wrap the .img element node in a .a node
+		if let linkText = attachment.linkURL?.absoluteString {
+			let hrefValue = Attribute.Value(withCSSString: linkText)
+			let hrefAttribute = Attribute(name: "href", value: hrefValue)
+			let linkElement = ElementNode(type: .a, attributes: [hrefAttribute], children: [element])
+			
+			return linkElement
+		}
+		
         return element
     }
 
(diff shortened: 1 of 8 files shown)


diegoreymendez:
@jim-rhoades - Can you provide some sample testing steps?  You can check out some of our other PRs for examples.

jim-rhoades:
Testing steps added!

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -108,9 +108,25 @@ class AttributedStringSerializer {
...
         } else {
-            for child in element.children {
-                let childContent = serialize(child, inheriting: childAttributes)
-                content.append(childContent)
+            // if the element is of type '.a' and contains a '.img'
+            if element.isNodeType(.a) && element.children.count == 1 {
+                if let imgElement = element.children.first as? ElementNode, imgElement.isNodeType(.img) {
+
+                    // get the link URL and assign it to the image attachment
+                    if let linkText = element.stringValueForAttribute(named: "href") {
```
diegoreymendez: Logic: If this is false, we're losing content.

We should actually set `linkText` to `""` when `href` is missing.

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -108,9 +108,25 @@ class AttributedStringSerializer {
...
+                        let imgAttributes = self.attributes(for: imgElement, inheriting: attributes)
+                        if let attachment = imgAttributes[NSAttachmentAttributeName] as? ImageAttachment,
```
diegoreymendez: Logic: There's a problem here since if this is false, we're losing content.

The truth is that the code we currently have doesn't support this scenario very well, so I'd recommend to deal with the other comments, and once that's done let's discuss this part of the code in Slack to try and provide a nice solution for it.

jim-rhoades:
I've made some changes to address the first two issues you mentioned, which also fixed the problem with one of the tests not passing.

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -108,19 +108,19 @@ class AttributedStringSerializer {
...
-                        let imgAttributes = self.attributes(for: imgElement, inheriting: attributes)
-                        if let attachment = imgAttributes[NSAttachmentAttributeName] as? ImageAttachment,
-                            let attributedString = implicitRepresentation(for: imgElement, inheriting: imgAttributes) {
-                            attachment.linkURL = URL(string: linkText)
-                            content.append(attributedString)
-                        }
-                    }
+            // if the node is of type '.a' with one child of type '.img', retrieve the '.img' node
+            if let imgElement = linkedImageElementFor(element) {
+                // get the link URL text
```
diegoreymendez: We'd normally avoid explaining in comments what the code does, since the code itself is readable enough to describe it.

In cases where the code is not easy to read, it normally means some effort can be put towards making it more readable (like you did with the method `linkedImageElementFor()`).

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -108,9 +108,19 @@ class AttributedStringSerializer {
         if let representation = implicitRepresentation(for: element, inheriting: childAttributes) {
             content.append(representation)
         } else {
-            for child in element.children {
-                let childContent = serialize(child, inheriting: childAttributes)
-                content.append(childContent)
+            if let imgElement = linkedImageElement(for: element) {
+                let linkText = element.stringValueForAttribute(named: "href") ?? ""
+                let imgAttributes = self.attributes(for: imgElement, inheriting: attributes)
+                if let attachment = imgAttributes[NSAttachmentAttributeName] as? ImageAttachment,
```
diegoreymendez: The specific problem we have at this point is that if this condition fails, we're losing content, since the element has no representation in the attributed string.

Upon inspecting the code in more depth I realized both of these conditions should succeed if we're dealing with an image element.  If they ever failed it would mean we have broken logic somewhere.

The ideal solution would be to reengineer our formatters, but that'd be too big of a project for the scope of the trial.

In order to avoid a silent failure, we could use something like this:

```swift
let attachment = imgAttributes[NSAttachmentAttributeName] as! ImageAttachment
let attributedString = implicitRepresentation(for: imgElement, inheriting: imgAttributes)!
```

PS: I know it can look counter-intuitive but it's basically a decision between silently ignoring an error situation vs letting it "explode" and more noticeable (so we can fix it).

jim-rhoades: Normally I would try to avoid force unwrapping optionals, but I can see how "letting it explode" would be better in this case - to make the failure noticeable. Will make that change momentarily.

Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
```diff
@@ -899,6 +899,14 @@ private extension AttributedStringParser {
             element.updateAttribute(named: key, value: .string(finalValue))
         }
 
+        if let linkText = attachment.linkURL?.absoluteString {
+            let hrefValue = Attribute.Value(withString: linkText)
+            let hrefAttribute = Attribute(name: "href", value: hrefValue)
```
frosty: I'm not very familiar with the Aztec codebase, but `"href"` here feels like something we could use a constant for, instead of using string literals.

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -371,6 +378,15 @@ private extension AttributedStringSerializer {
         guard let elementType = element.standardName else {
             return nil
         }
+        
+        if let imgElement = linkedImageElement(for: element) {
+            let linkText = element.stringValueForAttribute(named: "href") ?? ""
```
frosty: See earlier comment about `"href"`

Example/Example/AttachmentDetailsViewController.swift
```diff
@@ -7,10 +7,11 @@ class AttachmentDetailsViewController: UITableViewController
     @IBOutlet var alignmentSegmentedControl: UISegmentedControl!
     @IBOutlet var sizeSegmentedControl: UISegmentedControl!
     @IBOutlet var sourceURLTextField: UITextField!
+    @IBOutlet var linkURLTextField: UITextField!
     @IBOutlet var altTextField: UITextField!
 
     var attachment: ImageAttachment?
-    var onUpdate: ((ImageAttachment.Alignment, ImageAttachment.Size, URL, String?) -> Void)?
+    var onUpdate: ((ImageAttachment.Alignment, ImageAttachment.Size, URL, URL?, String?) -> Void)?
```
frosty: This closure is getting a little unwieldy, particularly for the urls and strings. I wonder if we should either add a comment that clarifies what each parameter does, or simply add names to the parameters: 

```
var onUpdate: ((_ alignment: ImageAttachment.Alignment, _ size: ImageAttachment.Size, _ imageURL: URL, _ linkURL: URL?, _ altText: String?) -> Void)?
```

Example/Example/AttachmentDetailsViewController.swift
```diff
@@ -67,7 +70,11 @@ class AttachmentDetailsViewController: UITableViewController
             return
         }
         let alt = altTextField.text
-        onUpdate(alignment.toAttachmentAlignment(), size.toAttachmentSize(), url, alt)
+        var linkURL: URL?
+        if let linkText = linkURLTextField.text {
+            linkURL = URL(string: linkText)
```
frosty: We could condense this like:

```
var linkURL = URL(string: linkURLTextField.text ?? "")
```

If you pass in an empty string, you get back `nil`.

Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
jim-rhoades: You're right, and I see there's already a `HTMLLinkAttribute` enum with a constant for that. I'll change it.

Example/Example/AttachmentDetailsViewController.swift
jim-rhoades: The rest of the code seems to favor making things self-explanatory, rather than adding comments... so I think adding names to the parameters is a good idea. Making that change.

Unfortunately that doesn't make the point where you're using the closure any more readable since it's still:
`onUpdate(alignment.toAttachmentAlignment(), size.toAttachmentSize(), url, linkURL, alt)`

(because "[Function types cannot have argument labels](https://forums.developer.apple.com/thread/60346)" ?  meh)

frosty: Right! But I guess at least in the sample you game, the variables you’re passing in all have explanatory names :)

jim-rhoades: 👍

PR #793, opened 2017-10-30
Make blockquote appearance customizable

azone:
Make blockquote appearance customizable


jleandroperez:
@azone we'll provide an **UIAppearance-friendly** mechanism, which allows us not to open up the LayoutManager.

This way we'll gain customizations for Traits + Containers 'for free', and we'll avoid the need of casting the `textView.layoutManager` as `LayoutManager` (which results in less code!).

Thanks for helping us out!!

Closed in favor of #794.

PR #794, opened 2017-10-30
Customizable Blockquote Style

jleandroperez:
### Description:
In this PR we're adding three new TextView properties:

- blockquoteBorderColor
- blockquoteBackgroundColor
- preBackgroundColor

### Details:
Because of the `dynamic` keyword, the three new properties are compatible with Apple's `UIAppearance()` mechanism. This allows us to do the following:

- Customize a specific instance:  `textView.blockquoteBackgroundColor = .green`
- Customize **all** of Aztec's instances: `TextView.appearance().blockquoteBackgroundColor = .green`

Apple's `appearance()` mechanism also allows for customization when contained in a specific ViewController, or define the UX for a specific set of traits.

### Testing:
1. Open `EditorDemoViewController.swift`
2. Paste this snippet in viewDidLoad `textView.blockquoteBackgroundColor = .green`
3. Launch Aztec
4. Insert a Blockquote

Verify that the blockquote is green.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -124,7 +124,7 @@ public protocol TextViewFormattingDelegate: class {
 //
 open class TextView: UITextView {
 
-    // MARK: - Properties: Attachments & Media
+    // MARK: - Aztec Delegates
 
     /// The media delegate takes care of providing remote media when requested by the `TextView`.
     /// If this is not set, all remove images will be left blank.
@@ -135,13 +135,15 @@ open class TextView: UITextView {
     ///
     fileprivate var textAttachmentImageProvider = [TextViewAttachmentImageProvider]()
 
-    // MARK: - Properties: Formatting
-
+    /// Formatting Delegate: to be used by the Edition's Format Bar.
+    ///
     open weak var formattingDelegate: TextViewFormattingDelegate?
+
 
     // MARK: - Properties: Text Lists
 
     var maximumListIndentationLevels = 7
+
 
     // MARK: - Properties: UI Defaults
 
@@ -154,6 +156,7 @@ open class TextView: UITextView {
                 NSParagraphStyleAttributeName: defaultParagraphStyle]
     }
 
+
     // MARK: - Properties: Processors
 
     /// This processor will be executed on any HTML you provide to the method `setHTML()` and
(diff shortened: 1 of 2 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -135,14 +135,16 @@ open class TextView: UITextView {
     ///
     fileprivate var textAttachmentImageProvider = [TextViewAttachmentImageProvider]()
 
-    // MARK: - Properties: Formatting
```
diegoreymendez: Why are we removing the mark?

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -165,12 +168,54 @@ open class TextView: UITextView {
     ///
     public var outputProcessor: Processor?
 
-    // MARK: - Properties: Text Storage
+
+    // MARK: - TextKit Aztec Subclass'ess
```
diegoreymendez: `/s/Subclass'ess/Subclasses`

No need for the apostrophe / single-quote.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -206,6 +251,7 @@ open class TextView: UITextView {
     }
 
 
+
```
diegoreymendez: No need for 3 lines.

Aztec/Classes/TextKit/TextView.swift
diegoreymendez: PS: if we want to rename this I'm fine, but it's a good one since all of it's content is related to formatting.

jleandroperez: Related to the change few lines above. This is part of the [delegates section](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/794/files/3836eb1eff0bf936c76da6b30e65a45b6b816463#diff-8d8015f3f021c0252c1569e4965dce80L127), so it's a bit of redundant

Aztec/Classes/TextKit/TextView.swift
jleandroperez: Thanks!

jleandroperez:
Thanks for the review sir!

azone:
Sounds good, but it can't customize the border width, could you support it? Sometimes, I may need it wider or thinner.

PR #795, opened 2017-10-31
Fixed large text memory issue

### LINKED ISSUE #791: Large text may cause TextView crash (Memory issue)
### Expected behavior

Not crash

### Actual behavior

crashed

1. Input or paste about over 30,000 characters
2. Scroll TextView if not crash

Here is the screenshot of Instruments:

azone:
Fixes #791 

To test:

1. Input or paste about over 30,000 characters
2. Scroll TextView


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -231,6 +233,13 @@ open class TextStorage: NSTextStorage {
 
         return textStore.attributes(at: location, effectiveRange: range)
     }
+
+    private func replaceTextStoreString(_ range: NSRange, with string: String) {
```
diegoreymendez: Considering `String.replaceSubrange()` is of complexity `O(m)` for most scenarios (where m is the sum of the length of the cached string and the modification), this method should be replaced by:

```
textStoreString = textStore.string
```

Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -239,6 +248,8 @@ open class TextStorage: NSTextStorage {
         detectAttachmentRemoved(in: range)
         textStore.replaceCharacters(in: range, with: str)
 
+        replaceTextStoreString(range, with: str)
```
diegoreymendez: We should move the update logic inside of `override func edited()` (inside the next call in this method), and only trigger it for `.editedCharacters`.

That way we don't need to call both at several point in the app.

diegoreymendez:
@azone - Very nice find and very nice approach.  Added a few comments on further improvements.

diegoreymendez:
@azone - Also, please note the PR has conflicts with the base branch.  Let me know when it's ready again.

diegoreymendez:
Merging since Travis is stuck.

PR #797, opened 2017-10-31
MediaAttachment: correct MediaAttachment width calculating

### LINKED ISSUE #796: MediaAttachment is out of block-quote's bounds
### Expected behavior

img_0261

### Actual behavior

img_0262

Insert image into block-quote (image must be wider than block-quote)

azone:
Fixes #796

To test:

Insert image or video into block-quote


Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -298,22 +298,31 @@ open class MediaAttachment: NSTextAttachment {
...
         if let storage = textContainer?.layoutManager?.textStorage,
-           let paragraphStyle = storage.attribute(NSParagraphStyleAttributeName, at: charIndex, effectiveRange: nil) as? NSParagraphStyle {
-            padding += paragraphStyle.firstLineHeadIndent + paragraphStyle.tailIndent
+            let paragraphStyle = storage.attribute(NSParagraphStyleAttributeName, at: charIndex, effectiveRange: nil) as? NSParagraphStyle {
+            let attachmentString = storage.attributedSubstring(from: NSMakeRange(charIndex, 1)).string
+            let headIndent = storage.string.isStartOfParagraph(at: attachmentString.startIndex) ? paragraphStyle.firstLineHeadIndent : paragraphStyle.headIndent
+
+            padding += headIndent
+
+            if paragraphStyle.tailIndent > 0 {
```
SergioEstevao: Can you explain why do we need this IF, in what scenarios the tailIdent can be positive or negative?

azone: Here is **Discussion** of tailIndent from apple:

> If positive, this value is the distance from the leading margin (for example, the left margin in left-to-right text). If 0 or negative, it’s the distance from the trailing margin.
>
> For example, a paragraph style designed to fit exactly in a 2-inch wide container has a head indent of 0.0 and a tail indent of 0.0. One designed to fit with a quarter-inch margin has a head indent of 0.25 and a tail indent of –0.25.

SergioEstevao: So could we rewrite this like this: 
```padding += abs(paragraphStyle.tailIndent) + abs(headIndent)```

jleandroperez:
Closing in favor of #828.

Thanks **a lot** sir!!

diegoreymendez:
@jleandroperez - Just double checking... this is the issue and #828 is the PR for this report.  Is that correct?

jleandroperez:
@diegoreymendez Nope! #796 is the issue. #828 is Sergio’s fix Mark II, while this one is the original PR (AKA Fix Mark I).

PR #802, opened 2017-11-02
Avoiding to call ensureRemovalOfParagraphAttributesAfterSelectionChange method if TextView have marked text

### LINKED ISSUE #801: TextView's marked text may remove block-quote
### Expected behavior

See the video below:
[ScreenRecording_11-02-2017 10:21.zip](https://github.com/wordpress-mobile/AztecEditor-iOS/files/1436312/ScreenRecording_11-02-2017.10.21.zip)

### Actual behavior

See the Video below:

[ScreenRecording_11-02-2017 10:19.MP4.zip](https://github.com/wordpress-mobile/AztecEditor-iOS/files/1436310/ScreenRecording_11-02-2017.10.19.MP4.zip)

1. Add block-quote
2. Switch input method to Chinese
3. Input Chinese but do not confirm
4. block-quote may removed by continue pressing keyboard

##### Tested on iPhone 7, iOS 1, Aztec beta-14

azone:
Avoiding to call ensureRemovalOfParagraphAttributesAfterSelectionChange method if TextView have marked text

Fixes #801 

To test:

1. Add block-quote
2. Switch input method to Chinese
3. Input Chinese but do not confirm

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -224,9 +224,7 @@ open class TextView: UITextView {
     ///
     override open var typingAttributes: [String: Any] {
         get {
-            if markedTextRange == nil {
-                ensureRemovalOfParagraphAttributesAfterSelectionChange()
-            }
+            ensureRemovalOfParagraphAttributesAfterSelectionChange()
             return super.typingAttributes
         }
         set {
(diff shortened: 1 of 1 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -224,7 +224,9 @@ open class TextView: UITextView {
     ///
     override open var typingAttributes: [String: Any] {
         get {
-            ensureRemovalOfParagraphAttributesAfterSelectionChange()
+            if markedTextRange == nil {
```
jleandroperez: Would it be possible to move this condition inside the `mustRemoveParagraphAttributesAfterSelectionChange` method?

That way the `typingAttributes` override would look the same, and all of the conditions would be contained as follows:

```
    private func mustRemoveParagraphAttributesAfterSelectionChange() -> Bool {
        return selectedRange.location == storage.length
            && storage.string.isEmptyParagraph(at: selectedRange.location)
            && markedTextRange == nil
    }
```

PR #809, opened 2017-11-06
Swift 3.2: Removing .characters references

jleandroperez:
### Details:
In this PR we're removing all of the `String` **.characters** references.

### To test:
- Verify that the unit tests are green
- Verify that the Swift 3.2 warnings are gone

cc @diegoreymendez + @SergioEstevao 

Thanks in advance!


jleandroperez:
Thank you sir!

PR #815, opened 2017-11-08
Make sure VideoAttachment always have the play icon by default.

### LINKED ISSUE #813: Video Attachment should alway have by default an play icon overlay image
### Expected behavior

- If the overlayImage is removed from a VideoAttachment by default the system should restore the play icon overlay

### Actual behavior

- If the system set the image to nil the play overlay goes away and needs to be restored manually

- Activate the media upload error mode
- Add a new video on the demo
- Wait for the error, 
- Tap on the video
- Tap somewhere else

SergioEstevao:
Fixes #813 

To test:
 - Start the empty demo app
 - Add a new media video object
 - Tap on it
 - Tap on another media
 - See that the icon on the video resets to the play button.
 - Activate the error mode by switch the `errorMode` variable to true on the `EditorDemoController`
 - Repeat the steps above.

### DIFF
--- a/Aztec/Classes/TextKit/VideoAttachment.swift
+++ b/Aztec/Classes/TextKit/VideoAttachment.swift
@@ -70,11 +70,7 @@ open class VideoAttachment: MediaAttachment {
     ///
     open override var overlayImage: UIImage? {
         set(newValue) {
-            if newValue == nil {
-                super.overlayImage = Assets.playIcon
-            } else {
-                super.overlayImage = newValue
-            }
+            super.overlayImage = newValue ?? Assets.playIcon
         }
 
         get {
--- a/Aztec/Classes/TextKit/VideoAttachment.swift
+++ b/Aztec/Classes/TextKit/VideoAttachment.swift
@@ -66,6 +66,21 @@ open class VideoAttachment: MediaAttachment {
         super.init(data: contentData, ofType: uti)
     }
 
+    /// An image to display overlaid on top of the media
+    ///
+    open override var overlayImage: UIImage? {
+        set(newValue) {
+            if newValue == nil {
+                super.overlayImage = Assets.playIcon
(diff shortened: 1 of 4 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -1078,17 +1076,16 @@ extension EditorDemoController: TextViewAttachmentDelegate {
...
                 richTextView.refresh(selectedAttachment)
             }
 
             // and mark the newly tapped attachment
-            let message = NSLocalizedString("Options", comment: "Options to show when tapping on a media object on the post/page editor.")
-            attachment.message = NSAttributedString(string: message, attributes: mediaMessageAttributes)
-            if attachment.overlayImage == nil {
-                attachment.overlayImage = Gridicon.iconOfType(.pencil).withRenderingMode(.alwaysTemplate)
-                attachment.overlayImage = Gridicon.iconOfType(.pencil, withSize: CGSize(width: 32.0, height: 32.0)).withRenderingMode(.alwaysTemplate)
+            if (attachment.message == nil) {
```
jleandroperez: No need for braces ` ( ) `

Aztec/Classes/TextKit/VideoAttachment.swift
```diff
@@ -66,6 +66,21 @@ open class VideoAttachment: MediaAttachment {
...
 
+    /// An image to display overlaid on top of the media
+    ///
+    open override var overlayImage: UIImage? {
+        set(newValue) {
+            if newValue == nil {
+                super.overlayImage = Assets.playIcon
+            } else {
+                super.overlayImage = newValue
+            }
```
jleandroperez: This one can be simplified as follows:

```
set {
    super.overlayImage = newValue ?? Assets.playIcon
}
```

Issue #771, opened 2017-10-16
Crash: HTMLStorage


elibud:
Crashlytics # 59dcf11abe077a4dcc102c88
```
0  Aztec                          0x100f1b0a0 @objc HTMLStorage.init() -> HTMLStorage (HTMLStorage.swift)
1  UIFoundation                   0x190efb96c -[NSLayoutManager(NSLayoutManager_Animation) destinationTextContainerForRange:inTextContainer:] + 200
2  UIFoundation                   0x190efbf44 -[NSLayoutManager(NSLayoutManager_Animation) layoutFragmentsForReplacingCharactersInRange:withAttributedString:rect:textContainer:] + 936
3  UIKit                          0x1909f18a8 -[UITextView _performPasteOfAttributedString:toRange:animator:completion:] + 752
4  UIKit                          0x1902d4e18 -[UITextPasteController _performPasteOfAttributedString:toRange:progressSupport:animator:completion:] + 908
5  UIKit                          0x1902d42f4 __58-[UITextPasteController _executePasteForSession:animator:]_block_invoke + 336
6  libdispatch.dylib              0x186351048 _dispatch_client_callout + 16
7  libdispatch.dylib              0x18635f2a0 _dispatch_sync_thread_bound_invoke + 68
8  libdispatch.dylib              0x186351048 _dispatch_client_callout + 16
9  libdispatch.dylib              0x18635db74 _dispatch_main_queue_callback_4CF$VARIANT$mp + 1016
10 CoreFoundation                 0x186973f20 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12
11 CoreFoundation                 0x186971afc __CFRunLoopRun + 2012
12 CoreFoundation                 0x1868922d8 CFRunLoopRunSpecific + 436
13 GraphicsServices               0x188723f84 GSEventRunModal + 100
14 UIKit                          0x18fe3f880 UIApplicationMain + 208
15 WordPress                      0x100236d28 main (main.m:11)
16 libdyld.dylib                  0x1863b656c start + 4
```

diegoreymendez:
For completeness, I believe these reports are related:

#230494-h
iPad Pro
iOS 11.0.2
WordPress 8.4 (@pmciano was unable to replicate on 8.5)

> Whenever I try to copy or paste something in HTML mode, the app invariably crashes.

#231525-h
iPhone 7 Plus
iOS 11.0.1
[...]

Issue #768, opened 2017-10-16
Loosing styles upon press backspace key when no contents in TextView


azone:
### Expected behavior

Has default typingAttributes

### Actual behavior

None Attributes (All attributes are removed)

### Steps to reproduce the behavior

1. Open Editor Demo
2. Delete all contents in TextView (Skip this step if no contents)
3. Continue press backspace key

Issue #465, opened 2017-05-08
Crash when editing after uknown HTML.


diegoreymendez:
<h3>Steps to reproduce:</h3>

1. Launch the empty editor
2. Switch to HTML mode
3. Type `<asd>`
4. Switch to visual mode
5. Move the caret after the unknown HTML attachment.
6. Type anything

Result: crash!

<h3>Additional info:</h3>

Seems like a range-mapping issue.

diegoreymendez:
Fixed.

PR #819, opened 2017-11-10
Add UI tests for high priority issues

brbrr:
Will incrementally add UI tests to existing High priority issues

Covered:
#385 
#675 
#465 
#771 
#768 

Fixed: #839


Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        richTextField.tap()
+    }
+    
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+    
+    // Gihtub issue #393
+    func testLongTitle() {
```
rachelmcr: This test only passes on smaller screens — it fails on larger screens (I checked on an iPad Pro) because the text isn't long enough to wrap there. We should make sure the test accounts for different screen sizes and passes on any device.

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        let titleLineHeight = 22
+        
+        let titleTextView = app.textViews[elementStringIDs.titleTextField]
+        titleTextView.tap()
+        
+        let oneLineTitleHeight = Int(titleTextView.frame.height)
+        
+        titleTextView.typeText("very very very very very very long title in a galaxy not so far away")
+        let twoLineTitleHeight = Int(titleTextView.frame.height)
+        XCTAssert(twoLineTitleHeight - oneLineTitleHeight == titleLineHeight )
```
rachelmcr: Instead of setting `titleLineHeight` (fragile due to device settings or design changes) and using it in this calculation, what about using `oneLineTitleHeight` and asserting that `twoLineTitleHeight` is twice that? (Or just that `twoLineTitleHeight` is a multiple of `oneLineTitleHeight` if we want to confirm that the title area expands.)

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
+//
+//  HighPriorityIssuesTests.swift
+//  AztecUITests
+//
+//  Created by brbrr on 11/10/17.
+//  Copyright © 2017 Automattic Inc. All rights reserved.
+//
+
```
rachelmcr: You can remove these default header comments from the top of the file.

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        
+        let oneLineTitleHeight = Int(titleTextView.frame.height)
+        
+        titleTextView.typeText("very very very very very very long title in a galaxy not so far away")
+        let twoLineTitleHeight = Int(titleTextView.frame.height)
+        XCTAssert(twoLineTitleHeight - oneLineTitleHeight == titleLineHeight )
+        //        XCTAssert(oneLineTitleHeight < twoLineTitleHeight)
+    }
+    
+    func testNewlinesInTitle() {
```
rachelmcr: I know this is the current behavior, but should Aztec allow new lines in the title field? cc @diegoreymendez for your thoughts on this.

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        let richTextField = app.textViews[elementStringIDs.richTextField]
+        richTextField.tap()
+    }
+    
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+    
+    // Gihtub issue #393
```
rachelmcr: I'd go ahead and add the full link to the issue here, to make it easier to find. (Also, just a small typo in "Gihtub")

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        XCTAssert(twoLineTitleHeight - oneLineTitleHeight == titleLineHeight )
+        //        XCTAssert(oneLineTitleHeight < twoLineTitleHeight)
+        
+        titleTextView.typeText("\nline 3")
+        let threeLineTitleHeight = Int(titleTextView.frame.height)
+        XCTAssert(threeLineTitleHeight - twoLineTitleHeight == titleLineHeight )
+        //        XCTAssert(twoLineTitleHeight < threeLineTitleHeight)
+    }
+    
+    func testInfinitLoopOnAssetDownload() {
```
rachelmcr: What issue does this test? I think it's worth adding comments with links to all the issues we're testing, for reference later on. (Also, just a small note that there's a typo - "Infinit" should be "Infinite")

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        switchContentView()
+        enterTextInHTML(text: "<img src=\"https://someinvalid.url/with-an-invalid-resource\">")
+        switchContentView()
+        gotoRootPage()
+        
+        let editorDemoButton = app.tables/*@START_MENU_TOKEN@*/.staticTexts["Empty Editor Demo"]/*[[".cells.staticTexts[\"Empty Editor Demo\"]",".staticTexts[\"Empty Editor Demo\"]"],[[[-1,1],[-1,0]]],[0]]@END_MENU_TOKEN@*/
```
rachelmcr: You can use `staticTexts[elementStringIDs.emptyDemo]` here

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        XCTAssert(editorDemoButton.exists, "Editor button not hittable. Arew you on right page?")
```
rachelmcr: Just a minor suggested edit:

`"Editor button not hittable. Are you on the right page?"`

Example/AztecUITests/XCTest+Extensions.swift
```diff
@@ -35,14 +36,41 @@ public struct elementStringIDs {
 
 extension XCTest {
     /**
+     Common method to type in different text fields
+     */
+    func typeToTextField(text: String, to: String) -> Void {
```
rachelmcr: Nice! I like these updates you've made to the helper functions.

Example/AztecUITests/HighPriorityIssuesTests.swift
brbrr: ✅

Example/AztecUITests/HighPriorityIssuesTests.swift
brbrr: ✅

Example/AztecUITests/HighPriorityIssuesTests.swift
brbrr: I totally agree with you and I was thinking about that one. The thing is that `oneLineTitleHeight` is about `38px` (due to underline and some padding), but every other line is `22px`. So its quite complicated to confirm that in a nice way. We can confirm that it's just bigger, but I don't think its good approach either.

Example/AztecUITests/HighPriorityIssuesTests.swift
brbrr: wow! That is some fancy XCode recording stuff holding few locator values. I'll replace it with the single value you suggest.

Example/AztecUITests/HighPriorityIssuesTests.swift
rachelmcr: Ah right, I see what you mean. I agree that it'd be best to confirm a specific change in line height, so it might be best to stick with this approach. I'll keep thinking about it. (In the future, if we implement visual regression tests for Aztec we could confirm in the UI test that it's just bigger and use the visual regression test to make sure the entire title is visible. I wouldn't rely on that until we start implementing those visual tests, though.)

Example/AztecUITests/HighPriorityIssuesTests.swift
brbrr: Added GH issue ✅

Example/AztecUITests/HighPriorityIssuesTests.swift
rachelmcr: Oh, I should clarify that I was thinking of replacing `staticTexts["Empty Editor Demo"]` with `staticTexts[elementStringIDs.emptyDemo]`. We can remove the comments here and just use `let editorDemoButton = app.tables.staticTexts[elementStringIDs.emptyDemo]`.

diegoreymendez:
@brbrr, @catehstn, @rachelmcr - While reviewing these tests I can't really help it but notice we're devoting a lot of effort into UI testing the library's example App.

While this is good for Aztec as a library, most of the UI code is not directly affecting WPiOS (since Aztec as a library only provides the editor view).

I feel like it may be a good idea to start pushing at least part of these into WPiOS, instead.  Thoughts?

brbrr:
@diegoreymendez Generally I agree with you. Our (mine w/ @rachelmcr) plan was to cover all High priority issues (which are worth to cover w/ UI tests) here since test flows are mostly short. So we don't need to bother with login flow etc.

After that is done (and maybe some other simple formatting tests) we will switch to WP apps. Is it sounds reasonable to you? 

_And maybe we should move this discussion elsewhere (call or P2)_

diegoreymendez:
Just to clarify though.  I believe that the work you're doing here is great for several different reasons.

Ideally we'd have both the library and the app covered by a similar set of tests.

Thanks for taking care, everyone!

brbrr:
@rachelmcr may I ask your attention to this one?

Example/AztecUITests/Pages/BlogsPage.swift
```diff
@@ -0,0 +1,36 @@
...
+        
+        showOptionsStrip()
+    }
+    
+    func showOptionsStrip() -> Void {
+        app.textViews[elementStringIDs.richTextField].tap()
+        expandOptionsSctrip()
+    }
+    
+    func expandOptionsSctrip() -> Void {
```
rachelmcr: On larger devices (e.g. iPad Pro) this function is toggling the extended toolbar open/closed rather than always expanding it. We should check the status of the toolbar before tapping — otherwise the expanded toolbar is toggled off in every other test.

If it isn't currently possible to check the status, we should open an issue to address that as it's also an accessibility issue. (VoiceOver users need to be aware of the toolbar status so they can expand it if needed, too.)

Example/AztecUITests/HighPriorityIssuesTests.swift
```diff
@@ -0,0 +1,147 @@
...
+        blogsPage.gotoEmptyDemo()
+    }
+    
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+    
+    // Github issue https://github.com/wordpress-mobile/AztecEditor-iOS/issues/385
+    func testLongTitle() {
```
rachelmcr: This test is still failing on larger devices (e.g. iPad Pro). If there isn't a way to get it to work for both large and small devices, could we restrict it to only run when testing smaller devices?

Example/AztecUITests/HighPriorityIssuesTests.swift
rachelmcr: I don't think we really need this test, given that this isn't the intended behavior. See: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/821

Example/AztecUITests/XCTest+Extensions.swift
```diff
@@ -35,18 +37,50 @@ public struct elementStringIDs {
...
     func enterTextInField(text: String) -> Void {
-        let app = XCUIApplication()
-        let richTextField = app.textViews[elementStringIDs.richTextField]
-
-        richTextField.typeText(text)
+        typeToTextField(text: text, to: elementStringIDs.richTextField)
+//        let app = XCUIApplication()
+//        let richTextField = app.textViews[elementStringIDs.richTextField]
+//
+//        richTextField.typeText(text)
```
rachelmcr: I'd go ahead and remove this commented-out section.

Example/AztecUITests/Pages/BlogsPage.swift
brbrr: #837

Example/AztecUITests/HighPriorityIssuesTests.swift
brbrr: I don't think its possible to tag which tests to run with current test runner (as it's done in wp-e2e-tests). The only solution I see is to move such tests to separate files and run them only for small screens. And in general, I think it would be better to address this issue once we'll have them running on multiple devices.

Example/AztecUITests/Pages/BlogsPage.swift
rachelmcr: For now, how about checking to see if the HTML button exists before toggling the extended toolbar? I'm thinking of something like this:

```
        let expandButton = app.children(matching: .window).element(boundBy: 1).children(matching: .other).element.children(matching: .other).element.children(matching: .other).element.children(matching: .button).element
        let htmlButton = app.scrollViews.otherElements.buttons[elementStringIDs.sourcecodeButton]
        
        if expandButton.exists && expandButton.isHittable && !htmlButton.exists {
            expandButton.tap()
        }
```

That way we don't break the existing tests on iPads, and we can improve the check later once that issue is resolved.

Example/AztecUITests/HighPriorityIssuesTests.swift
rachelmcr: I looked around and found [these hints and tips](http://averagepro.com/2015/11/06/xctest-ui-testing-hints-and-tips/) that include conditional code for testing on iPhone v. iPad. I updated the example shown there and got it to work in this test like this:

```
if (app.windows.element(boundBy: 0).horizontalSizeClass == .compact || app.windows.element(boundBy: 0).verticalSizeClass == .compact) {
            titleTextView.typeText("very very very very very very long title in a galaxy not so far away")
        } else {
            titleTextView.typeText("very very very very very very long title in a galaxy not so far away very very very very very very long title in a galaxy not so far away")
        }
```

Example/AztecUITests/Pages/BlogsPage.swift
brbrr: That's nice solution, thanks!

rachelmcr:
The changes look good to me. It looks like `testCopyPasteCrash` failed on Travis CI — I'm not sure what's going on there; I haven't had any trouble with it on local runs.

diegoreymendez:
Just a reminder to set the appropriate milestone when closing a PR or issue.  Thanks!

PR #814, opened 2017-11-08
Fixed Chinese input issue in header 1~6

azone:
Fixes #811 

To test:

1. Open editor
2. Press return
3. Turn Heading on (h1 or h2, ...)
4. Input any Chinese characters and confirm
5. Continue input Chinese characters and confirm

Issue #811, opened 2017-11-07
Can't input Chinese characters


azone:
### Expected behavior

Input Chinese characters in heading as usual

### Actual behavior

Can't input Chinese characters in heading after first confirm if there is empty line before heading.
See the video below:

[ScreenRecording_11-07-2017 17:26.MP4.zip](https://github.com/wordpress-mobile/AztecEditor-iOS/files/1449413/ScreenRecording_11-07-2017.17.26.MP4.zip)

### Steps to reproduce the behavior

1. Open editor
2. Press return
3. Turn Heading on (h1 or h2, ...)
4. Input any Chinese characters and confirm
5. Continue input Chinese characters and confirm (can't  input anything in this step)

All those invisible Chinese characters will appears if rotate the device.

jleandroperez:
### Note: 
In order to reproduce this one, please, set up the Chinese Simplified Pinyin keyboard.

Thank you!!

PR #820, opened 2017-11-10
Chinese H1 Input: Running only upon replacement

jleandroperez:
### Details:
In this PR we're updating the workaround introduced via #814:

- It'll be executed only when needed: upon text replacement
- We're switching to `fixAttributes` (vs `invalidateAttributes`)
- Added new unit tests!

Ref. #811

Needs Review: @diegoreymendez 

### To test:
1. Verify that the unit tests are green
2. Launch the empty editor
3. Insert a newline
4. Switch to the **Chinese Simplified Pinyin** keyboard
5. Type `s` and tap over any of the replacement suggestions
6. Repeat the step above. 

Verify that the keyboard suggestion selected in (6) remains onscreen.


Aztec/Classes/TextKit/TextStorage.swift
```diff
@@ -267,8 +267,15 @@ open class TextStorage: NSTextStorage {
...
-        let invalidateRange = NSMakeRange(range.location, attrString.length)
-        invalidateAttributes(in: invalidateRange)
+        // Whenever we're actually replacing text, let's trigger a `fixAttributes` call. This is done to prevent a glitch in which
+        // TextView may not render characters, in the same line, that carry partitioned attributes.
+        //
+        // Ref. https://github.com/wordpress-mobile/AztecEditor-iOS/issues/811
+        //
+        if range.length > 0  {
+            let invalidateRange = NSMakeRange(range.location, attrString.length)
+            fixAttributes(in: invalidateRange)
```
diegoreymendez: Is there any specific reason for calling `fixAttributes()` instead of `invalidateAttributes()`?

Invalidate attributes does the same, without forcing a synchronous execution when lazy attribute fixing is on.  See the discussion here: https://developer.apple.com/documentation/uikit/nstextstorage/1534025-invalidateattributes

jleandroperez:
@diegoreymendez thanks for bringing that up!. Replaced `fixAttributes` > `invalidateAttributes` accordingly. Ready for another look!

jleandroperez:
Closing in favor of #836. @diegoreymendez is a genius!!

PR #823, opened 2017-11-13
Restore Swift 3.2 support to Aztec.

### LINKED ISSUE #822: Restore Swift 3.2 support in Aztec.
### Description

We need to restore Swift 3.2 support.

### Details

Our migration to Swift 4.0 made it impossible to build Aztec with Swift 3.2.  This is not great because there seems to be incompatibility issues with Swift 3.2 targets interfacing with our 4.0 library due to issues with `NSAttributedStringKey` between versions.

screen shot 2017-11-10 at 11 54 29

diegoreymendez:
## Description

Fixes #822 

This PR adds back support for building Aztec as a Swift 3.2 target, since `NSAttributedStringKey` has some compatibility issues when mixing Swift 3.2 and 4.0 targets.

This is marked as high priority since it's blocking integration with Apps using a different Swift version than 4.0 (such as WPiOS right now).

### WPiOS Integration Test

The sister branch you can use to test the WPiOS integration is `try/integrate-aztec-swift-32`.

### Build Tests

Build the example app and run the unit tests with these configurations.

#### Configuration 1:

Example App: Swift 4.0
Aztec: Swift 4.0

#### Configuration 2:

Example App: Swift 4.0
Aztec: Swift 3.2

**IMPORTANT:** Changing the Example App to 3.2 is not recommended because the code in the Example App is not compatible with Swift 3.2.


diegoreymendez:
Thank you @SergioEstevao !

PR #825, opened 2017-11-15
Japanese text issues for 1.0.0-beta.14.

### LINKED ISSUE #709: Japanese text issues when switching between Visual and HTML modes
There appears to be a few issues while entering multibyte text (in this case Japanese) into Aztec.

japan_text

There are two potential issues here.
 
👉  Tapping enter in HTML mode while multibyte text exists clears it:

0. Make sure Japanese - Kana keyboard is active.
1. Run the editor demo and enter some random text.
2. Switch to HTML mode and hit the return button in the lower right of the keyboard.
3. Notice the text disappears
4. Switch to visual mode — notice the text returns.
5 Switch to HTML mode — notice the text is fine.

👉 Adding additional multibyte text in HTML mode can scramble previous text:

0. Make sure Japanese - Kana keyboard is active.
1. Run the editor demo and enter some random text.
2. Switch to HTML mode and hit the return button in the lower right of the keyboard.
3. Notice the text disappears
4. Switch to visual mode — notice the text returns.
5. Switch to HTML mode — notice the text is fine.
6. Paste some additional multibyte text into the editor (e.g. 私はものが好き)
7. Tap the return button and notice how the preceding text becomes scrambled.
8. Switch modes again (HTML -> Visual -> HTML) and see that the text is restored.

Note that there are a few other ways to trigger the previous text getting scrambled in HTML mode - this is just one example.

##### Tested on iPhone 7+ and Sim, iOS 10.3.3, Aztec `develop`

diegoreymendez:
## Description

Fixes #709 for Beta 14 (It's the same fix as #824 but for Beta 14).

It seems we were missing a call to `edited()`.

## Testing

1. Launch the app
2. Type anything using the Japanese keyboard.
3. Switch to HTML mode and edit some more.
4. Switch back to visual mode and make sure all Japanese characters are properly rendered throughout the session.

### DIFF
--- a/Aztec/Classes/TextKit/HTMLStorage.swift
+++ b/Aztec/Classes/TextKit/HTMLStorage.swift
@@ -83,20 +83,13 @@ open class HTMLStorage: NSTextStorage {
         endEditing()
     }
 
-    override open func addAttribute(_ name: String, value: Any, range: NSRange) {
-        textStore.addAttribute(name, value: value, range: range)
-    }
-
-    override open func removeAttribute(_ name: String, range: NSRange) {
-        textStore.removeAttribute(name, range: range)
-    }
-
     override open func replaceCharacters(in range: NSRange, with str: String) {
         beginEditing()
 
         textStore.replaceCharacters(in: range, with: str)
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Thanks @bummytime !

PR #831, opened 2017-11-17
Image output html css for size full

### LINKED ISSUE #830: Images with class size full don't write the class back to the CSS
### Expected behavior

If you insert an image by default the class size should be full and it should output something like this:

``````

### Actual behavior

It outputs something like this: 

``````

- Start the demo app
- Open the empty demo
- Add an image
- tap on image and select full size
- Switch to HTML and see result

SergioEstevao:
Fixes #830 

To test:
 - Open the demo app
 - Add an image
 - Switch to HTML
 - Check if the full attribute is applied

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -1003,9 +1003,11 @@ private extension AttributedStringParser {
         if attachment.alignment != .center {
             style += attachment.alignment.htmlString()
         }
-
-        style += style.isEmpty ? String() : String(.space)
-        style += attachment.size.htmlString()
+        
+        if attachment.size != .none {
+            style += style.isEmpty ? String() : String(.space)
+            style += attachment.size.htmlString()
+        }
 
         guard !style.isEmpty else {
             return nil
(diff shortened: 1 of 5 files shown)


designsimply:
I tested with iPhone 6S, iOS 11.1.2, WPiOS 8.9.0.20171122 and I was unable to see the `class="size-full"` attribute. 

Steps to reproduce:

1. Go to WPiOS > Blog Posts > + and add a title.
1. Insert an image from the device's media.
1. View the HTML to see if `class="size-full"` is present.
1. Go to WPiOS > Blog Posts > + and add a title.
1. Insert an image from the blog's media library.
1. View the HTML to see if `class="size-full"` is present.

<strong>Result:</strong> when the image was inserted from the device I saw `class="wp-image=11507` and when the image was inserted from the media library there were no attributes other than `src`.

img_1744

img_1745

diegoreymendez:
@designsimply - Can you move this info to the issue [here](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/830)? (as opposed to the PR)

It will help ensure the feedback is properly tracked and there's an explanation if we need to reopen the issue.

Thank you!

SergioEstevao:
Humm by default on the main app I don't think we are setting any size attribute. Should the size-full be the default @designsimply ?

PR #838, opened 2017-11-27
TextView: Customizable default text color

azone:
Customize default text color

To test:

1. Change `textView.textColor = UIColor.darkText` to `textView.textColor = UIColor.red` in demo
2. Run the demo


diegoreymendez:
The problem I see with this PR is that setting the font color should probably not be disconnected from HTML.

@SergioEstevao, @jleandroperez - Do you think there's a case for letting Apps set the text color without it affecting the underlying HTML?

SergioEstevao:
@diegoreymendez I see it as a possible Appearance configuration for the editor. Some people may want to use Aztec with a black background and white text by default.

jleandroperez:
+1 to Sergio's point!

PR #842, opened 2017-11-28
TextView: Export init, setHTML and getHTML methods to Objective-C

azone:
TextView missed some methods in Objective-C


eggli:
Just found these Obj C exports were removed in current beta release, accidentally?

PR #843, opened 2017-11-29
Linked Image Mechanism: Cleanup!

jleandroperez:
### Details:
This PR introduces no new functionality. We're:

1. Encapsulating the **linkedImage** parsing mechanism, into it's own method
2. Moving the `textAttachment` constant over to a String extension, for convenience
3. Adding a small method to ElementNode (and few unit tests!)

### To test:
1. Verify that the unit tests are green
2. Add a linked image `<a href="www.wordpress.com"></a>`
3. Try toggling back and forth, and verity that the HTML is preserved

Needs Review: @diegoreymendez 
Thanks in advance!

### DIFF
--- a/Aztec/Classes/Extensions/NSAttributedString+Attachments.swift
+++ b/Aztec/Classes/Extensions/NSAttributedString+Attachments.swift
@@ -2,6 +2,15 @@ import Foundation
 import UIKit
 
 
+// MARK: - Constants
+//
+extension String {
+    /// String containing the NSTextAttachment Character
+    ///
+    static let textAttachment = String(UnicodeScalar(NSAttachmentCharacter)!)
+}
+
+
 // MARK: - NSAttributedString Extension for Attachments
 //
 extension NSAttributedString
@@ -9,11 +18,6 @@ extension NSAttributedString
     /// Indicates the Attributed String Length of a single TextAttachment
     ///
     static let lengthOfTextAttachment = NSAttributedString(attachment: NSTextAttachment()).length
-
-
-    /// String containing the NSTextAttachment Character
-    ///
-    static let textAttachmentString = String(UnicodeScalar(NSAttachmentCharacter)!)
 
 
 
@@ -24,7 +28,7 @@ extension NSAttributedString
         var attributesWithAttachment = attributes
         attributesWithAttachment[.attachment] = attachment
 
-        self.init(string: NSAttributedString.textAttachmentString, attributes: attributesWithAttachment)
+        self.init(string: .textAttachment, attributes: attributesWithAttachment)
     }
 
     /// Loads any NSTextAttachment's lazy file reference, into a UIImage instance, in memory.
--- a/Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
+++ b/Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
@@ -234,21 +234,29 @@ public class ElementNode: Node {
     }
 
 
-    /// Returns the ElementNode instance, whenever there's a *single* child, of the specified nodeType (or nil otherwise).
+    /// If there's exactly just one child node, this method will return it's instance. Otherwise, nil will be returned
+    ///
+    func onlyChild() -> ElementNode? {
+        guard children.count == 1 else {
+            return nil
+        }
+
+        return children.first as? ElementNode
+    }
+
+
+    /// Returns the child ElementNode of the specified nodeType -whenever there's a *single* child-, or nil otherwise.
     ///
     /// - Parameter type: Type of the 'single child' node to be retrieved.
     ///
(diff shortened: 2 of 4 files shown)


Aztec/Classes/Extensions/NSAttributedString+Attachments.swift
```diff
@@ -2,6 +2,15 @@ import Foundation
 import UIKit
 
 
+// MARK: - Constants
+//
+extension String {
+    /// String containing the NSTextAttachment Character
+    ///
+    static let textAttachment = String(UnicodeScalar(NSAttachmentCharacter)!)
```
SergioEstevao: Loved this little touch!

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
```diff
@@ -234,6 +234,24 @@ public class ElementNode: Node {
     }
 
 
+    /// Returns the ElementNode instance, whenever there's a *single* children, of the specified nodeType (or nil otherwise).
+    ///
+    /// - Parameter type: Type of the 'single children' node to be retrieved.
+    ///
+    /// - Returns: the requested children (if it's the only children in the collection, and if the type matches), or nil otherwise.
+    ///
+    func onlyChild(ofType type: StandardElementType) -> ElementNode? {
```
diegoreymendez: In order to aim for single-responsibility would you mind splitting this method into two steps (retrieval, filtering)?

This'd help ensure our methods are easier to compose, and also easier to read and understand.

Example:

```swift

/// This one returns first child found, if it's the only one.
///
func onlyChild() -> ElementNode? {
    guard children.count == 1 else {
        return nil
    }

    return children.first as? ElementNode
}

/// This one returns the first child found, if it's the only one and it matches the specified type.
///
func onlyChild(ofType type: StandardElementType) -> ElementNode? {
    guard let child = onlyChild(),
        child.isNodeType(type) else {
            return nil
    }

    return child
}
```

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
```diff
@@ -234,6 +234,24 @@ public class ElementNode: Node {
...
+        guard children.count == 1,
+            let onlyChild = children.first as? ElementNode,
+            onlyChild.isNodeType(type)
+        else {
```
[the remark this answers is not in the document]
jleandroperez: Our guidelines don't fully apply to this statement, because the `guard` clause has multiple lines / conditions.

Moving the else to the preceding line, would have the sad effect of forcing us to further increase the indentation level of the snippet that goes inside the `else`.

IMO this way we truly improve readability, and avoid increasing indentation. 
I'll submit a formal request to add this to our guideline!

AztecTests/HTML/Nodes/ElementNodeTests.swift
```diff
@@ -37,4 +38,34 @@ class ElementNodeTests: XCTestCase {
         XCTAssert(style1 == style2)
         XCTAssert(style1 !== style2)
     }
+
+
+    /// Verifies that `onlyChildr` returns the receiver's only child, if it's type matches with the specified one.
+    ///
+    func testOnlyChildReturnsSingleChildrenIfItRepresentsAnImage() {
```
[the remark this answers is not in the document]
jleandroperez: Thank you sir!

Aztec/Classes/Extensions/NSAttributedString+Attachments.swift
jleandroperez: I absolutely LOVE 'extensible enums' (that's this trick's name!)

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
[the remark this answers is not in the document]
diegoreymendez: I still think this looks very inconsistent with how we format the rest of our code, as we don't separate `else` statement from precedint statements anywhere else.

The fact that auto-indent adds an extra tab should not affect how we format the code.

That said, as this won't affect the logic in the code, we can merge as it is.

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
```diff
@@ -234,6 +234,32 @@ public class ElementNode: Node {
...
+
+        return children.first as? ElementNode
+    }
+
+
+    /// Returns the child ElementNode of the specified nodeType -whenever there's a *single* child-, or nil otherwise.
+    ///
+    /// - Parameter type: Type of the 'single child' node to be retrieved.
+    ///
+    /// - Returns: the requested children (if it's the only children in the collection, and if the type matches), or nil otherwise.
```
diegoreymendez: /s/children/child

jleandroperez:
Thank you both!

PR #847, opened 2017-12-04
Try/refactor media attachment

SergioEstevao:
This PR changes the way attributes for MediaAttachment are implemented, instead of having custom properties directly implemented in the ImageAttachment and VideoAttachment, all the properties are stored and accessed through the use of `extraAttributes` property, and are calculated variables implemented in extensions.

In my point of view this has the following advantages:

- Simpler parser and serialiser, because we don't need to do extra steps for the specific properties
- Allows expansibility of properties by the users of the library.
- Still allows to create specific enums for properties to make it easier to use.

To test:
 - Check if the project compile and unit test are ok.
 - Start the app, check interactions with image and video are working ok.

### DIFF
--- a/Aztec/Classes/Formatters/Implementations/ImageFormatter.swift
+++ b/Aztec/Classes/Formatters/Implementations/ImageFormatter.swift
@@ -23,14 +23,13 @@ class ImageFormatter: StandardAttributeFormatter {
 
                 let url: URL?
 
-                if let urlString = element.attribute(named: "src")?.value.toString() {
-                    extraAttributes.removeValue(forKey: "src")
-                    url = URL(string: urlString)
+                if let srcString = element.attribute(named: "src")?.value.toString() {
+                    url = URL(string: srcString)
                 } else {
                     url = nil
                 }
 
-                let attachment = ImageAttachment(identifier: UUID().uuidString, url: url)
+                let attachment = ImageAttachment(identifier: UUID().uuidString, src: url)
                 attachment.extraAttributes = extraAttributes
                 attributeValue = attachment
             default:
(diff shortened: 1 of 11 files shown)


jklausa:
I really like the simplification and moving the responsibility here! 

My only nitpick would be that the `size` and `alignment` setters/getters are almost identical — maybe it would be possible to extract them into their own thing?

SergioEstevao:
@jklausa good point I think I can introduce an helper method to allow changes on the html class attribute.

diegoreymendez:
In this proposal, it seems like we're simplifying the converters, but what we're really doing is moving that complexity to the [setters and getters](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/847/files#diff-d8a133f794ed76fc0e76e3f633a1cf51R160).

That's much more complexity that I'd like to see executed each time a setter and / or getter gets called.

Doing [part of the HTML conversion](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/847/files#diff-d8a133f794ed76fc0e76e3f633a1cf51R111) inside of getters and setters in attachments is breaking the design responsibility of converters taking care of that in a single step.

We do need to simplify our formatters but I don't see this being the right direction.

We should instead make the conversion step customizable.

SergioEstevao:
@diegoreymendez keep in mind that the `class` attribute is one of the more complex attributes in html and as @jklausa says I can simplify the code to have a single method to set class attributes.

One thing that I don't like on the current approach is that attributes like size, that in the current format are specific to WordPress are part of the Aztec implementation. This implementation will allow to move it out completely of Aztec pod.

diegoreymendez:
@SergioEstevao - Mind if we close this for now?

diegoreymendez:
Closing.

PR #848, opened 2017-12-05
Figure + Figcaption Support Mark I

jleandroperez:
### Details:
This PR implements a simple mechanism in which the following input:

```
<figure>
<img src=".">
<figcaption><b>Some styled text!</b></figcaption>
</figure>
```

Would get mapped into an **ImageAttachment** instance, with it's `caption` field set to `Some styled text!` (along with it's styles, right?).

**figure** and **figcaption** are expected not to be stored as unknownHTML anymore.

This PR (if approved!) will be followed up with a second one, in charge of rendering captions onscreen.

Ref. #723

### To test:
1. Verity that the unit tests are good!
2. Try adding `<figure><figcaption/></figure>` snippet and converting back and forth.

cc @diegoreymendez 
Thanks!

PR #849, opened 2017-12-05
Media formatters refactor.

### LINKED ISSUE #841: Redesign ImageFormatter and VideoFormatter.
ImageFormatter and VideoFormatter don't fit well as `StandardAttributeFormatter` subclasses.

We should evaluate how to split these classes from the rest of the formatters, and introduce a new concept behind them (probably something along the lines of `ImageInserter` and `VideoInserter`).

jklausa:
This is way way way late, but after banging my head against Aztec for a week and then being sick, I finally managed to get something working. 

This cleans up how Image/Video Formatters work, by splitting them into their own thing — ElementConverters, since, well, they don't really format anything. 

Fixes #841, will need to updated after #847 lands.

To test:
Check if the project compile and unit test are ok.
Start the app, check interactions with image and video are working ok.


Aztec/Classes/Inserters/Implementations/ImageInserter.swift
```diff
@@ -0,0 +1,56 @@
+import UIKit
+
+class ImageInserter: MediaInserter {
+
+    func attachment(from representation: HTMLRepresentation) -> MediaAttachment? {
+        guard case let .element(element) = representation.kind else {
+            return nil
```
jklausa: This used to return `ImageAttachment(identifier: UUID().uuidString)`, whereas the `VideoFormatter` just threw up it's hands and `assert()`ed. 

I _think_ just returning `nil` here and doing nothing is safe, but I'd appreciate someone double-checking my logic.

SergioEstevao:
@diegoreymendez and @jklausa while I understand the idea behind this change name, looking at the code has it stands I think it actually makes sense to be called a formatter. The way it's being used we are formatting the attachment character with an attachment. We are not inserting anything on the string.

The Inserter name more sense when we were using the ImageFormatter and VideoFormatter for the toolbar actions, but looking at the current code we are no longer using them anymore for toolbar actions.

Unless I'm missing something I don't think this name change makes the code better.

diegoreymendez:
@SergioEstevao - The change will go deeper.  I've just provided some feedback to @jklausa  about how I feel we can make the distinction clearer as well as the responsibilities of this class.

Stay tuned for updates.

Aztec/Classes/Inserters/Base/Inserter.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Returns an attachment representing the HTML element.
+    ///
+    /// - Parameters:
+    ///
+    ///     - representation: HTML element to turn into an attachment.
+    ///     - inheritedAttributes: Attributes inherited from the parent element — to be turned into `extraAttributes` on the `Attachment` itself.
+    ///
+    /// - Returns: Attachment when appropriate, `nil` when there isn't a valid transformation into an Attachment.
+    func attachment(from element: ElementNode, inheriting inheritedAttributes: [AttributedStringKey: Any]) -> NSTextAttachment?
```
jleandroperez: I wonder if the result type could be a generic... `<T: NSTextAttachment>`, so that the classess that implement this may customize the result type, and we get away without casting all the things.

diegoreymendez: Two observations... 

1. This is starting to look more and more like a `Converter` than an `Inserter`.  It's not important right now, but we may need to revisit the naming (start thinking!).

2. @jleandroperez you may be right.

jklausa: Doesn't any type information get lost as soon as we actually attach it to the `NSAttributedString`? 

Attributes are typed as `[AttributedStringKey: Any]`, so even if the `Inserter`s returned something more specific than `NSTextAttachment`, the only place where this would be useful is after we created an attachment, but haven't attached it to a string yet — does that buy us anything useful?

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -105,8 +105,9 @@ class AttributedStringSerializer {
         let childAttributes = self.attributes(for: element, inheriting: attributes)
         let content = NSMutableAttributedString()
 
-        if let representation = implicitRepresentation(for: element, inheriting: childAttributes) {
-            content.append(representation)
+        if let inserter = inserter(for: element) {
+            let insertedString = inserter.insert(for: element, inheritedAttributes: childAttributes)
```
diegoreymendez: This is a great line to try and bring up something that feels off to me.

If you read the line as if it was english text, you'll notice that the "inserter" name is not very appropriate for the new entity we're creating.  Mostly because no insertion takes place at all here, but rather a "conversion" or "construction" of a string, based on an element.

I think we should consider other naming options that better represent what's happening here.

I know it's mostly on my for pushing the concept of a new `Inserter` entity, so please bear with me @jklausa.

My best suggestion right now would be something like `Converter`... which would make this code become:

```swift
if let converter = converter(for: element) {
    let string = converter.convert(element, inheritingAttributes: chldAttributes)
} else {
   // etc
}
```

jklausa:
Alright, I think this is in a good shape now. I updated the unit test that was failing, since I think the output is now improved. I'm still not 100% sure _why_ it's changed now, but all other tests are 👍

Aztec/Classes/ElementConverters/Implementations/BRElementConverter.swift
```diff
@@ -0,0 +1,7 @@
+import UIKit
+
+class BRElementConverter: ElementConverter {
```
jleandroperez: (Nipticky): We're using (nearly all over!) something like this:

```
// MARK: - FormatBar
//
open class FormatBar: UIView
```

And in few other spots, something like this:

```
/// Parses an attributed string into an HTML tree.
///
class AttributedStringParser {
```

I think we haven't really settled on anything there, but maybe (maybe) pick one of them two, and add it to the new classes, just for consistency?

Aztec/Classes/ElementConverters/Base/ElementConverter.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    /// Default implementation, element converters providing attachments should override this.
+    func attachment(from representation: HTMLRepresentation, inheriting inheritedAttributes: [AttributedStringKey: Any]) -> NSTextAttachment? {
+        return nil
+    }
+
+    func extraAttributes(for representation: HTMLRepresentation) -> [AttributedStringKey: Any]? {
+        return nil
+    }
+
```
jleandroperez: Extra newline

Aztec/Classes/ElementConverters/Implementations/ImageElementConverter.swift
```diff
@@ -0,0 +1,61 @@
...
+        
+        attachment.extraAttributes = extraAttributes
+
+        return attachment
+    }
+
+    func specialString(for element: ElementNode) -> String {
+        return String(UnicodeScalar(NSAttachmentCharacter)!)
+    }
+
```
jleandroperez: Extra newline 😄 (sorry, i'm annoying!!)

AztecTests/NSAttributedString/Conversions/AttributedStringSerializerTests.swift
```diff
@@ -118,7 +118,7 @@ class AttributedStringSerializerTests: XCTestCase {
     ///
     func testLineBreakTagWithinUnsupportedHTMLDoesNotCauseDataLoss() {
         let inHtml = "<span><br>Aztec, don't forget me!</span>"
-        let expectedHtml = "<p><span><br></span><span>Aztec, don't forget me!</span></p>"
+        let expectedHtml = "<p><span><br>Aztec, don't forget me!</span></p>"
```
jleandroperez: Yup. This looks SO MUCH Better!!

Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
```diff
@@ -356,55 +371,35 @@ extension AttributedStringSerializer {
...
 
-        switch elementType {
-        case .hr, .img, .video:
-            return NSAttributedString(string: .textAttachment, attributes: attributes)
-        case .br:
-            return NSAttributedString(.lineSeparator, attributes: attributes)
-        default:
-            return nil
+        for (key, converter) in elementConvertersMap {
+            if equivalentNames.contains(key.rawValue) {
```
jleandroperez: I really really like what you did here. Even considering the Naming equivalences!

jklausa:
@jleandroperez addressed! 

do we want to get #848 merged first, or this?

diegoreymendez:
@jklausa - I'm merging this one in first, as this provides a fundamental change that I'd like other PRs to be aware of.

diegoreymendez:
Remember to assign the milestone (to both the PR and issue) when a PR is closed.

PR #858, opened 2017-12-13
Basic Dynamic Type support

### LINKED ISSUE #478: Accessibility: support Larger Text.
You can enable “Larger Text” in iOS by going to Settings->General->Accessibility->Larger Text.

We must support this in Aztec.

jklausa:
This PR adds basic support for Dynamic Type — using the new API introduced in iOS 11. 
Supporting this without using iOS 11 specific API is a rather large pain in the butt — there isn't a good way to programatically get correct metrics for arbitrary font — so we'd have to hardcode scaling factors for the system fonts and have clients provide those too, if they want to specify a custom font.

I'm not sure if clients should be able to opt out of this behavior by setting some property — I'd say being accessible by default is the correct thing to do, but that's open to discussion.

Fixes #478.

Before:
simulator screen shot - iphone 8 plus - 2017-12-13 at 12 11 48

simulator screen shot - iphone 8 plus - 2017-12-13 at 12 11 55

After (with the biggest Accesiblity size enabled): 

simulator screen shot - iphone 8 plus - 2017-12-13 at 12 09 50
simulator screen shot - iphone 8 plus - 2017-12-13 at 12 09 39

### DIFF
--- a/Aztec/Classes/Formatters/Implementations/PreFormatter.swift
+++ b/Aztec/Classes/Formatters/Implementations/PreFormatter.swift
@@ -18,7 +18,15 @@ open class PreFormatter: ParagraphAttributeFormatter {
     /// Designated Initializer
     ///
     init(monospaceFont: UIFont = UIFont(descriptor:UIFontDescriptor(name: "Courier", size: 12), size:12), placeholderAttributes: [AttributedStringKey : Any]? = nil) {
-        self.monospaceFont = monospaceFont
+        let font: UIFont
+
+        if #available(iOS 11.0, *) {
+            font = UIFontMetrics.default.scaledFont(for: monospaceFont)
+        } else {
+            font = monospaceFont
+        }
+
+        self.monospaceFont = font
         self.placeholderAttributes = placeholderAttributes
     }
 
--- a/Aztec/Classes/TextKit/ParagraphProperty/Header.swift
+++ b/Aztec/Classes/TextKit/ParagraphProperty/Header.swift
@@ -32,7 +32,13 @@ open class Header: ParagraphProperty {
         }()
 
         public var fontSize: Float {
-            return HeaderType.fontSizeMap[self] ?? Constants.defaultFontSize
+            let fontSize = HeaderType.fontSizeMap[self] ?? Constants.defaultFontSize
+
+            if #available(iOS 11.0, *) {
+                return Float(UIFontMetrics.default.scaledValue(for: CGFloat(fontSize)))
+            } else {
+                return fontSize
+            }
         }
     }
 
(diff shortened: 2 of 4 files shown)


diegoreymendez:
I didn't manage to get it working with the following steps:

**Test 1:**

1. Launch the editor
2. Open the empty editor
3. Type "Test"
4. Go to settings, enable Large text.
5. Go back to the app.

The text should be large, but is still small.

**Test 2:**

1. Launch the editor
2. Go to settings, enable Large text.
3. Go back to the editor, tap on "Empty editor"
4. Type something.

The text should be large, but is still small.

**Screenshot:**

screen shot 2017-12-13 at 10 01 21

jklausa:
Ugh, sorry for that, should be fixed now. 

There's one edge-case now: the size adjustments *won't* apply to `Header`'s, `<pre>`, lists, etc until you go back and reload the text view — basically something equivalent to listening to a notification and doing `setHTML(getHTML())`.

Supporting this would require us to go through entirety of the text in the TextView and reapplying attributes to match the new text size environment — for, I'd argue, little gain — I imagine this isn't a setting people change frequently, it mostly just set-once-and-forget thing?

jklausa:
The test failure seem to be unrelated to this, I think it's the flaky UI tests :(

diegoreymendez:
@jklausa - It's working better now... however the HTLM mode editor is not affected by large text.  We should either include that change in this PR, or modify the description so that #478 is not closed automatically by this PR.

diegoreymendez:
Additionally, I just checked one of the UI tests (`testLongTitle()`), and it seems it'll need to be modified to suit the changes in this PR.

Do check the others as well.

jklausa:
Ugh, sorry for forgetting the HTML view. I'm still not super intimate with all the little nooks and crannies of Aztec (though, I think, the HTML view is not part of Aztec? It's "just" a separate `UITextView`.) 

I updated the demo app though.

I looked at the tests and few are indeed failing with some of the larger accesibility sizes enabled in Simulator. 

I'm not sure how to actually test those scenarios though in CI environement — we'd have to basically do a second pass of all the UI tests, passing some env variables to `XCUIApplication.launchArguments` — and basically doubling the runtime of the tests, to ~an hour.

brbrr:
Travis has some limited support for running the UI tests, and it didn't provide much feedback on failures. so maybe it worth to put UI tests into a separate scheme and run them only on buddybuild?

diegoreymendez:
To be honest as long as the tests don't fail after resetting the device (thus ensuring we're using the defaults) we should be fine.

It's very well possible the tests were failing for me due to that... can you confirm if they work fine using the default settings @jklausa ?

jklausa:
Yup, we don't change the content size for Travis runs, so if they were green there, they pass OOTB. 

I went ahead and rechecked to make sure they do pass locally too using the `Large` setting (fourth smallest).

(as an aside, why is the default size called "Large" and not "Medium" it doesn't make any sense apple, ugh)

PR #863, opened 2017-12-18
Make testLongTitle UI test font size agnostic

brbrr:
This test was relying on a hardcoded textfield height which is changing with different font sizes. Now it passing on all font sizes configurations.


jklausa:
I think this is still failing if you go to Settings -> Accessibility -> Larger Text -> Enable Larger Sizes and pick one of the biggest one — it'll force the text to overflow into more than 2 lines and fail: 
simulator screen shot - iphone 8 plus - 2017-12-19 at 02 02 55

brbrr:
Nice catch @jklausa! I've update test to make sure it will accept 2+ lines titles. But it still might be some other edge cases where this test could fail.

diegoreymendez:
@brbrr, @jklausa - Nice work.  I think as long as we can make this more resilient it'll be good.

If there's any edge case we cannot solve, just document it here for history.

jklausa:
👍

PR #869, opened 2017-12-25
Reverting a change that was made in commit d8eb7b8 fixed the crash. T…

shahla00:
The app was getting into an infinite loop as both FigureElementConverter and AttributedStringSerializer were initializing each other.

Fixes #

To test:


jleandroperez:
Apologies @shahla00, absolutely missed this PR!! 

Closing since it's been fixed via #874

PR #870, opened 2017-12-25
Sample App: Single Title Line

### LINKED ISSUE #821: Example app title should not allow multiple lines.
Our Example App's title field should not allow multiple lines of text.

shahla00:
### Details:
Limit the title to one line by using UITextField instead of UITextView. Also there is no need to have a separate UILabel for the placeholder text. We can make use of UITextField’s placeholder attribute.

Fixes #821

### To test:
1. Launch the Sample App
2. Verify that the Title Field is not multiline

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -92,30 +92,17 @@ class EditorDemoController: UIViewController {
         return textView
     }()
 
-    fileprivate(set) lazy var titleTextField: UITextView = {        
-        let textField = UITextView()
-
+    fileprivate(set) lazy var titleTextField: UITextField = {
+        let textField = UITextField()
+        textField.placeholder = NSLocalizedString("Enter title here", comment: "Label for the title of the post field. Should be the same as WP core.")
+        
         textField.accessibilityLabel = NSLocalizedString("Title", comment: "Post title")
         textField.delegate = self
         textField.font = UIFont.preferredFont(forTextStyle: UIFontTextStyle.headline)
         textField.returnKeyType = .next
-        textField.textColor = UIColor.darkText
+        textField.textColor = .darkText
         textField.translatesAutoresizingMaskIntoConstraints = false
-        textField.isScrollEnabled = false
         textField.backgroundColor = .clear
-
-        return textField
-    }()
-
-    fileprivate(set) lazy var titlePlaceholderLabel: UILabel = {
-        let placeholderText = NSLocalizedString("Enter title here", comment: "Label for the title of the post field. Should be the same as WP core.")
-        let textField = UILabel()
-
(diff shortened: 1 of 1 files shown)


Example/Example/EditorDemoController.swift
```diff
@@ -244,29 +230,12 @@ class EditorDemoController: UIViewController {
     // MARK: - Title and Title placeholder position methods
     func updateTitlePosition() {
         let referenceView: UIScrollView = editingMode == .richText ? richTextView : htmlTextView
-        titleTopConstraint.constant = -(referenceView.contentOffset.y+referenceView.contentInset.top)
+        titleTopConstraint.constant = -(referenceView.contentOffset.y + referenceView.contentInset.top) + 5
```
jleandroperez: Mind moving the magic numbers to constants? perhaps something like this would become handy (in the `struct Constants`):

```
let titleInsets = UIEdgeInsets(top: 5, left: 0, bottom: 5, right: 0)
```

Example/Example/EditorDemoController.swift
```diff
@@ -281,24 +250,17 @@ class EditorDemoController: UIViewController {
     private func configureConstraints() {
 
         titleHeightConstraint = titleTextField.heightAnchor.constraint(equalToConstant: titleTextField.font!.lineHeight)
-        titleTopConstraint = titleTextField.topAnchor.constraint(equalTo: view.topAnchor, constant: -richTextView.contentOffset.y)
-        updateTitleHeight()
+        titleTopConstraint = titleTextField.topAnchor.constraint(equalTo: view.topAnchor, constant: 5)
```
jleandroperez: Same as above, mind replacing the magic number with a constant?

Example/Example/EditorDemoController.swift
```diff
@@ -244,29 +230,12 @@ class EditorDemoController: UIViewController {
     // MARK: - Title and Title placeholder position methods
     func updateTitlePosition() {
         let referenceView: UIScrollView = editingMode == .richText ? richTextView : htmlTextView
-        titleTopConstraint.constant = -(referenceView.contentOffset.y+referenceView.contentInset.top)
+        titleTopConstraint.constant = -(referenceView.contentOffset.y + referenceView.contentInset.top) + 5
 
         var contentInset = referenceView.contentInset
-        contentInset.top = (titleHeightConstraint.constant + separatorView.frame.height)
+        contentInset.top = titleHeightConstraint.constant - 10
```
jleandroperez: Mind replacing the `-10` with, perhaps... `(Constants.titleInsets.top + Constants.titleInsets.bottom)` (or so!)

jleandroperez:
### Note(s):
+ Moved the issue title to it's body (because it was getting cut off) 
+ Added a shorter title 
+ Referenced the main issue
+ Added brief testing steps

PR #876, opened 2017-12-26
Fix a memory leak issue

shahla00:
Fixes #

To test:

### DIFF
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -11,11 +11,11 @@ class EditorDemoController: UIViewController {
 
     fileprivate var mediaErrorMode = false
 
-    fileprivate(set) lazy var formatBar: Aztec.FormatBar = { [unowned self] in
+    fileprivate(set) lazy var formatBar: Aztec.FormatBar = {
         return self.createToolbar()
     }()
 
-    fileprivate(set) lazy var richTextView: Aztec.TextView = { [weak self] in
+    fileprivate(set) lazy var richTextView: Aztec.TextView = {
 
         let paragraphStyle = Aztec.ParagraphStyle.default
         
@@ -53,7 +53,7 @@ class EditorDemoController: UIViewController {
         return textView
     }()
 
-    fileprivate(set) lazy var htmlTextView: UITextView = { [weak self] in
+    fileprivate(set) lazy var htmlTextView: UITextView = { in
         let defaultFont: UIFont
 
         if #available(iOS 11, *) {
--- a/Example/Example/EditorDemoController.swift
+++ b/Example/Example/EditorDemoController.swift
@@ -11,11 +11,11 @@ class EditorDemoController: UIViewController {
 
     fileprivate var mediaErrorMode = false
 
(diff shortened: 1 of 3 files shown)


aerych:
@jleandroperez could you take a peek at this PR when you're back?

shahla00:
These 3 PRs resolved this memory leak issue.

Example/Example/EditorDemoController.swift
```diff
@@ -1370,8 +1370,8 @@ private extension EditorDemoController
            detailsViewController.linkURL = url
         }
         
-        detailsViewController.onUpdate = { (alignment, size, url, linkURL, alt) in
-            self.richTextView.edit(attachment) { updated in
+        detailsViewController.onUpdate = { [weak self] (alignment, size, url, linkURL, alt) in
```
jleandroperez: Would it be possible to add this snippet at the first line, inside the **onUpdate** callback?

```
guard let `self` = self else {
   return
}
```

Benefits are:

1. You don't run all of the callback's code (if) self is released
2. By means of the `` wrapping trick, there is no need to use a keyword such as `strongSelf = self`
3. And of course, you wouldn't need to use `?` in any of the callback's lines
  
This makes sense when the callback is more than one line long. Personally, whenever the callback has a single line, and it involves a `weak self` reference, i go with the style you've used in here.

Example/Example/EditorDemoController.swift
```diff
@@ -11,11 +11,11 @@ class EditorDemoController: UIViewController {
 
     fileprivate var mediaErrorMode = false
 
-    fileprivate(set) lazy var formatBar: Aztec.FormatBar = {
+    fileprivate(set) lazy var formatBar: Aztec.FormatBar = { [unowned self] in
```
jleandroperez: Mind reverting this one?. Lazy var's closure is @noescape, and thus, does not retain self.

[Ref. Here](https://stackoverflow.com/questions/38141298/lazy-initialisation-and-retain-cycle)

Example/Example/EditorDemoController.swift
```diff
@@ -11,11 +11,11 @@ class EditorDemoController: UIViewController {
...
         return self.createToolbar()
     }()
 
-    fileprivate(set) lazy var richTextView: Aztec.TextView = {
+    fileprivate(set) lazy var richTextView: Aztec.TextView = { [weak self] in
```
jleandroperez: Same as above. Mind reverting this one? (since it's not needed)

Example/Example/EditorDemoController.swift
```diff
@@ -53,7 +53,7 @@ class EditorDemoController: UIViewController {
         return textView
     }()
 
-    fileprivate(set) lazy var htmlTextView: UITextView = {
+    fileprivate(set) lazy var htmlTextView: UITextView = { [weak self] in
```
jleandroperez: Same as above, not needed. Please revert 😄

shahla00:
@jleandroperez I incorporated the requested changes. Please let me know if I missed anything. Thanks for taking the time to review and provide feedback.

PR #884, opened 2018-01-02
ImageAttachment: Explicit Width and Height

### LINKED ISSUE #883: Images: Explicit Width / Height
### Details:
WordPress wp-admin's post editor isn't dynamically resizing the assets, based on the current CSS class. As a result, [we can observe the behavior outlined here.](https://github.com/wordpress-mobile/WordPress-iOS/issues/8384#issuecomment-354154635).

Let's explicitly serialize the **width** and **height** `` fields: this (should) improve browser's rendering time, and should also fix [WPiOS's #8384](https://github.com/wordpress-mobile/WordPress-iOS/issues/8384).

Note that this must be done in Aztec itself because:
- The serialization spot is not open
- It's standard HTML, not tied up to WP itself

jleandroperez:
### Description:
In this PR we're enhancing the AttributedStringParser, so that ImageAttachments get their **width** and **height** properly serialized, whenever the size is set to (anything) but `.none / .full`.

Needs Review: @diegoreymendez / @SergioEstevao 
Closes #883

### Details:
Motivation for this PR is [this WPiOS issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/8384#issuecomment-354154635).

**TL;DR version:** `WordPress.com/wp-admin`'s post editor doesn't pick up the size's CSS classes. Instead, it relies on explicit width / height fields, in order to render images in their 'actual final size'.

In this PR we're explicitly setting the image's size, so that wp-admin behaves in a WYSIWYG fashion.

### To test:
1. Run the unit tests. Verify they're green!
2. Add an attachment. Change it's size to `None / Full`, switch to HTML, and verify no width or height attribute shows up.
3. Change the attachment's size to `Thumbnail / Medium / Large`. Verify that the HTML produced, for each image size, contains explicit `width` and `height` fields.

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -910,6 +910,10 @@ private extension AttributedStringParser {
             imageElement.updateAttribute(named: attribute.name, value: attribute.value)
         }
 
+        for attribute in imageSizeAttributes(from: attachment) {
+            imageElement.updateAttribute(named: attribute.name, value: attribute.value)
+        }
+
         for (key,value) in attachment.extraAttributes {
             var finalValue = value
             if key == "class", let baseValue = imageElement.stringValueForAttribute(named: "class"){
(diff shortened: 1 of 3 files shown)


jleandroperez:
Thanks for the review sir, much appreciated!!.

Working on the broken UI Tests in #887 (should be patched up soon!).

PR #889, opened 2018-01-15
FormatBar: Fix vertical position after updating for safe area insets

frosty:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/8413.

I apologise that the fix is a `dispatch async`, but I tried a number of things and couldn't come up with another way to fix this. Essentially, we resize the toolbar as necessary when the safe area insets change, but for some reason in the situation described in the original ticket the system doesn't adjust the position of the bar to match its new height unless we dispatch it.

**To test:**

* Follow the steps in the original issue, and check that the format bar appears in the correct vertical position.

### DIFF
--- a/Aztec/Classes/GUI/FormatBar/FormatBar.swift
+++ b/Aztec/Classes/GUI/FormatBar/FormatBar.swift
@@ -443,7 +443,10 @@ open class FormatBar: UIView {
             heightConstraint?.constant = newHeight
             updateDividerInsets()
             invalidateIntrinsicContentSize()
-            layoutIfNeeded()
+
+            DispatchQueue.main.async {
+                self.layoutIfNeeded()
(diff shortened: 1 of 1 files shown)


Aztec/Classes/GUI/FormatBar/FormatBar.swift
```diff
@@ -443,7 +443,10 @@ open class FormatBar: UIView {
             heightConstraint?.constant = newHeight
             updateDividerInsets()
             invalidateIntrinsicContentSize()
-            layoutIfNeeded()
+
+            DispatchQueue.main.async {
```
nheagy: I trust that this `async` is necessary 😀 

What do you think about adding a comment to explain why it's there?

Issue #893, opened 2018-01-22
Captions + Links not working correctly.


diegoreymendez:
## Description

When loading WordPress captions with links, the Example App is failing to parse them correctly.

## Reproduction

Simply insert this (valid caption snippet) into HTML mode and switch back to visual:

```html
[caption]<a href="www.wordpress.com"><img src="image.jpg" height="90" width="135"></a>Some text[/caption]
```

**Expected**

The caption should be parsed correctly into a single image in visual mode.

**Actual**

Some of the caption text is shown as regular text.

diegoreymendez:
Final solution coming up in #910.

diegoreymendez:
Fixed by #910.

PR #895, opened 2018-01-25
Captions + links: option A.

diegoreymendez:
This is option A of a possible direction to resolve [captions + links](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/893).

#### Fixes:
- Allows `<figure>` elements with links.

#### Problems:
- `<figure>` elements are always wrapped inside links, since they're created as part of our leaf nodes.  This happens because we don't treat `<figure>` elements as block-level (ie: as paragraph styles).

Another PR will be opened offering a different solution to compare.

Fyi, @SergioEstevao


diegoreymendez:
Closing this in favor of https://github.com/wordpress-mobile/AztecEditor-iOS/pull/910

PR #898, opened 2018-01-30
A bad condition was causing eventual crashes when using header styles.

diegoreymendez:
Title has it.

A hotfix for beta 17 will be released once this is in.

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -508,8 +508,9 @@ private extension AttributedStringParser {
             }
         }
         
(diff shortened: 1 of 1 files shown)


Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
```diff
@@ -508,8 +508,9 @@ private extension AttributedStringParser {
             }
         }
         
-        let lastElement = paragraphNodes.last! // There's no scenario where having zero paragraph nodes would make sense.
```
SergioEstevao: Loved the comment that was there before :).

diegoreymendez: Totally!

diegoreymendez:
Thanks both @elibud and @SergioEstevao !!

PR #903, opened 2018-02-05
Feature/452 add support for code element

### LINKED ISSUE #452: Add support for code html tag.
Add support for the `<code>` tag.

Talk with the Android team about this to request their current implementation's details (including toolbar support).

SergioEstevao:
Fixes #452 

This PR adds support for the [`<code>`](https://www.w3.org/TR/html5/textlevel-semantics.html#the-code-element) element.

While the majority of the code in this PR is pretty standard: adding a new character formatter with styling for the attribute (monospace font, background color), parsing and serialisation; There is something different on this attribute, the style it applies is not unique to this element.

So to implement we are using a custom attribute, .codeHTMLRepresentation that uses an HTMLRepresentation.

In order to make it work while editing, it need an addition to ```insertText``` method on the TextView to support the copy of custom attributes while writing text inside a code element.

At the moment the code is only making sure that the new .codeHTMLRepresentation is being copied but this could be expanded for new custom attributes.

To test:
 - Start the demo app
 - Open the demo with content
 - Check if the code attributes are being parsed ok
 - Write some text inside the code element
 - Check if style is still correct
 - Switch to HTML mode and check that the HTML is correctly generated.


Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
```diff
@@ -777,6 +781,16 @@ private extension AttributedStringParser {
         return ElementNode(type: .u)
     }
 
+    /// Extracts all of the Underline Elements contained within a collection of Attributes.
+    ///
+    private func processCodeStyle(in attributes: [AttributedStringKey: Any]) -> ElementNode? {
+        guard let _ = attributes[.codeHtmlRepresentation] as? HTMLRepresentation else {
```
diegoreymendez: Can be simplified as:

```swift
guard attributes[.codeHTMLRepresentation] is HTMLRepresentation else {
```

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -1058,6 +1060,22 @@ open class TextView: UITextView {
         typingAttributesSwifted.removeValue(forKey: .link)
     }
 
+    /// This method makes sure that the Custom Code HTML attribute is copy across to the next character that is typed on the textview.
+    ///
+    /// - Parameter range: the range where the new text will be inserted
+    ///
+    private func ensureCopyOfCodeCustomTypingAttributes(at range: NSRange) {
+        guard typingAttributesSwifted[AttributedStringKey.codeHtmlRepresentation] == nil else {
```
diegoreymendez: Can be joined with the next guard.

SergioEstevao:
@diegoreymendez updated the code, ready for another look.

Aztec/Classes/Extensions/NSAttributedString+Analyzers.swift
```diff
@@ -6,6 +6,22 @@ import UIKit
...
+    /// Returns true if the text preceding a given location contains the specified attribute key.
+    ///
+    /// - Parameters:
+    ///   - location: location to check
+    ///   - key: the attributed to check
+    /// - Returns: True if the attribute is there
+    ///
+    func isLocation(_ location: Int, preceededBy key: NSAttributedStringKey) -> Bool {
+        let beforeRange = NSRange(location: location - 1, length: 1)
+        guard beforeRange.location >= 0 else {
```
diegoreymendez: We could guard against `location == 0` on top of the method, since there's no need to create the range in this scenario.

Aztec/Classes/Formatters/Implementations/CodeFormatter.swift
```diff
@@ -0,0 +1,70 @@
...
+            resultingAttributes[.font] = placeholderAttributes[.font]
+            resultingAttributes[.backgroundColor] = placeholderAttributes[.backgroundColor]
+        }
+
+        return resultingAttributes
+    }
+
+    func present(in attributes: [NSAttributedStringKey: Any]) -> Bool {
+        if let font = attributes[NSAttributedStringKey.font] as? UIFont {
+            return font == monospaceFont
```
diegoreymendez: This seems a bit risky to know if there's code.  What about using a code attribute instead?

SergioEstevao:
@diegoreymendez I update the code with your suggestions. Can you give it another look.

SergioEstevao:
@diegoreymendez added a button to support toggle of code style.

PR #904, opened 2018-02-05
Fix FormatBar for RTL layout direction

### LINKED ISSUE #902: FormatBar not behaving properly with Right-to-Left layout
### Expected behavior

The `FormatBar` with RTL layout should have a mirror behavior of the LTR layout behavior.

Normal LTR
original_ltr

Expected RTL
fixed_rtl

### Actual behavior

The FormatBar still scrolls from left to right with RTL layout, even when most of its UI has been mirrored correctly. This issue also makes the 'insert' button disappear making it impossible to insert media items.

Actual RTL
original_rtl

Building with Xcode:
- Edit Scheme
- Options
- Application Language
- Arabic (or Right-to-Left Pseudolanguage)
- Run the app
- Create a post and check the bar

On iPhone:
- Set the system language to Arabic
- Run the app
- Create a post and check the bar

etoledom:
Fixes #902 

This PR fixes the layout problem by flipping the scrollView horizontally.

fixed_rtl

### Building with Xcode:

1. Edit Scheme
2. Options
3. Application Language
4. Arabic (or Right-to-Left Pseudolanguage)
5. Run the app
6. Create a post and check the bar

### On device:

1. Set the system language to Arabic
2. Run the app
4. Create a post and check the bar


Aztec/Classes/GUI/FormatBar/FormatBar.swift
```diff
@@ -634,22 +660,17 @@ private extension FormatBar {
         scrollView.alwaysBounceHorizontal = false
         scrollView.translatesAutoresizingMaskIntoConstraints = false
         scrollView.delegate = self
-
+        if layoutDirection == .rightToLeft {
+            scrollView.transform = CGAffineTransform(scaleX: -1, y: 1)
```
frosty: Perhaps add a comment explaining what this does / why it's necessary?

frosty: Also we have very similar code in two places... perhaps a private `UIView` helper method? `scrollView.transformForLayoutDirection` or `flipForLayoutDirection` or something? (with a well-named method, maybe the comment wouldn't be necessary :))

etoledom:
Thanks for the review @frosty !
I made the suggested changes. 👍

etoledom:
Cool! I'll give it a quick look to that last issue

etoledom:
I added an animation block to animate the constraint changes. It seems to work fine but it looks a bit slower than the LTR animation. I didn't want to add another duration constant but if it's necessary I can do it :)

PR #907, opened 2018-02-06
Removes `AttributedStringKey`.

diegoreymendez:
This PR removes `AttributedStringKey` from Aztec.

It may also break Swift 3.2 compatibility making Swift 4 our supported language version.  This is fine, since we have limited resources.

You can test the integration using WPiOS branch `try/aztec-remove-swift32-compatibility`.


SergioEstevao:
I was thinking that a a typedef for this: ```[NSAttributedStringKey: Any]``` could have saved us a lot of work...

PR #908, opened 2018-02-07
Align center not being saved

### LINKED ISSUE #890: Image alignment lost for posts edited with WPiOS
Steps to reproduce:

1. Go to https://wordpress.com/ > My Sites > Blog Posts > Add.
1. Create a post with a centered thumbnail-sized image.
1. Publish and view the post to verify the image looks centered.
1. Open the same post in WPiOS and add any text to the body of the post and save changes.
1. View the post and compare the image alignment for the image in the editor versus the front end.

<strong>Result:</strong> image alignment is lost after editing a post with centered images in WPiOS.

I noticed the class `aligncenter` is missing in WPiOS as soon as the post is opened for editing.

Original post created via the web:

screen shot 2018-01-16 at tue jan 16 2 45 09 pm
<sup>Seen at https://wordpress.com/post/design5279.wordpress.com/11714 using Firefox 57.0.3 on macOS 10.13.2.</sup>

Post as soon as it is opened in WPiOS:

img_1779
<sup>iPhone 6S, iOS 11.2.1, WPiOS 9.1</sup>

##### Tested on iPhone 6S, iOS 11.2.1, WPiOS 9.1

/hat tip @RyanRhoten for raising the issue in #triage on Slack (internal ref: 894502-zen).

SergioEstevao:
Fixes #890

This PR changes the alignment setting on images to be by default `none` it also make sure that all alignments are written in the serialiser.

One thing that come up to my mind is: should we have a value for the alignment enum called "not defined" for cases where no alignment is set on HTML? it will be visually equivalent to none but it will tell us if no value was available  the original HTML and that we don't need to write anything.

To test:
 - Open the editor app on empty demo mode
 - Insert an image
 - Switch to HTML and see if alignm class setting is `alignnone`
 - Switch to visual tap on the image and change alignment to center
 - Switch to HTML mode and see if align class setting is `aligncenter`
 - Do the same for right and left alignments.

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -1030,9 +1030,7 @@ private extension AttributedStringParser {
     ///
     private func imageClassAttribute(from attachment: ImageAttachment) -> Attribute? {
         var style = String()
-        if attachment.alignment != .center {
-            style += attachment.alignment.htmlString()
-        }
+        style += attachment.alignment.htmlString()        
         
         if attachment.size != .none {
             style += style.isEmpty ? String() : String(.space)
--- a/Aztec/Classes/TextKit/ImageAttachment.swift
+++ b/Aztec/Classes/TextKit/ImageAttachment.swift
@@ -16,7 +16,7 @@ open class ImageAttachment: MediaAttachment {
 
     /// Attachment Alignment
     ///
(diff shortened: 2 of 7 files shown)


diegoreymendez:
> One thing that come up to my mind is: should we have a value for the alignment enum called "not defined" for cases where no alignment is set on HTML? it will be visually equivalent to none but it will tell us if no value was available the original HTML and that we don't need to write anything.

I like offering the option, yep.  I'd keep the "None" wording for it though.

PR #912, opened 2018-03-05
TextView: Allow linking text with a pasted URL

frosty:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/6904

This PR detects URLs on the pasteboard during a paste operation, and creates a link with the currently selected text if there is any. This should match the way Calypso works.

aztec-paste

**To test:**

* Copy a URL to the pasteboard – for example, hit the Share button in Safari and copy the link
* Select some text in Aztec and hit paste in the popup menu
* The selected text should become a link
* If you try pasting onto an empty line (no text selected), the existing paste handling code should just insert the URL.
* Ensure pasting other items still works as expected. I found if I tried to copy an image directly out of Safari I couldn't always paste it as I might expect to, but that seems to be a pre-existing issue. Sometimes the URL just inserts instead – I think this is due to `loadAttributedString()` not returning anything.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -409,13 +409,7 @@ open class TextView: UITextView {
     }
 
     open override func paste(_ sender: Any?) {
-        if let urlTypes = UIPasteboardTypeListURL as? [String],
-            UIPasteboard.general.contains(pasteboardTypes: urlTypes),
-            let pastedURL = UIPasteboard.general.url,
-            selectedRange.length > 0 {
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Works great @frosty!

PR is approved.

Just one suggestion: since we have a very crowded method there, it could be a bit nicer to read if we wrapped that logic in a method such as:

```swift
        if tryPastingURL() {
            return
        }
```

frosty:
Thanks @diegoreymendez, and thanks for the great suggestion! I just pushed that change :)

PR #918, opened 2018-03-21
NSParagraphStyle's isEqual now compares correctly against ParagraphStyle.

### LINKED ISSUE #917: Style loss when using NSParagraphStyle objects
### Description

When using `NSParagraphStyle` we're sometimes losing styling information.

### Details

This can be checked by running `testFigcaptionGetsItsStyleProperlyParsed()` several times.  For some reason it doesn't fail every time, but when it does it's because we're losing style information from our `ParagraphStyle` objects.

It seems the culprit is that `NSParagraphStyle.isEqual()` fails to recognize the difference with `Aztec.ParagraphStyle` objects with the same base properties.

diegoreymendez:
### Description

Fixes #917 through method swizzling.

`NSParagraphStyle.isEqual()` now properly compares using `isEqual()` against `ParagraphStyle` objects.

Added two unit tests that trigger the problem.

### Testing

 Run the unit tests multiple times and make sure they never fail.

### DIFF
--- a/Aztec/Classes/TextKit/ParagraphStyle.swift
+++ b/Aztec/Classes/TextKit/ParagraphStyle.swift
@@ -9,7 +9,7 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
     
     /// This is sort of a nasty hack to be able to initialize the class at runtime
     ///
-    private static let initializeClass: () = {
+    static let initializeClass: () = {
         swizzleSuperclass()
     }()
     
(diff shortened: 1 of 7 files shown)


Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -419,4 +429,29 @@ extension ParagraphStyle {
...
+                return
+        }
+        
+        method_exchangeImplementations(originalMethod, swizzledMethod)
+    }
+}
+
+extension NSParagraphStyle {
+    @objc func swizzledIsEqual(_ object: Any?) -> Bool {
+        guard object_getClass(self) == object_getClass(object) else {
```
SergioEstevao: What happens when you compare a NSParagraphStyle with a NSMutableParagraphStyle?

diegoreymendez: Well worst case scenario they're not considered equal (which I don't think is necessarily a bad thing).

That said, I'll try reviewing this logic for that scenario.

diegoreymendez:
@SergioEstevao - Ready for another look.

PR #919, opened 2018-03-22
Use set instead of array for formatting identifiers.

SergioEstevao:
This PR does a small refactor the way we return FormattingIdentifier, instead of using an array that allows duplicates we use a Set.

To test:
 - Open the demo app
 - Move the cursor around and see if the format toolbar updates correctly for the styles


SergioEstevao:
@diegoreymendez develop or 9.7? I just use the commit SHA to refer to the pod?

diegoreymendez:
@SergioEstevao - Don't worry about the integration actually... we'll need to do one big integration soon.

Can you check the unit tests?

SergioEstevao:
@diegoreymendez updated with latest from develop and tests are now passing.

PR #928, opened 2018-04-02
Xcode 9.3 support.

diegoreymendez:
This PR adds Xcode 9.3 support for Aztec.

Make sure all targets build fine and run with Xcode 9.3.

WPiOS PR for testing is [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/9023).


diegoreymendez:
There are conflicts since this PR was branched off the latest that was being used in Xcode.  Once this is approved I'll resolve all conflicts and merge.

jklausa:
I'm wondering about one thing — wouldn't this change mean that Aztec would effectively only support Xcode 9.3 from this point forward? If the `libxml2` is only being supplied as a part of 9.3, this would mean you can't build Aztec with earlier Xcode's, right?

diegoreymendez:
As we discussed privately, you're absolutely right.  Right now our resources are limited though, so at least for Aztec we're only supporting the latest version of Xcode (and all previous and future compatible versions).

Supporting more versions of Xcode will be handled on a case by case scenario, shall the need arise.

diegoreymendez:
Thanks @jleandroperez - I'll fix the conflicts and merge.

PR #942, opened 2018-05-02
Data lost near comments.

### LINKED ISSUE #939: Block Elements (h2...h5, blockquote, etc..) are lost when near comments
### Expected behavior

No element should be lost independently of other elements around it, unless it's a valid merge of elements (two blockquotes)

### Actual behavior

When comments tags are immediately before a block tag the element in front is lost. 

For example if we have the following HTML:
```
<!-- wp:quote --><h2>This is an amazing editor</h2><!-- /wp:quote -->
```

And we pass through it Aztec by just opening and saving we get the following:
```
<p><!-- wp:quote -->This is an amazing editor</p><p><!-- /wp:quote --></p>
```

 - Start a new empty editor
 - Switch  to HTML mode
 - Insert the following: `<p><!-- wp:quote -->This is an amazing editor</p><p><!-- /wp:quote --></p>`
 - Switch to visual and back to HTML
 - See the error

diegoreymendez:
### Description:

Fixes #939.  Comments before block-level elements no longer cause data loss issues.

### Credits:

The initial PR and unit tests were provided by @SergioEstevao.  I reopened the PR here to make sure the review history looks clean.  I also modified the unit tests to only affect what's fixed in this PR.

### Important:

There's a failing unit test that's coming from develop and outside the scope of this PR.  The test is:

`testHeaderStyleDoesNotComeBackFromNonExistanceWheneverDeleteBackwardResultsInEmptyParagraphBeforeHeaderStyle`

### Testing:

Just run the unit tests, and make sure all (except the failing test mentioned above) succeed.


SergioEstevao:
@diegoreymendez just added just some more test around comments to future proof changes in the future.

While I understanding the adding of `<p>` around the comments we probably need to think how to improve this in the future.

PR #952, opened 2018-05-08
Creates the WordPressEditor CocoaPods sub-spec.

diegoreymendez:
### Details

Creates the WordPressEditor CocoaPods sub-spec.

This sub-spec will contain a WordPress-specific implementation of Aztec.  This PR offers an initial proposal that shares the WP-specific processors between the example app and WPiOS, but more classes will be included in the future.

### Testing:

- Run the unit tests.
- Test the demo App and make sure everything's working.
- Try out the WPiOS integration branch: `try/use-wordpresseditor-pod`.


SergioEstevao:
After looking at this, while the code is all good, I now think this should be a sub-project on the main app.

I'm just fearful that changes that need to be done on the editor for specific WordPress changes will envolve creating a new  Aztec pod version or at least new a version of the sub-spec.

What do you think @diegoreymendez ? What are our main goals with this change?

diegoreymendez:
> I'm just fearful that changes that need to be done on the editor for specific WordPress changes will envolve creating a new Aztec pod version or at least new a version of the sub-spec.

When using a subspec the versioning of both is tied, afaik.  You publish one and both are automatically published.  In WPiOS you'd also use the same version for both the spec and subspec (since the subspec imports the Aztec lib).

> What do you think @diegoreymendez ? What are our main goals with this change?

Basically provide us with a mechanism to make sure changes to Aztec are integrated right away into the WP logic, and that we don't have duplicated processors here and in WPiOS (that make us fix bugs twice).

diegoreymendez:
More context: the reason I put the subspec here is that it'd allow us to quickly iterate Aztec changes into WP logic.  Once the WP subspec is mature enough we could create a separate library.

Also, I think `DocumentUploadProcessor.swift` should go back to WPiOS.

Let me know if you have more doubts or other questions @SergioEstevao.

SergioEstevao:
@diegoreymendez ok I think we can start like this and see how it goes.

If DocumentUploadProcessor is going back to WPiOS, ImgUploadProcessor and VideoUploadProcessor should also go back.

diegoreymendez:
> If DocumentUploadProcessor is going back to WPiOS, ImgUploadProcessor and VideoUploadProcessor should also go back.

Yes, I believe so.  Do you agree though?  Could those be necessary for the example App?  Does the code involving those processors need to be shared in the library?

SergioEstevao:
This upload processors are mainly t handle replacement of text on the Post
content when this happens on the background (async post) so I don't think
we need them on the Example app.

diegoreymendez:
@SergioEstevao - I removed the files, which are now in the integration branch in WPiOS.  Shall I move ahead and merge this?

PR #958, opened 2018-05-22
Integrates renderers into Aztec libraries.

diegoreymendez:
### Description:

In this PR:

- We're moving our standard renderers into the `Aztec` and `WordPressEditor` targets.
- We're updating the local version of those renderers, to the latest version from WPiOS.

Next steps will be to integrate this into WPiOS and remove the duplicated renderers from it.  This will also enable us to implement a custom Gutenberg renderer for self-closed blocks.

### Testing:

- Run the unit tests.
- Try the editor out and make sure comments, more-comments and unsupported HTML is rendered correctly.


diegoreymendez:
Hahaha... thanks Jorge!  Miss you too!

PR #967, opened 2018-06-01
Make pre mergeable

### LINKED ISSUE #575: Add PRE tag support.
We currently do not support PRE tags.  Whitespace is relevant inside these.

To test:

1. Go to the HTML editor and paste this.

```
<pre>line 1
line 2
line 3</pre>
```

2. Switch to visual mode and back to HTML.  Make sure all whitspace is unmodified.

SergioEstevao:
Fixes #575 

This PR make sure that a `pre` that spreads multiple lines is processed correctly and kept inside a single element. It also make sures that when editing an existing pre that it's not separated in multiple elements.

To test:
 - Start the demo app
 - Open the demo with content for standard HTML
 - Check that the PRE element is processed and displayed correctly
 - Switch to HTML and check there is still a single PRE
 - Switch to visual mode and write inside the PRE and add a new line
 - Switch to HTML mode and see if it's still a single PRE.


SergioEstevao:

SergioEstevao:
@diegoreymendez know that we solved other issues with the merge of elements, do you want to give a second look on this?

diegoreymendez:
Works as described.

Issue #938, opened 2018-04-25
Support for Galleries


SiobhyB:
Requests to add a gallery to a post or page comes up fairly often in support. Most recently (for me) in 277029-h. 

If iOS users want to use multiple images within a post at the moment, then they need to either upload them one by one (which can take up a **lot** of space!) or move to a browser so that they can add a gallery there.

It would be wonderful if this popular feature was support within the app.

formosattic:
Also requested in 279055-h.

SiobhyB:
From a four star review:

> Great app to create blog posts on the go. The only one feature that’s missing and I’d absolutely love would be he option to select multiple photos and insert as a gallery like you’re able to in the web dashboard. Without this, I still have to use the web version to create new entries as all my posts consist of multiple images. Please add this if you can, it would really make all the difference.

sjaved87:
Another user requested this feature in #1265645-zen

SiobhyB:
Another request for this feature in 1270552-z.

SiobhyB:
From a three star review:

> The app is very useful but it is missing some key features that I need to manage my blog. There is no tab for WordAds on the app and there are no options for inserting photos- I use circular photo effects for all my blog posts but cannot insert these on the app. I therefore have to switch to a browser to add the circular images, making things a lot more difficult as I constantly have to switch between devices (I need to use the app as all my photos are stored on my iPad, not my laptop).

I'm assuming the user's referring to the circular gallery effect.

SiobhyB:
Another request for this feature in 1280268-z.

nellofn:
Another public request for iOS Gallery support in https://ios.forums.wordpress.org/topic/space-on-ipad-and-iphone

KokkieH:
Via [Twitter](https://github.com/wordpress-mobile/AztecEditor-iOS/issues/938)

SiobhyB:
In 1502213-z, deep linking caused some confusion for a user who regularly makes use of the gallery feature. It'd be ideal if they were able to upload galleries from the app.

thehenrybyrd:
From a 1-star review:
> Not possible to create a Photo-Gallery! 😡🤬

I assume we won't be adding gallery support to Aztec, but I do see we have an issue over in https://github.com/wordpress-mobile/gutenberg-mobile/issues/685 for adding the Gutenberg gallery block.

PR #973, opened 2018-06-11
Initial code support for galleries

diegoreymendez:
### Details:

Initial code support for Galleries.  Unfortunately, rendering the galleries is not yet implemented as this turned out to be much more complicated than initially planned.

Related to #938.

### Important:

While the visual results are far from impressive right now, this brings us so much closer to being able to support Galleries in Aztec.

### Description:

In this PR we:

- Reorganized the file structure a bit.  Verify this by opening the project file.
- Added [`ElementToTagConverter`](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-bf398e64e794673d835aea36c26d8ad9R5) which should replace `HTMLProcessor` (much simpler to use).  Used it to implement a [`GalleryElementToTagConverter`](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/973/files#diff-039b6f6028a93d6e5c116ef0d1efc133R4) for the gallery conversion process.
- Added [`GenericElementToTagConverter`](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-a5a138be431d0988d66b04e9b6b1d160R3) to convert elements to tags.  This takes some code out from `HTMLSerializer` and simplifies it.
- Turned [`DefaultHTMLSerializer`](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-5cd019148209f40b2fcd85081d9a4fd5L1) into just [`HTMLSerializer`](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-27a446462711ad3787bdd36b873fbbb2R10).  Customization will be handled through the previously mentioned `ElementToTagConverter`.
- Added [`HTMLSerializerCustomizer`](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-27a446462711ad3787bdd36b873fbbb2R4) which is how plugins can implement custom [element-to-tag converters](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-668aa2c6bb496dbe292a6e49cd3d563fR88).
- `HTMLSerializer` can now prettify [on each call](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-27a446462711ad3787bdd36b873fbbb2R40) instead of having to be initializer with the prettifier ON / OFF.
- Added [void element support](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-6e3cf337795b90b71a195f1f9bea17d4R18) to `Element`.
- Added [several HTML elements](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-6e3cf337795b90b71a195f1f9bea17d4R51) to the list of default HTML elements.
- Added a new [attachment-to-element](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-23878244b0c4a147c2065ceff3171c6bR4) converter type so we can allow plugins to customize output conversion of gallery attachments.  This also allowed us to remove a lot of details from [`AttributedStringParser`](https://github.com/wordpress-mobile/AztecEditor-iOS/compare/develop...issue/938-support-for-galleries?expand=1#diff-3691618ea04c6533e3ab78931758dc50L895).
- Added [some unit tests](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/973/files#diff-576b03c00e954d7947b922a9e14f3d6cR5).

### Before:

Gallery captions were maintained as raw text during the visual / HTML mode changes.

gallerybefore

### After:

galleryafter

### Testing:

Simply follow the steps shown in the GIFs below, to test that the gallery caption is properly converter back and forth.


SergioEstevao:
The test are failing but it's just because the gallery elements are self-closing but the tests are not expecting it.

Aztec/Classes/Libxml2/Converters/Out/HTMLSerializer.swift
```diff
@@ -1,9 +1,206 @@
...
+
+    /// Converters
+    private let genericElementConverter = GenericElementToTagConverter()
+    
+    private let customizer: HTMLSerializerCustomizer?
+    
+    /// Default Initializer
+    ///
+    /// - Parameters:
+    ///     - prettyPrint: Indicates whether if the output should be pretty-formatted, or not.
```
SergioEstevao: This parameter no longer exists.

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
```diff
@@ -199,7 +199,15 @@ public class ElementNode: Node {
     /// - Returns: `true` if this is a block-level element.  `false` otherwise.
     ///
     public func isBlockLevel() -> Bool {
```
SergioEstevao: Why not make this function and the next two readyonly calculated vars?

Aztec/Classes/Processor/HTMLProcessor.swift
```diff
@@ -66,8 +66,8 @@ open class HTMLProcessor: Processor {
     
     // MARK: - Initializers
     
-    public init(tag: String, replacer: @escaping Replacer) {
-        self.tag = tag
+    public init(for element: Element, replacer: @escaping Replacer) {
```
SergioEstevao: A bit worried about this in the context of using HTMLProcessor in other places why force the use of an Element class? this could be just an convenience init no?

Aztec/Classes/TextKit/MediaAttachment.swift
```diff
@@ -268,7 +268,7 @@ open class MediaAttachment: NSTextAttachment {
         ensureImageIsUpToDate(in: textContainer)
 
         if image == nil {
-            return .zero
+            return CGRect(x: 0, y: 0, width: 20, height: 20)
```
SergioEstevao: Is this to make sure that something is show visually even when an image is not set?

SergioEstevao: If yes, can we create a constant called minimum size, or something similar?

WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Calypso/GalleryShortcode/GalleryElementConverter.swift
```diff
@@ -0,0 +1,151 @@
+import Aztec
+import Foundation
+
+public extension Element {
+    static let gallery = Element("gallery")
+}
+
+/// Provides a representation for `<img>` element.
```
SergioEstevao: Comment outdated.

Aztec/Classes/TextKit/MediaAttachment.swift
diegoreymendez: Sounds good!

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
diegoreymendez: I only use computed properties with algorythms with `O(1)` complexity.  That way the caller knows that whenever they're accessing a computed property there won't be extra processing and they can access it multiple times without big trouble.

This method contains a call to `Swift.Array.contains()` which has a `O(n)` complexity, so I'd prefer the caller to know that it's best to avoid calling this multiple times in a row.

SergioEstevao:
Thanks for the changes! :shipit:

PR #975, opened 2018-06-13
Separates concerns in Plugin.

diegoreymendez:
### Description:

This PR splits `Plugin` into `Plugin.InputCustomizer` and `Plugin.OutputCustomizer` to better separate concerns.

There are no functional changes. The idea of this PR is to make the code slightly easier to maintain and follow, as `Plugin` was becoming a bit too crowded with responsibilities.

### Testing:

Some manual tests against the sample content.
Run the automatic unit tests.


diegoreymendez:
Thanks!

PR #979, opened 2018-06-15
Fixes gutenberg self-closing blocks.

### LINKED ISSUE #959: Implement Gutenberg renderer for self-closing blocks.
Implement Gutenberg renderer so we can represent self-closing blocks.

### LINKED ISSUE #968: Support Gutenberg self-closing blocks.
Add support for Gutenberg self-closing blocks into WordPressEditor.

A lot of the necessary code [has already been merged](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/962), but we had to [disable](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/962/files#diff-41398ea2465d3938b92b9bba467884dfR29) self closing-block parsing due to an issue with our `NSAttributedString` -> HTML Tree converter.

diegoreymendez:
### Description:

WordPressEditor now supports Gutenberg self-closing blocks.

Fixes #959 
Fixes #968

### Details:

The problem we had is that Aztec doesn't recognize attachments as block-level style representations.  So it wraps Gutenberg blocks inside `<p>` tags.

The ideal solution would be to rewrite the logic for the output HTML serialization from scratch.

The proposed solution in this PR is to post-process the HTML to remove the self-closing Gutenberg blocks from the wrapping `<p>` tags.  It's ugly and hacky, but it's the only easy way I found to get this working in a reasonable amount of time.

### Testing:

Manual testing for now.  I'll be adding automated testing for all of our Gutenberg compatibility code soon.

### DIFF
--- a/Example/Example/SampleContent/gutenberg.html
+++ b/Example/Example/SampleContent/gutenberg.html
@@ -138,3 +138,5 @@
 <!-- wp:separator -->
 <hr class="wp-block-separator" />
 <!-- /wp:separator -->
+
+<!-- wp:latest-posts /-->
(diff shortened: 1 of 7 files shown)


SergioEstevao:
@diegoreymendez so I've tested this and it looks Aztec is now able to handle the self-close Gutenberg blocks (latest posts and categories) correctly.

Presentation wise is not perfect has you can bellow:

screen shot 2018-06-15 at 11 27 21

But I think it's good enough for now, we can improved the presentation on a later PR. For Example parse the name of the block and show inside the same way we do for unknow HTML elements.

PR #986, opened 2018-06-18
Fix single spaces are being removed

### LINKED ISSUE #886: Spaces between tags removed on Save and HTML toggle
### Expected behavior

Spaces between tags should be preserved when I toggle HTML mode or when I save my post.

### Actual behavior

Spaces are removed between words that have separate tags.

Note that the GIF below shows WPiOS, but this can also be replicated in the Aztec demo app:

1. Type some text (multiple words) – e.g. "one two three four"
2. Select one word in the text (double tap it) and apply a tag such as bold or a link
3. Select the next word in the text (double tap it) and apply another tag
4. Switch to HTML mode and then back again, and notice that the words are joined

This also happens in WPiOS if you complete steps 1 - 3 on a new post and then tap `...` in the nav bar > Save as Draft. After the save completes, you'll notice the words join up.

links-2

SergioEstevao:
Fixes #886 

To test:
 - Run the demo app
 - Open the editor on HTML mode and write ```<strong>WordPress<strong> <em>App</em>```
 - Switch to Visual Mode
 - Make sure the space between WordPress and App is kept.

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringSerializer.swift
@@ -237,6 +237,10 @@ private extension AttributedStringSerializer {
     }
     
     private func sanitize(_ text: String) -> String {
+        guard text != String(.space) else {
+            return text
+        }
+
         let hasAnEndingSpace = text.hasSuffix(String(.space))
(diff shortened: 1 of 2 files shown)


diegoreymendez:
Issue created for the spacing problem: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/991

PR #988, opened 2018-06-18
Issue/987 refactor closing of void elements

### LINKED ISSUE #987: Make sure that self-closing void elements are closed the same way as Gutenberg
### Expected behavior

When opening post with self-closing elements like ```<br/>  <hr/>``` I'm expecting Aztec to keep the same format on the elements and write back on the same format.

### Actual behavior

Instead we get the following ```<br>  <hr>```, while this is correct HTML it make changes to the HTML that then are detected has changes on the post and cause unnecessary saves/versions.

- Open the demo app
- Open the Gutenberg demo
- Switch to HTML
- Notice that all ``` <hr/> and <br/>``` where changed to ```<br>  <hr>```.

SergioEstevao:
Fixes #987 

To test:
 - Open the Gutenberg demo
 - Switch to HTML mode
 - Check that all tags like ```<br/>  <hr/>``` keep the closing forward dash

### DIFF
--- a/Aztec/Classes/ElementConverters/Output/Implementations/GenericElementToTagConverter.swift
+++ b/Aztec/Classes/ElementConverters/Output/Implementations/GenericElementToTagConverter.swift
@@ -17,7 +17,7 @@ private extension GenericElementToTagConverter {
     private func openingTag(for node: ElementNode) -> String {
         let attributes = serialize(attributes: node.attributes)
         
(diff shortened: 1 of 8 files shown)


SergioEstevao:
After talked with @diegoreymendez we decided this is not needed for Gutenberg compatibility has the current implementation is valid HTML and doesn't interfere with Gutenberg blocks parsing.

PR #992, opened 2018-06-19
Makes PRE mergeable while maintaining newlines.

diegoreymendez:
This PR changes the logic for merging `PRE` slightly so that it maintains the separation of paragraphs and lines.

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
@@ -367,7 +367,10 @@ private extension AttributedStringParser {
             return false
         }
 
-        leftMerger.children = leftMerger.children + rightMerger.children
+        // Pre has a custom joining logic because it joins different paragraphs without removing the paragraph separator.
+        let junctureNodes: [Node] = leftMerger.type == .pre ? [TextNode(text: String(.paragraphSeparator))] : []
+        
+        leftMerger.children = leftMerger.children + junctureNodes + rightMerger.children
 
         return true
     }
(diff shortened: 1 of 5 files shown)


Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
```diff
@@ -381,17 +384,15 @@ private extension AttributedStringParser {
...
         }
+        
+        let lastNodeName = lastNode.name
 
         var mergeCandidates: ArraySlice<MergeablePair>
         
-        // TODO: Remove this hack!!  This is a horrible horrible hack, but it's the simplest solution until we can
-        // refactor this Parser to work in a different way, analyzing the NSAttributedString directly for merges.
-        if mergeableNodes.last?.left.type == .figure {
+        if Element.mergeableBlockLevelElementWithoutBlockLevelChildren.contains(lastNode.type) {
```
SergioEstevao: Should we keep the comment above or this is no longer an hack now ?

diegoreymendez: The comment I put there wasn't clear even to me (and I wrote it).  It was offering no value, in my opinion.

Since it wasn't clear at all, I prefer the code to speak for itself, and let whoever analyzes it draw his/her own conclusions.

diegoreymendez: PS: also, it doesn't really apply to the line I put it in... it's more of a comment on the whole class, but even if I add it back I'd have to think how to word it for clarity.

diegoreymendez:
Thanks!

PR #999, opened 2018-06-21
Issue/997 video poster

### LINKED ISSUE #997: Videos without poster images are not displaying a frame from the video.
### Expected behavior

When we have HTML like this: 

```
<video src="https://videos.files.wordpress.com/kUJmAcSf/bbb_sunflower_1080p_30fps_normal.mp4">
```

We should be able to display a frame from the video.

### Actual behavior

Only a small placeholder was being displayed like this:

screen shot 2018-06-21 at 23 07 36

- Start the app
- Open the empty demo content
- Copy the above HTML to the editor while in HTML mode.
- Switch to Visual mode

SergioEstevao:
Fixes #997 

This PR changes the way poster images are being implemented on VideoAttachment. So now the src attribute on VideoAttachment represent the video and the posterURL represents the poster image URL if any.

This way when the TextView ask for an image for the attachment we check if it's a video attachment and then use the correct property for the image, according to this:

- If it's a video attachment and we have a poster image, simple use the poster image URL
- If it's a video attachment but not poster is provided try to extract a frame directly from the video url and use it

This could be also a opportunity to do other two things:

- Refactor the `TextViewAttachmentDelegate` method ```func textView(_ textView: TextView,
                  attachment: NSTextAttachment,
                  imageAt url: URL,
                  onSuccess success: @escaping (UIImage) -> Void,
                  onFailure failure: @escaping () -> Void)``` to just sent the attachment object and not the url and depending of the type of attachment use the correct information

- Merge the `TextViewAttachmentDelegate` and `TextViewAttachmentImageProvider` in a single class and avoid to have two ways of fetching images for an attachment.

I simple did the quickest solution but if any of the above is of interest I can go a do follow up PRs

To test:
 - Open the new sample content Videos and see if the video shows correctly even when no poster image is selected.
 - On the clean HTML demo simple insert this on the HTML mode:
```<video src="https://videos.files.wordpress.com/AvC6H2JI/video-de223da1f6.mp4"  />```
 - Check if the video shows correctly in visual mode

### DIFF
--- a/Aztec/Classes/TextKit/VideoAttachment.swift
+++ b/Aztec/Classes/TextKit/VideoAttachment.swift
@@ -15,31 +15,17 @@ protocol VideoAttachmentDelegate: class {
 ///
 open class VideoAttachment: MediaAttachment {
 
-    /// Attachment video URL
-    ///
-    open var srcURL: URL?
-
     /// Video poster image to show, while the video is not played.
     ///
-    open var posterURL: URL? {
-        get {
-            return self.url
-        }
-
-        set {
-            super.updateURL(newValue)
-        }
-    }    
+    open var posterURL: URL?
     
     /// Creates a new attachment
     ///
     /// - parameter identifier: An unique identifier for the attachment
     ///
     required public init(identifier: String, srcURL: URL? = nil, posterURL: URL? = nil) {
-        self.srcURL = srcURL
-        
-        super.init(identifier: identifier, url: posterURL)
-
+        super.init(identifier: identifier, url: srcURL)
+        self.posterURL = posterURL
         self.overlayImage = Assets.playIcon
     }
 
@@ -48,15 +34,14 @@ open class VideoAttachment: MediaAttachment {
     required public init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
 
-        if aDecoder.containsValue(forKey: EncodeKeys.srcURL.rawValue) {
-            srcURL = aDecoder.decodeObject(forKey: EncodeKeys.srcURL.rawValue) as? URL
+        if aDecoder.containsValue(forKey: EncodeKeys.posterURL.rawValue) {
+            posterURL = aDecoder.decodeObject(forKey: EncodeKeys.posterURL.rawValue) as? URL
         }
     }
 
     /// Required Initializer
     ///
-    required public init(identifier: String, url: URL?) {
-        self.srcURL = nil
+    required public init(identifier: String, url: URL?) {        
         super.init(identifier: identifier, url: url)
     }
 
@@ -82,13 +67,13 @@ open class VideoAttachment: MediaAttachment {
 
     override open func encode(with aCoder: NSCoder) {
         super.encode(with: aCoder)
-        if let url = self.srcURL {
-            aCoder.encode(url, forKey: EncodeKeys.srcURL.rawValue)
+        if let posterURL = self.posterURL {
+            aCoder.encode(posterURL, forKey: EncodeKeys.posterURL.rawValue)
         }
     }
 
     fileprivate enum EncodeKeys: String {
-        case srcURL
+        case posterURL
     }
 
 
@@ -130,7 +115,6 @@ extension VideoAttachment {
             fatalError()
         }
 
-        clone.srcURL = srcURL
         clone.posterURL = posterURL
 
         return clone
--- a/Aztec/Classes/TextKit/VideoAttachment.swift
+++ b/Aztec/Classes/TextKit/VideoAttachment.swift
@@ -22,6 +22,8 @@ open class VideoAttachment: MediaAttachment {
     /// Creates a new attachment
     ///
     /// - parameter identifier: An unique identifier for the attachment
+    /// - parameter srcURL: the url for the video to display
+    /// - parameter posterURL: the url for a poster image for the video
     ///
     required public init(identifier: String, srcURL: URL? = nil, posterURL: URL? = nil) {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: Aztec/Classes/TextKit/VideoAttachment.swift [hunk 1]
```diff
@@ -15,31 +15,17 @@ protocol VideoAttachmentDelegate: class {
...
-
-        set {
-            super.updateURL(newValue)
-        }
-    }    
+    open var posterURL: URL?
     
     /// Creates a new attachment
     ///
     /// - parameter identifier: An unique identifier for the attachment
```
diegoreymendez: Update the comments for the missing parameters.

Example/Example/EditorDemoController.swift
```diff
@@ -1032,9 +1032,48 @@ extension EditorDemoController {
...
+                                                            return
+                                                        }
+                                                        let image = UIImage(cgImage: cgImage)
+                                                        onCompletion(image)
+            })
+        }
+    }
+
     func textView(_ textView: TextView, attachment: NSTextAttachment, imageAt url: URL, onSuccess success: @escaping (UIImage) -> Void, onFailure failure: @escaping () -> Void) {
+        var urlToRequest = url
```
diegoreymendez: Let's try to see how we can make this method more readable.  Right now it has a lot of detail in it.

For instance:
- Line [1076 onward](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/999/files#diff-e7e4649f5f7465f68641a838cc55412fR1076) could go into a method of its own.
- We could use a guard in [line 1059](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/999/files#diff-e7e4649f5f7465f68641a838cc55412fR1059) since the else condition returns.

I know part of this code was not touched by this PR, but since we're adding code to an already large method, I think we should ensure its readability.

Example/Example/EditorDemoController.swift
```diff
@@ -1032,9 +1032,48 @@ extension EditorDemoController {
...
+            guard asset.isExportable else {
+                onError()
+                return
+            }
+            let generator = AVAssetImageGenerator(asset: asset)
+            generator.appliesPreferredTrackTransform = true
+            generator.generateCGImagesAsynchronously(forTimes: [NSValue(time: CMTimeMake(2, 1))],
+                                                     completionHandler: { (time, cgImage, actualTime, result, error) in
+                                                        guard let cgImage = cgImage else {
+                                                            onError()
```
diegoreymendez: Even if we're not currently using it, we should return the error object in the block, since if we ever want to handle it it will be necessary.

Example/Example/EditorDemoController.swift
```diff
@@ -1032,9 +1032,48 @@ extension EditorDemoController {
 
 extension EditorDemoController: TextViewAttachmentDelegate {
 
+    func exportPreviewImageForVideo(atURL url: URL, onCompletion: @escaping (UIImage) -> (), onError: @escaping () -> ()) {
```
diegoreymendez: This method is not part of `TextViewAttachmentDelegate`.  Can we move it to a more appropiate location?

SergioEstevao:
The code fetches the first frame of the video, if it’s black it will show that. If you want to double check
Use the same url in Gutenberg for a video block.

Example/Example/EditorDemoController.swift
```diff
@@ -1032,8 +1032,32 @@ extension EditorDemoController {
 
 extension EditorDemoController: TextViewAttachmentDelegate {
 
-    func textView(_ textView: TextView, attachment: NSTextAttachment, imageAt url: URL, onSuccess success: @escaping (UIImage) -> Void, onFailure failure: @escaping () -> Void) {
+    func exportPreviewImageForVideo(atURL url: URL, onCompletion: @escaping (UIImage) -> (), onError: @escaping () -> ()) {
```
diegoreymendez: Let's move this to a more appropriate location since it's really not related to `TextViewAttachmentDelegate`.

Example/Example/EditorDemoController.swift
```diff
@@ -1032,8 +1032,32 @@ extension EditorDemoController {
...
+                onError()
+                return
+            }
+            let generator = AVAssetImageGenerator(asset: asset)
+            generator.appliesPreferredTrackTransform = true
+            generator.generateCGImagesAsynchronously(forTimes: [NSValue(time: CMTimeMake(2, 1))],
+                                                     completionHandler: { (time, cgImage, actualTime, result, error) in
+                                                        guard let cgImage = cgImage else {
+                                                            DispatchQueue.main.async {
+                                                                onError()
```
diegoreymendez: We shouldn't swallow the error object here, let's send it to the closure.

Example/Example/EditorDemoController.swift
```diff
@@ -1032,8 +1032,32 @@ extension EditorDemoController {
...
+                                                        }
+                                                        let image = UIImage(cgImage: cgImage)
+                                                        DispatchQueue.main.async {
+                                                            onCompletion(image)
+                                                        }
+            })
+        }
+    }
 
+    func downloadImage(from url: URL, success: @escaping (UIImage) -> Void, onFailure failure: @escaping () -> Void) {
```
diegoreymendez: Let's move this method to a more appropriate location since it's not part of `TextViewAttachmentDelegate`.

Example/Example/EditorDemoController.swift
diegoreymendez: As discussed privately, since this is code in the demo VC, we can skip this if you want.

Issue #983, opened 2018-06-15
Images surrounded by figure elements are merged in a single figure


SergioEstevao:
### Expected behavior

When this HTML is inserted: ```
<figure>
  <img src=\"a.png\" class=\"alignnone\">
  <figcaption>caption a</figcaption>
</figure>
<figure>
  <img src=\"b.png\" class=\"alignnone\">
  <figcaption>caption b</figcaption>
</figure>
```
The resulting output should be the same.

### Actual behavior

Instead we are getting this as a result

```
<figure>
  <img src=\"a.png\" class=\"alignnone\">
  <figcaption>caption a</figcaption>
  <img src=\"b.png\" class=\"alignnone\">
  <figcaption>caption b</figcaption>
</figure>
```

### Steps to reproduce the behavior

- On the HTML empty demo
- Copy paste the HTML presented in the expected section on the code editor
- Switch to visual and back to HTML


##### Tested on iPhoneX, iOS 11, Aztec beta 20

Issue #970, opened 2018-06-01
Blockquotes splitting up.


diegoreymendez:
Blockquotes are splitting up when they should not.

![blockquotessplittingup](https://user-images.githubusercontent.com/1836005/40855798-0602ce8c-65ac-11e8-8596-17cec4db208a.gif)

Issue #934, opened 2018-04-12
Visual to HTML paragraph-style-merging-logic needs changes.


diegoreymendez:
Our current merging logic for paragraph styles works "kind of fine" in most cases, but it's not perfect and is causing issues with captions.

It's also losing some info about how paragraph-styles should be merged across paragraphs.

We need to implement the following logic in `AttributedStringParser.swift`:

```
Enumerate paragraphs
	Loop through current paragraph styles (i: index)
		if i == last
			create new element // the last paragraph style can NEVER be merged
		end

		if previousParagraph.paragraphStyles[i] === currentParagraph.paragraphStyles[i] (the editor needs some extra logic to make sure instances are maintained)
			extend existing element
		else
			create new element and exit loop
		end if
	End loop

	// We can keep the current processing mechanism for sub-paragraph styles.
End enumeration
```

PR #1002, opened 2018-06-25
Replaces the paragraph properties output conversion and merging logic.

### LINKED ISSUE #993: Gutenberg code blocks are being broken by Aztec.
### Expected behavior

Gutenberg code blocks should not be broken by Aztec.

### Actual behavior

They are not merged properly, and the comment delimiters are removed for some reason.

### Tech info

Block-level elements inside of `<pre>` should be able to merge even when they're the topmost block-level element, and they should insert a `.paragraphSeparator` between the paragraphs.

Also, here's an example code block:

```html
<!-- wp:code -->
<pre class="wp-block-code"><code>testing
test.   ing
     testing some more</code></pre>
<!-- /wp:code -->
```

codeblockfail

diegoreymendez:
### Description:

Replaces the paragraph properties output conversion and merging logic.

This will allow us to fix #993 and #934.

### Details:

The reason why this change is important is that we can now customize how paragraph-styles are merged on a case by case basis.  Meaning we can decide to support merging two adjacent paragraph styles based on any kind of equality we want to implement.

Right now supporting captions is impossible without these changes.

### Testing:

Run the unit tests.
Open the content samples and convert to visual and back to HTML.

### DIFF
--- a/Aztec.xcodeproj/project.pbxproj
+++ b/Aztec.xcodeproj/project.pbxproj
@@ -155,6 +155,7 @@
 		F1C05B991E37F99D007510EA /* Character+Name.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1C05B981E37F99D007510EA /* Character+Name.swift */; };
 		F1C05B9D1E37FA77007510EA /* String+CharacterName.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1C05B9C1E37FA77007510EA /* String+CharacterName.swift */; };
 		F1C05B9F1E37FD2F007510EA /* NSAttributedString+CharacterName.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1C05B9E1E37FD2F007510EA /* NSAttributedString+CharacterName.swift */; };
+		F1CF7F9020E14E2200B24173 /* NewAttributedStringParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1CF7F8F20E14E2200B24173 /* NewAttributedStringParser.swift */; };
 		F1DE83D51EF20493009269E6 /* UIColor+Parsers.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1DE83D41EF20493009269E6 /* UIColor+Parsers.swift */; };
 		F1E1B7762062BD47004642BB /* NSMutableAttributedStringParagraphProperty.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1E1B7752062BD47004642BB /* NSMutableAttributedStringParagraphProperty.swift */; };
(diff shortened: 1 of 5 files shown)


diegoreymendez:
Fyi @koke - I managed to squeeze this into nearly two days of coding.  I know we want to switch priorities but this has been bringing issues to us for a while.

diegoreymendez:
@SergioEstevao - I think we should avoid merging this PR into the current release of WPiOS.  We can wait to integrate your changes before merging this.

diegoreymendez:
@SergioEstevao - After cleaning the project I'm getting 3 failing unit tests.  I'll be fixing those before I move ahead with the PR.

diegoreymendez:
Ready for review.

SergioEstevao:
Just tested it and this also fixes #983 !

diegoreymendez:
Added a small change so that this will also fix https://github.com/wordpress-mobile/AztecEditor-iOS/issues/970.

I'll update the description.

diegoreymendez:
Cherry-picked some breaking unit tests by @SergioEstevao into this PR, since they're fixed by these changes.

Aztec/Classes/TextKit/ParagraphProperty/HTMLLi.swift
```diff
@@ -15,6 +15,14 @@ class HTMLLi: ParagraphProperty {
     }
 
     static func ==(lhs: HTMLLi, rhs: HTMLLi) -> Bool {
-        return lhs === rhs
+        return false
```
SergioEstevao: Should this always return false? shouldn't we delegate to the method isEqual ?

Aztec/Classes/NSAttributedString/Conversions/AttributedStringParser.swift
```diff
@@ -365,263 +379,267 @@ private extension AttributedStringParser {
...
-        mergeablePair.left.children = mergeablePair.left.children + junctureNodes + mergeablePair.right.children
-
-        return true
-    }
-
-
-    /// Finds the last valid Mergeable Pair within a collection of mergeable nodes
+    /// - Parameters:
+    ///     - newProperties: the properties from the paragraph being converted.
+    ///     - previousConversions: the conversions used for the previous paragraph.
```
SergioEstevao: Missing comment for styleNodes

Aztec/Classes/TextKit/ParagraphProperty/HTMLLi.swift
diegoreymendez: This should ideally be `lhs == rhs`, but the problem we have is that Aztec (the editor logic) is using the same `HTMLLi` instance for separate list items, so you never get two bullets.  This is not new, it's always been like that but with the previous merging logic it wasn't noticeable.

Making Aztec's logic smart enough to know where to apply `HTMLLi` instances means changes to the list formatter, which is way beyond my current scope with this PR.

SergioEstevao: Ok them, probably flag this on a new issue for us to fix later on.

PR #1006, opened 2018-07-06
Fixes a bug that was causing paragraphs to merge.

### LINKED ISSUE #1005: Paragraphs are being merged when they shouldn't.
Paragraphs are being merged when they shouldn't.

paragraphsmerged

diegoreymendez:
### Description:

Fixes a bug that was causing paragraphs to merge when they shouldn't.

Fixes #1005 

This is what the original issue looked like:

paragraphsmerged

### Details:

- [Added unit tests](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1006/files#diff-40513c251786f2880b02c4f1f38730f3R40) that can automatically detect this issue if it ever arised again.
- [Reorganizes the code a bit](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1006/files#diff-8f0997fe69197252915ff1a894213dfdR6) since to fix this issue I have to reuse some of the code we already have from [a new location](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1006/files#diff-48ee906bc2164480c9dfa48c13d0f995R6).
- Also fixes a bug with [how we were handling unicode characters](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1006/files#diff-56b82478ff0ea96a9b99d6498f29ff37R18).  This issue surfaced in one of our new unit tests.
- [Changes the expectations of a unit test](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1006/files#diff-40513c251786f2880b02c4f1f38730f3R180).  Unfortunately it seemed to be working fine previously due to the bug we had, but once the bug was fixed, a problem with non-closed Gutenberg delimiters surfaced.  I expect it to be minor.

### Testing:

Run the unit tests.

Also:
- Create a new empty Gutenberg / Calypso post in the demo app.
- Type different paragraphs of text in visual mode.
- Switch to HTML and back and make sure the text is not merged into a single paragraph.

### DIFF
--- a/Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
+++ b/Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
@@ -105,6 +105,12 @@ public class ElementNode: Node {
     }
 
     // MARK: - Node Queries
+    
+    public func attribute(named name: String) -> Attribute? {
+        return attributes.first { (attribute) -> Bool in
+            return attribute.name.lowercased() == name.lowercased()
+        }
+    }
 
     func stringValueForAttribute(named attributeName: String) -> String? {
 
--- a/WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Gutenberg/GutenbergAttributeEncoder.swift
+++ b/WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Gutenberg/GutenbergAttributeEncoder.swift
@@ -21,15 +21,15 @@ class GutenbergAttributeEncoder {
     
     // MARK: - Encoding Strings
     
-    func closerAttributes(_ text: String) -> Attribute {
+    func closerAttribute(_ text: String) -> Attribute {
         return attribute(.blockCloser, withValue: text)
     }
     
-    func openerAttributes(_ text: String) -> Attribute {
+    func openerAttribute(_ text: String) -> Attribute {
         return attribute(.blockOpener, withValue: text)
     }
     
-    func selfClosingAttributes(_ text: String) -> Attribute {
+    func selfClosingAttribute(_ text: String) -> Attribute {
         return attribute(.selfCloser, withValue: text)
     }
     
--- a/WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Gutenberg/GutenbergAttributeEncoder.swift
+++ b/WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Gutenberg/GutenbergAttributeEncoder.swift
@@ -1,0 +1,52 @@
+import Foundation
+import Aztec
+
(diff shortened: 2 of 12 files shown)


diegoreymendez:
This PR will fix https://github.com/wordpress-mobile/WordPress-iOS/issues/9706 for Wordpress.

Aztec/Classes/Libxml2/DOM/Data/ElementNode.swift
```diff
@@ -105,6 +105,12 @@ public class ElementNode: Node {
     }
 
     // MARK: - Node Queries
+    
+    public func attribute(named name: String) -> Attribute? {
+        return attributes.first { (attribute) -> Bool in
```
jleandroperez: No need to change anything. I'm just annoying: `(attribute) -> Bool in`, in this case, can be compressed to just `attribute in` (because the return type gets inferred)

WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Gutenberg/GutenbergAttributeEncoder.swift
```diff
@@ -0,0 +1,52 @@
...
+    private func attribute(_ attribute: GutenbergAttribute, withValue text: String) -> Attribute {
+        let base64String = encode(text)
+        
+        return Attribute(name: attribute, value: .string(base64String))
+    }
+    
+    // MARK: - Base64 Encoding
+    
+    private func encode(_ string: String) -> String {
+        let data = string.data(using: .utf16)!
```
jleandroperez: The force unwrap creeps me out. I know it shouldn't usually fail... but we may wanna consider replacing that with a guard?

diegoreymendez: As discussed privately:

- There seems to be no way to handle this without ignoring the error.
- We're not sure if there's any condition that can make this fail, or what condition that is.
- This code has only been moved.  It has been in production for a while and is not breaking.

So I'm moving ahead as it is.

nellofn:
Just a heads up that users are still reporting this issue in iOS version 10.3.1:

1265474-zen
1266666-zen
1267186-zen
1267119-zen
1266984-zen
1266764-zen

I let them know to keep an eye on a future app release, and will reach out to them all once there's an update on this.

diegoreymendez:
Thanks for the info @nellofn.

You should probably add these to [the issue in the WPiOS repo](https://github.com/wordpress-mobile/WordPress-iOS/issues/9706) for consistency since that's where we'd normally have the history.

nellofn:
Good to know @diegoreymendez. I saw the issue as closed, so that's why I added it here.

PR #1007, opened 2018-07-09
WordPressEditor is now a separate pod.

diegoreymendez:
WordPressEditor was previously a subspec of the Aztec pod, now it's a separate one.

This PR makes it possible for us to publish new numbered versions of Aztec again.

For some reason I forgot to create this PR to merge these changes back into `develop`.  To avoid slowing down other dependent work, and since this integration is already live in WPiOS, I'll move ahead and merge once the automated build is green.


diegoreymendez:
Merging without a review as detailed above.

Issue #1024, opened 2018-08-16
Paste oddities.


diegoreymendez:
### Issue:

When pasting content in Aztec, the some of the visual styling shown in visual mode is not properly maintained when switching to HTML mode and back.

Some of the styling is even changed.

### Tasks:

- [x] After pasting text that ends with a certain text color (as shown below), switching to HTML mode and back should not change the color for all of the content.
- [x] When pasting content from Safari, paste the HTML in order to maintain the styling.
- [x] Allow apps to discard certain HTML elements or attributes when pasting HTML into Aztec.
- [x] When pasting attributed text, drop all non-supported styles.

### How to reproduce:

![pasteissues](https://user-images.githubusercontent.com/1836005/44210126-09d71380-a13c-11e8-8b70-c7551fedfc25.gif)

diegoreymendez:
Closing.  We can open a more specific issue if necessary.

PR #1026, opened 2018-08-17
Better support for pasting HTML in Aztec.

### LINKED ISSUE #1020: Is this possible to insert custom html code programatically in between of html
### Expected behavior
Hi there I need to insert my custom html at a any point in editor at run time.

### Actual behavior
I am unable to find a function which can add any custom html at any point . I can only find a method to get html and set it again but it will add it at the end of code not in the middle or any point in editor. I am looking for a function just like 

> insertText()

which can add certain text at any point. I hope you understand my concern.

Thanks

diegoreymendez:
### Description:

This PR improves drastically our support for pasting HTML content in Aztec.

Moves forward #1024.
Fixes #1020.

### Details:

In this PR I:

- Cleaned up the `UIPasterboard+Helpers.swift` code a bit.
     - Converted some calculated properties into methods, since the `Data` -> `NSAttributedString` conversion is most likely `O(n)` at least.
     - Reordered some methods alphabetically, to make it easier to look them up.
- Added some extra methods to `UIPasteboard` for handling URLs and HTML.
- Improved the pasting logic and cleaned up its code a bit.
- Added a method to be able to replace ranges in the attributed string with new HTML content.

Here's an example of how this looks:

pastehtml

### Known issues / limitations:

The pasted HTML can contain a lot of information (CSS inline styles, IDs, etc).

We'll add a customization step in the plugin so that each app can define if there's any HTML content than should be stripped when pasting (most likely attributes that the app doesn't want to use).

### Testing:

Note: for all tests use the standard empty editor and the Gutenberg / Calypso one.

**Test for regressions:**

1. Make sure pasting URLs work.  Try to undo the operation.
2. Make sure pasting regular text works.  Try to undo the operation.
3. Run the unit tests.

**Test pasting HTML content:**

1. Open Safari, and go to the Wikipedia page for WordPress.
2. Copy any chunk of text.
3. Paste it anywhere in the editor.
4. Switch to HTML and back, and make sure the content looks the same as when you pasted it.

### DIFF
--- a/Aztec/Classes/Extensions/UIPasteboard+Helpers.swift
+++ b/Aztec/Classes/Extensions/UIPasteboard+Helpers.swift
@@ -7,67 +7,86 @@ import UIKit
 //
 extension UIPasteboard {
 
-    /// Attempts to load the Pasteboard's contents into a NSAttributedString Instance, if possible.
+    /// Attempts to retrieve the Pasteboard's contents as an attributed string, if possible.
     ///
-    func loadAttributedString() -> NSAttributedString? {
-
-        if let string = aztecAttributedString {
+    func attributedString() -> NSAttributedString? {
+        
+        if let string = aztecAttributedString() {
             return string
         }
 
-        if let string = RTFDAttributedString {
+        if let string = rtfdAttributedString() {
             return string
         }
 
-        if let string = RTFAttributedString {
+        if let string = rtfAttributedString() {
             return string
         }
 
-        if let string = richTextAttributedString {
+        if let string = richTextAttributedString() {
             return string
         }
 
-        return plainTextAttributedString
+        return plainTextAttributedString()
+    }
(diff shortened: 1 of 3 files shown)


SergioEstevao:
@diegoreymendez I tried to paste the following html : 
```
<p>Hello to <a href="http://wordpress.com">Wordpress</a></p>
``` 
That translates to:
<p>Hello to <a href="http://wordpress.com">Wordpress</a></p>

When doing a normal paste the link is preserved, when doing copy and match style the link is lost.

Is this expected?

SergioEstevao:
So I did some tests on WordPress.com and Google and it looks that the copy paste of HTML with URLs is working has expected. So I will approve this PR :).

diegoreymendez:
Thanks @frosty and @SergioEstevao !

PR #1030, opened 2018-08-21
Improve pasting label

### LINKED ISSUE #1028: Improve label on pasting actions in Edit Menu
There are currently 2 ways to paste clipboard content into Aztec via an Edit Menu: `Paste` and `Paste and Match Style`. These options seem to be confusing to folks, especially the `Paste and Match Style` option.

### Proposal

Change `Paste and Match Style` menu item to instead say `Paste Without Formatting`. This label is more explicit.

* `Paste` should paste the content as is, and retain style if possible
* `Paste without formatting` should paste the content as plain-text, removing styling

Side note: I'm not sure if we should also retain links for the `paste without formatting` option. Thoughts @mattmiklic @elibud ?

1. Copy some content like a paragraph with some links from a mobile web browser
2. Start a new post or open an existing post
3. Tap somewhere in canvas to show Edit Menu
4. See menu items

diegoreymendez:
### Description:

"Paste and Match" has been renamed to "Paste without Formatting".
There's also a minor refactor in the code, to make it simpler and to make sure links are stripped when pasting without formatting.

**EDIT:** "Paste without Formatting" didn't have undo support.  I've just added it.

Fixes #1028 

### Testing:

**Test 1:**
- Open the plain HTML standard demo.
- Copy any styled text.
- Move to caret to any other position.
- Paste.

Make sure the original style is maintained in the pasted text.

**Test 2:**

Repeat test 1 but copy styled text from Safari.

**Test 3:**
- Open the plain HTML standard demo.
- Copy any styled text.
- Move to caret to any other position.
- Paste without Formatting.

Make sure the formatting from the caret position is used for the pasted text.

**Test 4:**

Repeat test 3 but copy styled text from Safari.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -437,18 +437,17 @@ open class TextView: UITextView {
     }
 
     @objc open func pasteAndMatchStyle(_ sender: Any?) {
-        guard let string = UIPasteboard.general.attributedString()?.mutableCopy() as? NSMutableAttributedString else {
+        guard let string = UIPasteboard.general.string else {
             super.paste(sender)
             return
         }
 
-        let range = string.rangeOfEntireString
-        string.addAttributes(typingAttributesSwifted, range: range)
-        string.loadLazyAttachments()
+        let attributedString = NSMutableAttributedString(string: string, attributes: typingAttributesSwifted)
+        attributedString.loadLazyAttachments()
 
-        storage.replaceCharacters(in: selectedRange, with: string)
+        storage.replaceCharacters(in: selectedRange, with: attributedString)
         notifyTextViewDidChange()
-        selectedRange = NSRange(location: selectedRange.location + string.length, length: 0)
+        selectedRange = NSRange(location: selectedRange.location + attributedString.length, length: 0)
     }
     
     // MARK: - Try Pasting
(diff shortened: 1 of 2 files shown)


diegoreymendez:
Fyi @mattmiklic and @iamthomasbishop

iamthomasbishop:
> Just realized that "paste without formatting" removes any images or video that we have in the content? is that expected? desired?

I wouldn't expect images/videos to be removed. However, if there was additional styling on the source image, I would expect that to be removed.

diegoreymendez:
> Just realized that "paste without formatting" removes any images or video that we have in the content? is that expected? desired?

Good catch @SergioEstevao, that's definitely a bug.

I'll fix it and add an automated test to check for this, before resubmitting my PR for review.

diegoreymendez:
@SergioEstevao - This is ready for another look.

I fixed the issue you reported, and added unit tests to make sure pasting of videos and images works fine for both paste operations.

Let me know if there's something else.  I really appreciate that you found such a nasty issue!

koke:
When you have just plain text in the pasteboard (e.g. copy from title or any other text field), should we still show the "Paste without Formatting" option? I don't think I'd bother unless it's an easy change, but I thought it was worth mentioning.

diegoreymendez:
@koke - By looking at the API, I'm not sure if that check would be straightforward.  It'd warrant at least some investigation on my end... I imagine it's doable but it feels to me like the effort-to-reward ratio wouldn't be very high?

As a side note - if we want to take care of that I'd open a different issue.  Let me know.

PR #1035, opened 2018-08-29
Implements editing of target attribute in .A elements

SergioEstevao:
Refs: https://github.com/wordpress-mobile/WordPress-iOS/pull/10059 and https://github.com/wordpress-mobile/WordPress-iOS/issues/7670

This PR adds support for the target attribute on `a` elements.

This is implemented using the already existing HTMLRepresentation element.

To test:
 - Open the demo app.
 - Open the demo with content
 - Select the links provided in the sample content
 - Tap on the link button
 - see if the target shows correctly
 - Edit the target and change the value
 - Switch to HTML 
 - Check the value is updated correctly.

### DIFF
--- a/Aztec/Classes/Formatters/Implementations/LinkFormatter.swift
+++ b/Aztec/Classes/Formatters/Implementations/LinkFormatter.swift
@@ -2,6 +2,8 @@ import UIKit
 
 class LinkFormatter: StandardAttributeFormatter {
 
+    var target: String?
+
     init() {
         super.init(attributeKey: .link,
                    attributeValue: NSURL(string:"")!,
@@ -9,6 +11,7 @@ class LinkFormatter: StandardAttributeFormatter {
     }
 
     override func apply(to attributes: [NSAttributedStringKey: Any], andStore representation: HTMLRepresentation?) -> [NSAttributedStringKey: Any] {
+        var finalRepresentation: HTMLRepresentation?
 
         if let representation = representation,
             case let .element(element) = representation.kind {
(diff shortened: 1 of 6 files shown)


Aztec/Classes/Formatters/Implementations/LinkFormatter.swift
```diff
@@ -2,13 +2,16 @@ import UIKit
 
 class LinkFormatter: StandardAttributeFormatter {
 
+    var target: String?
```
diegoreymendez: I think it would be a bit cleaner if we changed this to:

```swift
private let target: String?
```

And changed the initializer to:

```swift
init(target: String? = nil) {
    self.target = target
    ...
```

Rationale: it's not really very likely that we will need to either read or write the target of a formatter again after instantiating it.

Aztec/Classes/Formatters/Implementations/LinkFormatter.swift
```diff
@@ -21,15 +24,27 @@ class LinkFormatter: StandardAttributeFormatter {
...
-            }            
+            }
+            finalRepresentation = representation
         } else {
 
             // There's no support fora link representation that's not an HTML element, so this
             // scenario should only be possible if `representation == nil`.
             //
             assert(representation == nil)
+            var attributes = [Attribute]()
```
diegoreymendez: Readability: Add some spacing around the code below.  Right now it's a bit too tight.

SergioEstevao:
@diegoreymendez addressed your review comments, you can also check the integration PR here: wordpress-mobile/WordPress-iOS#10059

Issue #1032, opened 2018-08-27
Raise Automated Testing Coverage above 70%


diegoreymendez:
### Description:

In order to provide a more stable library we need to improve our automated tests coverage.

We're currently sitting at 63.96% unit testing coverage [according to xcov](https://wordpress-mobile.github.io/AztecEditor-iOS/).

We need to bring it up to at least 70% (lowered from %80 for a better depiction of what was achieved).

diegoreymendez:
Changing the name and description to match what was actually achieved during last week.

The 80% goal was an interesting one, but at this point it's unsure if there'll be any other effort in that direction.

PR #1043, opened 2018-08-31
Adds more unit tests.

diegoreymendez:
### Description:

Adds more unit tests.

Fixes an issue in ImageAttachment+WordPress that was causing the `class` attribute to never be removed.

Moves forward #1032 

### Testing:

Just run the unit tests and make sure they don't fail.


diegoreymendez:
Thank you @SergioEstevao !

PR #1046, opened 2018-09-04
Adds an ENTER-key optional-override-callback to Aztec.

diegoreymendez:
### Description:

Adds an ENTER-key optional-override-callback to Aztec, so that consuming Apps can override what happens when ENTER is pressed in the editor.

Also added some unit tests to make sure the override works as expected.

### Testing:

Simply run the unit tests.
Optionally define an override that just returns `true` and make sure newlines are not added to the editor.


diegoreymendez:
I'm thinking this may not be necessary at all.  If the user wants to intercept a keypress they can subclass our TextView.

What do you think?

diegoreymendez:
Closing as we'll just subclass to intercept the ENTER keypresses.

PR #1068, opened 2018-10-12
Proposal to fix pasting from Notes App

### LINKED ISSUE #1066: Paste from Apple Notes is not working properly
In iOS 11, copying and pasting (mostly) worked just fine. Thanks to @aerych, we were able to do a test:

image from ios.

In iOS 12, it turns the text into something like

>[HEAD] 
[META] 
[META] 
[TITLE] 
[META][STYLE] 
[BODY]

Creating a post with bold, italic, headings, and bulleted lists in Notes, then copy/pasting into Aztec seems to consistently result in this behaviour. 

##### Tested on iPhone X[S], iOS 12.1, Aztec (??? - version in WP 10.9)

etoledom:
Fixes #1066 

As @aerych mentioned, pasting from the Notes App was working correctly on WPv10.8.1.
The aztec changes for that release was from `1.0.0-beta.25.1` to `1.0.2` (I believe).

The important PR that seems to have broken pasting from Notes app is #1026 (Better support for pasting HTML).

The down side of that change is that now, pasting from Notes will be translated into HTML, and that HTML brings a series of tags that are not recognized by Aztec, like `<head>`, `<meta>`, `<style>` and `<body>`.

In particular, the `<body>` tag brings all the content, but since the tag is not recognized, the content is replaced with the `[BODY]` placeholder.

Ideally, we would add support for all those tags (I guess?), but it would be to handle rendering full page HTML code ready to be rendered in a browser.

As an example, this the HTML generated for a simple *bold* text pasted from Notes:
```html
<!DOCTYPE html PUBLIC \"-//W3C//DTD HTML 4.01//EN\" \"http://www.w3.org/TR/html4/strict.dtd\">
<html>
	<head>   
		<meta name=\ "Generator\" content=\ "Cocoa HTML Writer\">
		<style type=\ "text/css\">
			p.p1 {margin: 0.0px 0.0px 0.0px 0.0px; font: 17.0px \'.SF UI Text\'; color: #454545}
			span.s1 {font-family: \'.SFUIText-Bold\'; font-weight: bold; font-style: normal; font-size: 17.00pt}
		</style>
	</head>

	<body>
		<p class=\ "p1\">
			<span class=\ "s1\">bold</span>
		</p>
	</body>
</html>
```
### The proposed solution:

Instead of adding support for the tags (for now), I opted to make a check if the HTML code present the `<body>` tag, in such case we would abort parsing the HTML code and fall back to the previous methods.

This seems to work properly:
[Video here](
https://cldup.com/V0iSL-hcku.mp4)

### To test:

- Copy paste from the Notes App as shown on the video.
- Test for possible regressions: (Test instructions from #1026)
  - Make sure pasting URLs work. Try to undo the operation.
  - Make sure pasting regular text works. Try to undo the operation.
  - Run the unit tests.
  - Test pasting HTML content:
  - Open Safari, and go to the Wikipedia page for WordPress.
  - Copy any chunk of text.
  - Paste it anywhere in the editor.
  - Switch to HTML and back, and make sure the content looks the same as when you pasted it.

cc @jkmassel @SiobhyB 

PD: There seems to be some tests failing, but they are also failing on the `develop` branch. 🤔

### DIFF
--- a/Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
@@ -98,6 +98,10 @@ public class HTMLConverter {
 
 private extension HTMLConverter {
     func hasBodyNode(_ nodes: [Node]) -> Bool {
+        return hasBodyNode(ArraySlice<Node>(nodes))
+    }
+
+    private func hasBodyNode(_ nodes: ArraySlice<Node>) -> Bool {
         if nodes.isEmpty {
             return false
         }
@@ -106,8 +110,7 @@ private extension HTMLConverter {
         case let element as ElementNode where element.name == Element.body.rawValue:
             return true
         default:
-            let remainingChildren = Array(nodes.dropFirst())
-            return hasBodyNode(remainingChildren)
+            return hasBodyNode(nodes.dropFirst())
         }
     }
 }
--- a/Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
+++ b/Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
@@ -58,7 +58,22 @@ public class HTMLConverter {
         
         return attributedString
     }
-    
+
+
+    /// Check if the given html string is supported to be parsed into Attributed Strings.
+    ///
+    /// In some cases, like pasting from the Notes app, the generated HTML will have a `<body>` tag, and that
+    /// is not yet supported. In those cases is preferible to abort html parsing.
+    ///
+    /// - Parameter html: The html string to check.
+    /// - Returns: A bool value indicating if the given html string can be handled correctly.
+    ///
+    func isSupported(_ html: String) -> Bool {
+        let processedHTML = pluginManager.process(html: html)
+        let rootNode = htmlToTree.parse(processedHTML)
+        return hasBodyNode(rootNode.children) == false
+    }
+
     /// Converts an attributed string string into it's HTML string representation.
     ///
     /// - Parameters:
@@ -77,4 +92,22 @@ public class HTMLConverter {
         return pluginManager.process(outputHTML: html)
     }
     
+}
+
+// MARK: - Helpers
+
+private extension HTMLConverter {
+    func hasBodyNode(_ nodes: [Node]) -> Bool {
+        if nodes.isEmpty {
+            return false
+        }
+
+        switch nodes.first {
+        case let element as ElementNode where element.name == Element.body.rawValue:
+            return true
+        default:
+            let remainingChildren = Array(nodes.dropFirst())
+            return hasBodyNode(remainingChildren)
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
```diff
@@ -78,3 +93,21 @@ public class HTMLConverter {
...
+    func hasBodyNode(_ nodes: [Node]) -> Bool {
+        if nodes.isEmpty {
+            return false
+        }
+
+        switch nodes.first {
+        case let element as ElementNode where element.name == Element.body.rawValue:
+            return true
+        default:
+            let remainingChildren = Array(nodes.dropFirst())
```
diegoreymendez: I'd suggest to change this method slightly to avoid re-creating the array for each element dropped.

Array slices are really easy to use, and the change here would fit their purpose imho (since `dropFirst()` creates a slice).

From https://developer.apple.com/documentation/swift/arrayslice:

>The ArraySlice type makes it fast and efficient for you to perform operations on sections of a larger array. Instead of copying over the elements of a slice to new storage, an ArraySlice instance presents a view onto the storage of a larger array. And because ArraySlice presents the same interface as Array, you can generally perform the same operations on a slice as you could on the original array.

You can reformulate the code like so:

```swift
    func hasBodyNode(_ nodes: [Node]) -> Bool {
        return hasBodyNode(ArraySlice<Node>(nodes))
    }
    
    private func hasBodyNode(_ nodes: ArraySlice<Node>) -> Bool {
        if nodes.isEmpty {
            return false
        }
        
        switch nodes.first {
        case let element as ElementNode where element.name == Element.body.rawValue:
            return true
        default:
            return hasBodyNode(nodes.dropFirst())
        }
    }
```

Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
```diff
@@ -58,7 +58,22 @@ public class HTMLConverter {
...
+    /// In some cases, like pasting from the Notes app, the generated HTML will have a `<body>` tag, and that
+    /// is not yet supported. In those cases is preferible to abort html parsing.
+    ///
+    /// - Parameter html: The html string to check.
+    /// - Returns: A bool value indicating if the given html string can be handled correctly.
+    ///
+    func isSupported(_ html: String) -> Bool {
+        let processedHTML = pluginManager.process(html: html)
+        let rootNode = htmlToTree.parse(processedHTML)
+        return hasBodyNode(rootNode.children) == false
```
diegoreymendez: Wondering a bit here but... wouldn't it be better to use the elements inside the `body` tag instead of marking the HTML as not-supported?

Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
etoledom: That's a great improvement! Thank you!

Aztec/Classes/NSAttributedString/Conversions/HTMLConverter.swift
etoledom: That's an interesting idea!

I believe that if we continue processing HTML, we still have problems when pasting from the Notes app. Basically we will lose all the stylings unless we have a way to take into account the style tags inside the header tag.

I was trying something similar on a first approach to solve this issue and I got stuck on that problem.

diegoreymendez: Then let's move ahead as it is.  I think it's more important to have a working solution right now.  Good call.

PR #1070, opened 2018-10-16
Proposal to fix pasting embeds

### LINKED ISSUE #589: Add support for Embeds
As reported by @rachelmcr (by an original report from @kwonye):

> I tested some oEmbeds and shortcodes in Aztec, and found that oEmbeds in Aztec for iOS don’t seem to be working. When I paste in a plain URL such as a YouTube or Twitter link, the URL appears in my post instead of converting to an embed. (The same URLs embedded when pasted into Aztec on Android or in a shortcode.)

We should make sure we support all the popular embeds.

This ticket requires a prior check against Aztec iOS for consistency.

jkmassel:
Currently, when pasting a URL that could be transformed into an embed (such as a YouTube link), it’ll be turned into an `<a>` tag instead, which is frustrating for several users we're talking to. This solves the issue in the simplest possible way, which may not be the best long-term, so this should probably be treated as a proof-of-concept for a fix, rather than the fix itself (unless everyone likes the direction). 

This pull request pre-empts the URL transformation of links if there’s no text selected (for instance, the user might actually want to make some text like “click here” be the link to the YouTube video, and we want the link transformation to work in that case). This allows a pasteboard that contains only a URL to be evaluated for whether it can be embedded, and if it can, it does. If not, it'll paste it as a regular link.

This PR implements support for YouTube and Vimeo embeds. The full list is a lot longer (https://codex.wordpress.org/Embeds), but it shouldn’t be difficult to implement each – it’ll just take time to write all the detection logic. 

Fixes #589 (partially)

**To test:**

Open "Empty Demo"
Paste in a YouTube or Vimeo link

**Before**
Turns into an `<a>` tag.

**After**
Just pastes the url on its own line – this'll allow WordPress to turn the link into an embed.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -226,7 +226,7 @@ open class TextView: UITextView {
     }
     
     public func load(_ plugin: Plugin) {
-        pluginManager.load(plugin)
+        pluginManager.load(plugin, in: self)
     }
 
     // MARK: - TextKit Aztec Subclasses
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -123,6 +123,11 @@ public protocol TextViewFormattingDelegate: class {
 //
 public protocol TextViewPasteboardDelegate: class {
 
+    /// Called by the TextView when it's attempting to paste the contents of the pasteboard.
+    ///
+    /// - Returns: True if the paste succeeded, false if it did not.
+    func tryPasting(in textView: TextView) -> Bool
+
     /// Called by the TextView when it's attempting to paste a URL.
     ///
     /// - Returns: True when the paste succeeded, false if it did not.
@@ -457,13 +462,7 @@ open class TextView: UITextView {
     }
 
     open override func paste(_ sender: Any?) {
-
-        let pasteHandled = pasteboardDelegate.tryPastingURL(in: self)
-            || pasteboardDelegate.tryPastingHTML(in: self)
-            || pasteboardDelegate.tryPastingAttributedString(in: self)
-            || pasteboardDelegate.tryPastingString(in: self)
-
-        guard pasteHandled else {
+        guard pasteboardDelegate.tryPasting(in: self) else {
             super.paste(sender)
             return
         }
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -119,6 +119,30 @@ public protocol TextViewFormattingDelegate: class {
     func textViewCommandToggledAStyle()
 }
 
+// MARK: - TextViewPasteboardDelegate
+//
+public protocol TextViewPasteboardDelegate: class {
+
+    /// Called by the TextView when it's attempting to paste a URL.
+    ///
+    /// - Returns: True when the paste succeeded, false if it did not.
+    func tryPastingURL(in textView: TextView) -> Bool
+
+    /// Called by the TextView when it's attempting to paste HTML content.
+    ///
+    /// - Returns: True when the paste succeeded, false if it did not.
+    func tryPastingHTML(in textView: TextView) -> Bool
+
+    /// Called by the TextView when it's attempting to paste an attributed string.
+    ///
+    /// - Returns: True when the paste succeeded, false if it did not.
+    func tryPastingAttributedString(in textView: TextView) -> Bool
+
+    /// Called by the TextView when it's attempting to paste a string.
+    ///
+    /// - Returns: True when the paste succeeded, false if it did not.
+    func tryPastingString(in textView: TextView) -> Bool
+}
 
 // MARK: - TextView
 //
@@ -138,7 +162,12 @@ open class TextView: UITextView {
     /// Formatting Delegate: to be used by the Edition's Format Bar.
     ///
     open weak var formattingDelegate: TextViewFormattingDelegate?
-    
+
+    /// Pasteboard Delegate: Handles Cut, Copy, and Paste commands. Can be overridden
+    /// by a subclass to customize behaviour.
+    ///
+    open var pasteboardDelegate: TextViewPasteboardDelegate = AztecTextViewPasteboardDelegate()
+
     // MARK: - Behavior configuration
     
     private static let singleLineParagraphFormatters: [AttributeFormatter] = [
@@ -428,8 +457,12 @@ open class TextView: UITextView {
     }
 
     open override func paste(_ sender: Any?) {
-        let pasteHandled = tryPastingURL() || tryPastingHTML() || tryPastingAttributedString() || tryPastingString()
-        
+
+        let pasteHandled = pasteboardDelegate.tryPastingURL(in: self)
+            || pasteboardDelegate.tryPastingHTML(in: self)
+            || pasteboardDelegate.tryPastingAttributedString(in: self)
+            || pasteboardDelegate.tryPastingString(in: self)
+
         guard pasteHandled else {
             super.paste(sender)
             return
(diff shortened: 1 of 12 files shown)


diegoreymendez:
Thanks for the submission @jkmassel !

Before commenting on the PR I wanted to bring up a few architectural choices in Aztec, that are not documented anywhere right now.

### On Aztec vs WordPressEditor

When you open the project file, you'll notice there are two targets (or libraries): `Aztec` and `WordPressEditor`.

The reason why we have this distinction is that `Aztec` was designed to be a standard-HTML-editor library.  This means it does not really know (or care about) things that are not WordPress-specific concepts, such as embeds.

For that purpose we have `WordPressEditor` which is a library built on top of `Aztec` that takes care of making Aztec play nicely with WordPress posts.

This library takes care of customizations for things such as shortcodes, autop, removep, Gutenberg-formatted posts, etc.  And ideally embeds too.

### On the PR

With that in mind, it would be preferred to add this fix to `WordPressEditor` instead of `Aztec`.

On top of that, there's a mechanism that could be better suited to achieve this, through our Plugins API:

- [Caption shortcode input processor.](https://github.com/wordpress-mobile/AztecEditor-iOS/blob/develop/WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Calypso/CaptionShortcode/CaptionShortcodeInputProcessor.swift#L6)
- [Caption shortcode output processor.](https://github.com/wordpress-mobile/AztecEditor-iOS/blob/develop/WordPressEditor/WordPressEditor/Classes/Plugins/WordPressPlugin/Calypso/CaptionShortcode/CaptionShortcodeOutputProcessor.swift#L6)

It would allow us to let the editor still paste URLs as normal links, but still let us get a different HTML representation for the ones we want.

The advantages are:

- We'd still be able to use our link editing UI if we wanted to, since links would still be links except when converting the post to HTML.
- We'd be able to keep Aztec's core library as WordPress-agnostic as possible.
- We'd be a bit more aligned with the current architecture.

I apologize in advance because I realize a lot of the confusion comes from a lack of proper architectural documentation, but I still believe this PR was a good way to start moving things forward.

On the upside, I think the regexes for identifying the links we want to keep as embed can probably stay the same, even if we move towards the plugin approach.

Let me know what you think.

jkmassel:
Thanks for the phenomenal explanation of all of that @diegoreymendez!

I took a look at the caption shortcode processors you provided, but I have a couple of concerns about implementing it that way (based on my limited understanding of Aztec).

Here are a few flows I can see that could cause some trouble:

1. A user opens a post they've previously created elsewhere. That post contains a link that looks like: `<a href="http://youtu.be/12345">http://youtu.be/12345</a>`. The shortcode processors run when the content is loaded into the editor, which means that before the user does anything, we've dirtied the content (which isn't a great user experience, because they may not understand why they're prompted with a "Do you want to save your changes" message – they didn't meant to make changes). Additionally, this would make it impossible for a post to contain links in this format while still being editable by our app – simply by opening the post on mobile, the user would be changing the link to an embed, which might not be their intention.

2. The user pastes an embeddable URL into the editor, and it's auto-transformed into an `<a>` tag by Calypso. The shortcode processors don't run in response to `paste:`, which means that the transformation doesn't happen at this time. We could run the processors when the editor contents are saved, but we run into the same problem of transforming the user's content without them knowing about it.

As far as I can tell, the best time to do transformation on embedded-able URLs is as the user pastes them, then if they wish to adjust it later, they can do that as a separate operation – that avoids any surprises for the user.

==========

This is probably a good time to mention that I'm pretty new in this code, so there's a good chance I've missed something, so feel free to correct me! 😅  

But if not, I'm wondering what you would think about moving `tryPastingURL()`, `tryPastingHTML()`, `tryPastingAttributedString()`, and `tryPastingString()` from `TextView`, into an Aztec-specific class? If this class conformed to a protocol with those methods, the class could either be overridden, or an entirely new class could be created that conforms to the protocol – either would allow `WordPressEditor` to customize the functionality of that operation.

If it helps, I can update the PR with a suggestion of what that might look like, if we decide this is an actual issue.

diegoreymendez:
@jkmassel - I'll be answering soon.  I haven't forgotten! :)

I'm processing your comments to try and get the full picture (while handling priorities).

diegoreymendez:
> Thanks for the phenomenal explanation of all of that @diegoreymendez!
> 
> I took a look at the caption shortcode processors you provided, but I have a couple of concerns about implementing it that way (based on my limited understanding of Aztec).
> 

Before moving ahead, I want to mention that my suggestion to use a solution like the one we used for shortcodes, was for embeds in the following form:

```
[embed width="123" height="456"]http://www.youtube.com/watch?v=dQw4w9WgXcQ[/embed]
```

I still think that for those types of embeds, we'll want to use this conversion mechanism.

But [it seems some plain text links](https://codex.wordpress.org/Embeds) can be embeds as well, which is what you're tackling in this PR.

And pasting plain text links requires additional logic... hence this PR.

> Here are a few flows I can see that could cause some trouble:
> 
> 1. A user opens a post they've previously created elsewhere. That post contains a link that looks like: `<a href="http://youtu.be/12345">http://youtu.be/12345</a>`. The shortcode processors run when the content is loaded into the editor, which means that before the user does anything, we've dirtied the content (which isn't a great user experience, because they may not understand why they're prompted with a "Do you want to save your changes" message – they didn't meant to make changes). Additionally, this would make it impossible for a post to contain links in this format while still being editable by our app – simply by opening the post on mobile, the user would be changing the link to an embed, which might not be their intention.

Keep in mind this may not be relevant for this PR in particular, but for context...

This should not be a problem because when converting an `[embed ...]` we can either decide to use a custom element (such as `<embed x>y.</embed>`), or simply add extra data to the link tag to mark it's an embed.

> 2. The user pastes an embeddable URL into the editor, and it's auto-transformed into an `<a>` tag by Calypso. The shortcode processors don't run in response to `paste:`, which means that the transformation doesn't happen at this time. We could run the processors when the editor contents are saved, but we run into the same problem of transforming the user's content without them knowing about it.
> 

This is a bug, because if we don't run the processors, we're going to see the same problem with shortcodes.

> As far as I can tell, the best time to do transformation on embedded-able URLs is as the user pastes them, then if they wish to adjust it later, they can do that as a separate operation – that avoids any surprises for the user.
> 
> ==========
> 
> This is probably a good time to mention that I'm pretty new in this code, so there's a good chance I've missed something, so feel free to correct me! 😅
> 
> But if not, I'm wondering what you would think about moving `tryPastingURL()`, `tryPastingHTML()`, `tryPastingAttributedString()`, and `tryPastingString()` from `TextView`, into an Aztec-specific class? If this class conformed to a protocol with those methods, the class could either be overridden, or an entirely new class could be created that conforms to the protocol – either would allow `WordPressEditor` to customize the functionality of that operation.
> 
> If it helps, I can update the PR with a suggestion of what that might look like, if we decide this is an actual issue.

So going back to the scope of this PR... I agree we'll want to customize the paste logic.  This has been on my mind for a while but just never took the spotlight.

My main concern with this PR as-is, is that the fix is in the wrong location (`Aztec` instead of `WordPressEditor`).

Do you think you can work on a customization path that we can use so that this logic lives in `WordPressEditor` instead?

`WordPressPlugin` is where most of the customizations live, so I'd suggest checking there first.  And of course let me know if I can assist in any way.

jkmassel:
>Do you think you can work on a customization path that we can use so that this logic lives in WordPressEditor instead?

Yes! This PR has been updated to do this. From here, we could adjust how it works – as a for-instance, it might be nicer to turn the link into an Embed shortcode on paste, rather than just putting the URL in the document for auto conversion by WP. 

But this might be an ok starting point, and then everything else we're talking about could as one or more separate PRs?

diegoreymendez:
> But this might be an ok starting point, and then everything else we're talking about could as one or more separate PRs?

I still need to see the full PR but I'd say yes, let's keep the scope of this PR as it is.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -428,115 +457,25 @@ open class TextView: UITextView {
     }
 
     open override func paste(_ sender: Any?) {
-        let pasteHandled = tryPastingURL() || tryPastingHTML() || tryPastingAttributedString() || tryPastingString()
-        
+
+        let pasteHandled = pasteboardDelegate.tryPastingURL(in: self)
```
diegoreymendez: This is just a suggestion to go the extra mile:

Why not put this entire `try...` sequence into a main method in `pasteboardDelegate`?

This code would turn into something like:

```swift
let pasteHandled = pasteboardDelegate.paste(in: self) // or tryPaste() if you prefer
```

The advantage is that it would allow us to customize not only the logic inside each of those methods but also their priority ordering.

jkmassel: This is a really great suggestion!!

Implemented in 30be23350587ac7ce6ab09f914b7c8106ddac805

I aligned the names in the protocol for consistency as well

WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+public struct EmbedURLProcessor{
```
diegoreymendez: No need to change this for this PR, but one thing I wanted to bring up is that it'd be nice if we (at some point) changed this to be modularized in a way that we can have instances of `EmbedURLProcessor` for each platform we support (instead of having methods like `isYouTubeEmbed`.

Again, I think the current solution is awesome for the scope of this PR, but I wanted to bring up the point in case we move forward with other PRs.

jkmassel: That sounds great – might make sense to even do that once we have a few platforms in there and can get a feel for how it is working with a longer list – the list supported by WP is _very_ long , and this architecture may not scale that well 😅.

jkmassel:
I added a test case and fix for links with underscores – seemed like a shame to come this far and not add that small fix :)

There's a new test case in there, and everything else passes, so I feel safe merging this with that change added.

PR #1082, opened 2018-11-13
Non breaking spaces are no longer removed by WordPressEditor.

diegoreymendez:
### Description:

Non breaking spaces were being removed by `WordPressEditor` and were not being properly encoded when switching back to HTML mode in Aztec.

### Testing:

1. Open either the Aztec or WordPressEditor empty demo.
2. Switch to HTML mode.
3. Write several `&nbsp;`s, in multiple lines, or one after the other.
4. Switch to visual mode and back to HTML mode.

Make sure the `&nbsp;`s are still there and properly encoded for HTML.

Props to @charliescheer for finding this issue and reporting it.

### DIFF
--- a/Aztec/Classes/Extensions/String+HTML.swift
+++ b/Aztec/Classes/Extensions/String+HTML.swift
@@ -33,8 +33,9 @@ extension String {
     ///
     private var entities: [Entity] {
         return [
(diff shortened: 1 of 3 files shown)


diegoreymendez:
Thanks @etoledom !

PR #1088, opened 2018-11-20
Undo is no longer reset when switching editing modes.

diegoreymendez:
### Description:

This PR fixes the undo stack being reset when switching editing modes.

undoredo

### Additional Information:

By resolving this issue we will be able to properly record when the content has changes by asking the undo manager directly.

### Known limitations:

The undo stack is different in visual mode and HTML mode.  Atomic edits are only recorded in the mode you're in when they're registered.

The editing mode you're not in will only record a big change when switching.

### Testing:

**Test 1:**

- Open any of the editor demos with content.
- Undo

Make sure the initial content is never removed.

**Test 2:**

- Open an empty editor demo.
- Write whatever you want.
- Copy and paste content.
- Apply styles.
- Switch to HTML mode.
- Do some more edits.
- Switch back to visual.
- Undo as much as you want, and Redo.

Make sure it all works.

### DIFF
--- a/Aztec/Classes/EditorView/EditorView.swift
+++ b/Aztec/Classes/EditorView/EditorView.swift
@@ -79,7 +79,7 @@ public class EditorView: UIView {
                 
                 htmlTextView.becomeFirstResponder()
             case .richText:
-                richTextView.setHTMLUndoable(htmlTextView.text)
+                richTextView.setHTMLUndoable(htmlTextView.text)                
                 richTextView.becomeFirstResponder()
             }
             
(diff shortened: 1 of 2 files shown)


diegoreymendez:
@etoledom - Thanks for the review.  That's a bug indeed.

diegoreymendez:
@etoledom - Ready for another look.  I tried a cleaner approach, the one I propose here is simpler and seems to work fine.

diegoreymendez:
Thank you for the thorough review @etoledom.

PR #1092, opened 2018-11-22
Fix blockquote cite's removal.

diegoreymendez:
### Description:

This moves forward #1090.  The issue is not completely solved as it's a complex combination of several issues.  This PR just fixes an initial part of it.

This PR makes sure `<cite>` elements are removed when the blockquote style is removed.

### Demo:

blockquoteciteissue

### Testing:

1. Open the empty standard HTML demo.
2. Switch to HTML mode.
3. Paste:

```html
<blockquote>
   <p>Some quote</p>
   <cite>by Me</cite>
</blockquote>
```

4. Switch back to visual mode and toggle the blockquote in the cite line.
5. Switch back to HTML mode and make sure the `<cite>` tag is removed.


diegoreymendez:
Thanks @etoledom!

There are indeed more problems coming up, and I'll be tackling them one by one.

PR #1093, opened 2018-11-22
Fix persistence of custom string attributes.

diegoreymendez:
### Description:

Moves forward #1090.

This PR fixes part of the logic that takes care of persisting custom attributes while typing.  Previously we were forcing the typing attributes to be maintained while typing, which means there was no way to turn some styles off (such as `<cite>` and `<code>`).

### Demo:

styleremoved

### Known limitations:

While this improves how those styles are maintained as we type and move through the document, there are still issues with how blockquotes are expanded in visual mode:

- Gutenberg blockquotes are still breaking.
- Multi-line cites still need fixing.
- While cite is currently being removed, the italic styling it enables is not.

### Testing:

1. Open the empty standard HTML editor.
2. Switch to HTML mode.
3. Paste:

```html
<blockquote>
   <p>Some quote</p>
   <cite>by Me</cite>
</blockquote>
```

4. Switch to visual mode.
5. Place the caret in the extra-line (the last whiteline without styling).
6. Type and make sure the blockquote and cite styling is removed.

### DIFF
--- a/Aztec/Classes/Extensions/NSAttributedString+Analyzers.swift
+++ b/Aztec/Classes/Extensions/NSAttributedString+Analyzers.swift
@@ -6,22 +6,6 @@ import UIKit
 //
 extension NSAttributedString {
 
-    /// Returns true if the text preceding a given location contains the specified attribute key.
-    ///
-    /// - Parameters:
-    ///   - location: location to check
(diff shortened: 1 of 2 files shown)


diegoreymendez:
Thanks @jleandroperez !

PR #1094, opened 2018-11-26
Blockquote + cite is now ended when pressing ENTER.

diegoreymendez:
### Description:

Moves forward #1090.

Pressing enter while in visual-mode, in a line that has both `<blockquote>` and `<cite>`, will cause the blockquote to be closed and the styles removed.

If the user wants a multi-line citation, they can use the SHIFT + ENTER combo.

### Details:

The code changes are not as clean as I'd like, because our current way to access formatter causes a lot of code duplication, but this is a quick fix to keep the ball rolling.

### Demo:

blockquoteandciteclosed

### Testing:

Follow the steps shown in the demo above.
Run the unit tests.


diegoreymendez:
Thanks @etoledom !

PR #1096, opened 2018-11-28
Remove the placeholderAttributes formatter property.

diegoreymendez:
### Description:

Formatters had a property named `placeholderAttributes` which once had a purpose, but currently has none.

This is a small maintenance PR that will be followed up with a simplification of formatters.  This is a small step towards fixing #1090.

To do so, I'll unify all registered formatters in a single spot in `Aztec.TextView`.

### Testing:

Make sure the example app runs fine.
Make sure the unit tests run fine.


diegoreymendez:
Thanks @etoledom

PR #1097, opened 2018-12-04
Fixes the disappearing toolbar bug.

### LINKED ISSUE #1067: FormatBar disappears after selecting header or list style
Original report by @ CivelXu in https://github.com/wordpress-mobile/WordPress-iOS/issues/10221, and reproduced in iPhone X simulator.

### Expected behavior

FormatBar is always visible

### Actual behavior

FormatBar disappears

### Steps
1. Select FormatBar header button or list button.
2. Select a header or list style.

### Screenshot
image

diegoreymendez:
### Description:

Fixes #1067.
This fix will also (most likely) solve https://github.com/wordpress-mobile/WordPress-iOS/issues/9208 once we integrate these changes into WPiOS.

### Details:

Changes:

- Wraps `OptionsTableViewController` inside a `UIInputViewController` when presenting the options as a keyboard replacement.
- Moves the Options table presentation logic to WordPressEditor, so we can use the same code once we integrate this into WPiOS without having to duplicate the changes.

### Testing:

**Test 1:**

This test covers the original issue reported in #1067.

Run the demo app using your device with iOS 12.
While editing pick a header style or a list style.
Once selected make sure the toolbar is still visible.

**Test 2:**

Free-Play with the toolbar!  Make sure the options are hidden and shown when it makes sense.  The behavior should not have changed.


diegoreymendez:
Fyi @rachelmcr and @elibud.  Once this is merged I'll integrate the changes into WPiOS.

At that point the toolbar issues and the issues with text going behind the keyboard should not longer happen.

diegoreymendez:
I've had to update .travis.yml to use Xcode 10.1 since it was showing some errors that I don't get locally.  Hoping that'll produce a working build with Travis.

diegoreymendez:
@etoledom - The issue I fixed is similar but had a different cause.

I've just submitted a commit that should fix the issue you reported, which was honestly really nasty.

Nice catch.

diegoreymendez:
Thanks @etoledom !

PR #1100, opened 2018-12-06
Don’t create ranges intersecting grapheme clusters

### LINKED ISSUE #1098: Crash: Aztec String+RangeConversion crash
Mirrored from: https://github.com/wordpress-mobile/WordPress-iOS/issues/10548

jkmassel:
Fixes #1098 

**To test:**

Run the unit tests, before and after!


diegoreymendez:
The logic seems mostly fine, but the naming of the methods seems off since Swift `String` ranges don't really correspond with UTF8 encoding ranges.

Try this in a playground:

```swift
let testString = "🌏"

print(testString.count)
print(testString.utf8.count)
```

Swift String ranges seem to always correspond exactly with one full visible character, so in a way that explains why UTF16 ranges don't always have a corresponding equivalent Swift String range.

My suggestion would be to remove the UTF8 reference from the method names and documentation.  For example:

```swift
private func findValidUTF8LowerBound(for utf16Range: Range<String.UTF16View.Index>) -> String.Index {
```

...could become...

```swift
private func findValidLowerBound(for utf16Range: Range<String.UTF16View.Index>) -> String.Index {
```

jkmassel:
Thanks @diegoreymendez!!

This is updated! :)

PR #1101, opened 2018-12-11
End Gutenberg blocks on double enter.

diegoreymendez:
### Description:

This PR offers a partial / an initial solution to https://github.com/wordpress-mobile/WordPress-iOS/issues/10465

Pressing enter twice at the end of a blockquote should end the Gutenberg blockquote block as well.  This is actually true for any Gutenberg block type.

### Details:

TBD


diegoreymendez:
This is ready for review.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -309,17 +309,21 @@ open class TextView: UITextView {
...
+    override open var typingAttributes: [String : Any] {
         get {
-            ensureRemovalOfParagraphAttributesAfterSelectionChange()
-            return super.typingAttributes
+            return myTypingAttributes
         }
+        
         set {
+            // We're still setting the typing the paren't typing attributes in case it's used directly
+            // by any iOS feature.
```
etoledom: Small detail: Seems to be something wrong with the redaction of the comment. Could you double check it? :)

diegoreymendez:
> Testing the blockquote with a cite tag works, but the new text that is written outside the block is still inside a cite tag:

Actually this was sort of unintended, so I'll try to get this fixed as part of this PR.

> As you can see in the snippet above, smart dashes are being really painful! And very good to mess things up. From that snippet, just switching to visual mode and back to html view, I get this result:

This is correct but not introduced by this PR, so I'll just tackle that separately.

There seems to be an iOS 12 bug somewhere that's breaking smart dashes, but I have to debug to say for sure.

Aztec/Classes/TextKit/TextView.swift
diegoreymendez: Addressed.

diegoreymendez:
@etoledom - This is ready for another look.

Some details of the changes I did:

1. It seems the double dash issue was caused by [this call](https://github.com/wordpress-mobile/AztecEditor-iOS/blob/develop/Aztec/Classes/EditorView/EditorView.swift#L77), which for some strange reason seems to ignore the smart dash config.
2. I added specific initialization lines to turn off smart dashes and quotes in `EditorView`.  I did check though and they were off before too.  I validated that these didn't fix the issue but I still think it's safer to keep them.
3. I reintroduced the removal of paragraph properties through our formatters when pressing double enter.  Formatters are important because they may be able to remove things that are not paragraph properties, but still should be removed when one is (an example is `<cite>`, since it's not a block-level element but should be removed when the blockquote is ended).

There's obviously still more work to do here, but I think the PR should be now in a better position to move forward and keep tackling issues.

PS: I think I'll release a new version of Aztec after these changes since the double dash issue is something we want out there ASAP.

PR #1102, opened 2018-12-11
Fixes a bug in the paragraph ranges calculation.

diegoreymendez:
### Description:

This PR fixes a bug in the paragraph ranges calculation.

It also adds 3 unit tests to make sure this issue does not regress.

In this PR I also move some methods to a more appropriate `NSAttributedString+ParagraphRange.swift` extension, since they were previously in an extension that had little to do with their purpose.

### Details:

The issue was that before calculating the paragraph ranges intersecting a specified subrange, we weren't expanding the subrange to paragraph bounds.  If the provided subrange was zero-lenght, this method wasn't returning any paragraph, even if the selection location was within an existing paragraph.

Let me know if any of this needs clarification.

### Note:

These fixes were broken off from a larger PR that's coming up, to make revision and verification easier.

### Testing:

The issue can be reproduced by reverting the changes to the code and running the provided unit tests.  
The fix can be tested by running the unit tests with all the provided changes.


AztecTests/Extensions/NSAttributedStringParagraphRangeTests.swift
```diff
@@ -0,0 +1,40 @@
+import XCTest
+@testable import Aztec
+
+class NSAttributedStringParagraphRangeTests: XCTestCase {
+    
+    func testParagraphRangesWithoutSeparator() {
```
jkmassel: WDYT about appending `SpanningZeroPointRange` (or something similar) to each these test names? I'd imagine in the future we may have other tests that ensure everything works fine for strings with paragraphs spanning ranges greater than zero?

Aztec/Classes/Extensions/NSAttributedString+ParagraphRange.swift
```diff
@@ -0,0 +1,184 @@
...
+    /// Finds the paragraph ranges in the specified string intersecting the specified range.
+    ///
+    /// - Parameters range: The range within the specified string to find paragraphs.
+    ///
+    /// - Returns: An array containing an NSRange for each paragraph intersected by the specified range.
+    ///
+    func paragraphRanges(spanning range: NSRange) -> ([(NSRange, NSRange)]) {
+        var paragraphRanges = [(NSRange, NSRange)]()
+        let swiftRange = string.range(fromUTF16NSRange: range)
+        let paragraphsRange = string.paragraphRange(for: swiftRange)
```
jkmassel: I was surprised that 
```swift
string.paragraphRange(for: swiftRange)
```
returns the valid ranges for the entire string, even if the initial range is empty!

I'm thinking that might be worth adding to the inline docs for the parent method? Just reading the unit tests, I'd expect that given a zero point range, we'd get back no matching paragraphs.

Aztec/Classes/Extensions/NSAttributedString+ParagraphRange.swift
```diff
@@ -0,0 +1,184 @@
...
+        }
+        
+        return paragraphRanges
+    }
+    
+    /// Finds the paragraph ranges in the specified string intersecting the specified range.
+    ///
+    /// - Parameters range: The range within the specified string to find paragraphs.
+    ///
+    /// - Returns: An array containing an NSRange for each paragraph intersected by the specified range.
```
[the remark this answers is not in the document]
diegoreymendez: Updated.

Aztec/Classes/Extensions/NSAttributedString+ParagraphRange.swift
```diff
@@ -0,0 +1,184 @@
...
+    ///
+    func paragraphRanges(spanning range: NSRange) -> ([(NSRange, NSRange)]) {
```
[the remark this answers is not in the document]
diegoreymendez: Sounds like a good idea.  Changed!

Aztec/Classes/Extensions/NSAttributedString+ParagraphRange.swift
diegoreymendez: In terms of the methods we implemented I changed the naming to make them clearer.

The new method name and documentation should make it much clearer what's going on:

```swift
    /// Given a range within the receiver, this method returns an array of ranges for each
    /// paragraph that intercects the provided range.
    ///
    /// - Parameters:
    ///     - range: The initial range that paragraphs must intersect to qualify as valid results.
    ///
    /// - Returns: An array of `ParagraphRange` objects describing the range and the enclosing range
    ///     of each paragraph that intersects the input range.
    ///
    func paragraphRanges(intersecting range: NSRange) -> ([ParagraphRange]) {
```

Regarding the snippet you mention, since that method is not defined / maintained by us, but by Apple I'd say it's up to the reader to know how the method works (as with any other call to standard methods).

Agreed, it's not properly documented in [`StringProtocol`](https://developer.apple.com/documentation/swift/stringprotocol/2923378-paragraphrange), but it works exactly like its [NSString](https://developer.apple.com/documentation/foundation/nsstring/1408548-paragraphrange) counterpart.

AztecTests/Extensions/NSAttributedStringParagraphRangeTests.swift
diegoreymendez: Added documentation to each test, and a more precise name to describe what's being tested.

diegoreymendez:
@jkmassel - All feedback addressed.  Can you let me know what you think?

Thanks!

diegoreymendez:
Thanks @jkmassel !

PR #1104, opened 2018-12-12
Fixes a bug that was causing some headers to disappear.

diegoreymendez:
### Description:

Fixes an issue that was causing empty block-level elements to be removed.

### Testing:

To reproduce the original issue, you can test this in `develop`.  To test the fix, run the same steps in this branch:

1. Launch the editor
2. Switch to HTML mode and paste this:

```html
<h1></h1>
<h2></h2>
Hello there
```

3. Switch back to visual mode, don't do any edits.
4. Switch back to HTML.

The header tags should still be there.


diegoreymendez:
Thanks @etoledom !

PR #1105, opened 2018-12-12
Feature/add support for more embeds

jkmassel:
Adds support for embeds from:
- Twitter
- Flickr
- Issu
- Instagram
- Facebook

Also adds support for YouTube playlists

### DIFF
--- a/WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
+++ b/WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
@@ -9,7 +9,7 @@ public struct EmbedURLProcessor{
     }
 
     public var isValidEmbed: Bool{
-        return isYouTubeEmbed || isVimeoEmbed
+        return isYouTubeEmbed || isVimeoEmbed || isTwitterEmbed
     }
 
     /// Tests the url to see if it's a valid YouTube URL.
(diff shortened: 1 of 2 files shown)


WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
```diff
@@ -49,6 +57,74 @@ public struct EmbedURLProcessor{
...
+            pattern("https?://www\\.facebook\\.com/notes/[\\S]+$"),                       // Note
+            pattern("https?://www\\.facebook\\.com/.*/videos/[\\S]+$"),                   // Video
+        ])
+    }
+
+    private func pattern(_ pattern: String) -> NSRegularExpression {
+        return try! NSRegularExpression(pattern: pattern, options: [.caseInsensitive])
+    }
+
+    private func matchesAnyOf(_ regexes: [NSRegularExpression]) -> Bool{
```
diegoreymendez: IMHO a more "Swifty" name for this method would be `matches(anyOf:)`, since the "matches" part refers to the nature of the method, and "any of" refers specifically to the first input parameter.

WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
```diff
@@ -49,6 +57,74 @@ public struct EmbedURLProcessor{
...
+    public var isFacebookEmbed: Bool {
+        return matchesAnyOf([
+            pattern("https?://www\\.facebook\\.com/.*/posts/[\\S]+$"),                    // Post
+            pattern("https?://www\\.facebook\\.com/.*/photos/[\\S]+$"),                   // Photo
+            pattern("https?://www\\.facebook\\.com/notes/[\\S]+$"),                       // Note
+            pattern("https?://www\\.facebook\\.com/.*/videos/[\\S]+$"),                   // Video
+        ])
+    }
+
+    private func pattern(_ pattern: String) -> NSRegularExpression {
```
diegoreymendez: This is very subjective, I realize... but I think this method would be clearer to read if it was named `expression(with string:)` or `regulardExpression(with string:)`.

Mostly because these names would exactly describe what the method does / returns.

WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
jkmassel: Nice! That's a way better way to do it :) 

Updated!

WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
```diff
@@ -9,7 +9,16 @@ public struct EmbedURLProcessor{
     }
 
     public var isValidEmbed: Bool{
-        return isYouTubeEmbed || isVimeoEmbed
+
+        return [
```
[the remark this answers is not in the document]
jkmassel: Just that I wanted each on its own line so the diffs would be cleaner, and didn't think of your suggested format 😜 

Updated!

WordPressEditor/WordPressEditor/Classes/Processors/EmbedURLProcessor.swift
diegoreymendez: `regex` reads great to me.

PR #1127, opened 2019-01-30
Make heading styling distinguishable from paragraph

pinarol:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/351

This PR updates the style of Heading block based on https://github.com/wordpress-mobile/gutenberg-mobile/issues/351#issuecomment-456522533 

WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/10912
mobile-gutenberg PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/525

**Changes Made**
1. Heading font sizes are updated
2. Headings made bold by default
3. ShadowBoldFormatter is added to mimic bolding on already bold headings, it uses shadow attribute to mimic the extra boldness.

**The Problem**
We want to use bold font with Heading blocks by default. However, we don't want to generate bold output in the html unless user intentionally selects B option from toolbar. That makes it a bit challenging for us to decide how to make this happen. This PR proposes a solution that mimics bolding behavior by giving extra shadow and letter spacing.

**To Test**

1. Example app

1.1. Test Heading

- Verify that the font looks bolder than it was before
- Select part of the text and tap B button
- Verify that the selected area gets bolder
- Switch between H2 H3 H4 types of heading sizes
- Verify that the boldness of selected area is kept

1.2 Test paragraph

Test that paragraph is not effected:
- Paragraphs should continue to appear non-bold by default
- Select part of the text and tap B button
- Verify that the selected area gets bolder

2. Test with Gutenberg and Classic editor by Test Steps in WPiOS PR
https://github.com/wordpress-mobile/WordPress-iOS/pull/10912

### DIFF
--- a/Aztec/Classes/Formatters/Implementations/ShadowBoldFormatter.swift
+++ b/Aztec/Classes/Formatters/Implementations/ShadowBoldFormatter.swift
@@ -12,11 +12,12 @@ class ShadowBoldFormatter: AttributeFormatter {
         }
         
         static let blurRadiusNoBlur: CGFloat = 0.0
+        static let defaultColor: UIColor = .black
         
         // Creates a no blur NSShadow instance with given offset values
         static func shadow(width: CGFloat = DefaultOffset.width,
                            height: CGFloat = DefaultOffset.height,
-                           color: UIColor = UIColor.black) -> NSShadow {
+                           color: UIColor) -> NSShadow {
             let shadow = NSShadow()
             shadow.shadowBlurRadius = Shadow.blurRadiusNoBlur
             shadow.shadowOffset = CGSize(width: width, height: height)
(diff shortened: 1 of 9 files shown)


pinarol:
hey @diegoreymendez 👋  @iamthomasbishop will review the new style on the device and will provide some style feedbacks but I think technical side of the PR is ready to review.

Aztec/Classes/Formatters/Implementations/ShadowBoldFormatter.swift
```diff
@@ -0,0 +1,90 @@
...
+            static let width: CGFloat = 0.75
+            static let height: CGFloat = 0.0
+        }
+        
+        static let blurRadiusNoBlur: CGFloat = 0.0
+        
+        // Creates a no blur NSShadow instance with given offset values
+        static func shadow(width: CGFloat = DefaultOffset.width,
+                           height: CGFloat = DefaultOffset.height,
+                           color: UIColor = UIColor.black) -> NSShadow {
```
pinarol: @diegoreymendez 👋 The color choice is sth I want to ask, I couldn't find where to look for to see the what determines the color of headings? So I am thinking choosing black can be problematic in the future, shadow should actually be same as the font color, but not sure how to do it ? any help is appreciated

Aztec/Classes/TextKit/ParagraphProperty/Header.swift
```diff
@@ -98,7 +98,7 @@ open class Header: ParagraphProperty {
 //
 private extension Header {
     struct Constants {
-        static let defaultFontSize = Float(14)
+        static let defaultFontSize = Float(16)
```
pinarol: the old value of default font size corresponded to h5 so I updated this as h5's new value also

Aztec/Classes/Formatters/Implementations/ShadowBoldFormatter.swift
diegoreymendez: I'd stick to the font color... and I'd probably avoid a default (in the method definition) if we can.

Considering both calls to this method have access to the string attributes, I'd pick up the color from there (and default there to black if the color is missing).

pinarol:
hey @diegoreymendez 👋 I have updated the color issue.

Issue #1125, opened 2019-01-23
Conform to Community Standards


stevebaranski:
To promote consistency across `WordPress-iOS` frameworks & to conform to [recommended community standards](https://github.com/wordpress-mobile/AztecEditor-iOS/community), we should update this repository to include all of the following:

- [X] Description
- [X] README
- [x] Code of conduct
- [X] Contributing
- [X] License
- [x] Issue templates
- [X] Pull request template

We should also modify the PR checks we have in place for consistency with `WordPress-iOS` (e.g., build checks, approval / merge requirements, etc.).

diegoreymendez:
This can be closed.  If there's any specific change required let's track it through a new issue.

PR #1134, opened 2019-02-12
Adds the Code of Conduct to Aztec.

diegoreymendez:
## Description:

Adds the same Code of Conduct we're using in WordPress-iOS, to Aztec.

Moves ahead #1125


diegoreymendez:
Thank you for the review @stevebaranski .

Issue #1137, opened 2019-02-18
Bold tag duplication issue.


diegoreymendez:
### Issue

HTML content with a `b` tag is causing the output to contain both a `b` and a `strong` tag.

### Steps to reproduce the behavior

![strongbold](https://user-images.githubusercontent.com/1836005/52953679-eac57280-3366-11e9-9eca-1643c84dec5e.gif)

PR #1138, opened 2019-02-20
Fix bold converting issue.

diegoreymendez:
## Description:

There was an issue with a bold converter that was causing `b` tags to convert to a combination of `b` and `strong` tags.  This PR fixes that.

Fixes https://github.com/wordpress-mobile/AztecEditor-iOS/issues/1137

Additionally, this PR simplifies the `equivalentNames` calculated properly to work directly with `Elements` and to remove some extra code that wasn't needed.

### Test 1:

Run the unit tests.

### Test 2:

1. Open the empty editor demo.
2. Switch to HTML mode.
3. Type `<b>Hello!</b>`.
4. Switch to visual and back to HTML mode.

Make sure the bold tag hasn't been duplicated with a `strong` tag.

### DIFF
--- a/Aztec/Classes/Converters/ElementsToAttributedString/Implementations/GenericElementConverter.swift
+++ b/Aztec/Classes/Converters/ElementsToAttributedString/Implementations/GenericElementConverter.swift
@@ -233,7 +233,7 @@ private extension GenericElementConverter {
         let equivalentNames = element.type.equivalentNames
         
         for (key, formatter) in elementFormattersMap {
-            if equivalentNames.contains(key.rawValue) {
+            if equivalentNames.contains(key) {
                 return formatter
             }
         }
--- a/Aztec/Classes/Converters/StringAttributesToAttributes/Implementations/BoldStringAttributeConverter.swift
+++ b/Aztec/Classes/Converters/StringAttributesToAttributes/Implementations/BoldStringAttributeConverter.swift
@@ -7,7 +7,8 @@ import UIKit
 ///
 open class BoldStringAttributeConverter: StringAttributeConverter {
     
-    let cssAttributeMatcher = BoldCSSAttributeMatcher()
+    private let cssAttributeMatcher = BoldCSSAttributeMatcher()
+    private let defaultBoldElement = Element.strong
     
     public func convert(
         attributes: [NSAttributedStringKey: Any],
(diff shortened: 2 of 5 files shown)


diegoreymendez:
/fyi @pinarol - I think this may fix an issue with Gutenberg were `bold` tags were being converted to `strong` tags.

SergioEstevao:
@diegoreymendez this is working great for the ```<strong>``` and ```<b>``` do you want to do the same for ```<i>``` and ```<em>```

SergioEstevao:
And will it possible to add some unit tests to validate this scenarios?

diegoreymendez:
@SergioEstevao - Will do, but in a different PR in order to better split the scope of our changes.

pinarol:
hey @diegoreymendez I just did some simple tests with this PRs branch on the gb-example app and there are a few things that got my attention, just wanted to ask you.

1. The B button on the toolbar isn't becoming active when we use <b></b> tags but the bold effect is there, so this looks a bit odd, wdyt?
2. If I tap B button on toolbar the <b></b> tags are replaced by <strong> tags, were you referring to this with this sentence?
> think this may fix an issue with Gutenberg were `bold` tags were being converted to `strong` tags.

bold-tag

These might be separate issues, just wanted to let you know.

diegoreymendez:
@pinarol - Regarding the first issue indicated, that's a bug / limitation in the Gutenberg codebase which can be reproduced in web as well.

strongbold

In fact the definition of bold in Gutenberg refers specifically to the `strong` tag only:

```javascript
export const bold = {
	name,
	title: __( 'Bold' ),
	tagName: 'strong',
	className: null,
	edit( { isActive, value, onChange } ) {
		const onToggle = () => onChange( toggleFormat( value, { type: name } ) );

		return (
			<Fragment>
				<RichTextShortcut
					type="primary"
					character="b"
					onUse={ onToggle }
				/>
				<RichTextToolbarButton
					name="bold"
					icon="editor-bold"
					title={ __( 'Bold' ) }
					onClick={ onToggle }
					isActive={ isActive }
					shortcutType="primary"
					shortcutCharacter="b"
				/>
				<RichTextInputEvent
					inputType="formatBold"
					onInput={ onToggle }
				/>
			</Fragment>
		);
	},
};
```

2. Yep, I think this PR will prolly fix that.  I'm seeing that `b` tags are no longer being converted to `strong` tags in my tests.

diegoreymendez:
Merging even though travis isn't reporting progress as the last commit is checked ✅ .

This has also been reviewed twice, and I've made sure to run the unit tests and do some manual testing as well.

PR #753, opened 2017-10-05
iOS11: Format Bar Glitch

jleandroperez:
### Details:
In this PR we're fixing a glitch in which the Format Bar would loose sync with the actual Attributed String styles.

Fixes #748
cc @diegoreymendez 

### To test:
1. Empty document
2. Toggle bold
3. Type `Bold` + Newline
4. Toggle italics
5. Type `Italics`

As a result, after typing the first character (Step 5), the *Italics* format bar indicator does NOT get dehighlighted.

PR #752, opened 2017-10-05
iOS 11: Delete Backwards Workaround

jleandroperez:
### Details:
iOS 11's `insertText` / `deleteBackward`'s inner workings result into loosing the typingAttributes. This fix is analog to the one introduced via #725.

Fixes #749

cc @diegoreymendez 
Thanks in advance!

### To test:
1. Verify that the unit tests pass
2. Launch Aztec > Empty document
3. Toggle Bold
4. Type Here + Newline
5. Toggle Italics
6. Type Theree + Backspace twice
7. Type e
8. Verify that the new `e` character is in italics.

PR #1141, opened 2019-02-25
Removes iOS 11 formatting workarounds

diegoreymendez:
## Description:

Removes several workarounds we had added for iOS 11 support.

This PR also fixes a unit test that had bad logic (that was being "hidden" by the iOS 11 workarounds).

This PR should offer us the advantage of improving performance in Aztec slightly, since we're running less code at each keystroke.

## Related Issues:

Format bar style desynch: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/753
Delete Backwards workaround: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/752
Losing styles after typing 1 character: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/725

### Test 1:

1. Download the oldest iOS 11 simulator you can find.
2. Run the unit tests.
3. Run the app and test the issues reported above.

### Test 2:

Repeat test 1 in iOS 12.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -613,20 +613,7 @@ open class TextView: UITextView {
 
         ensureRemovalOfLinkTypingAttribute(at: selectedRange)
 
-        // WORKAROUND: iOS 11 introduced an issue that's causing UITextView to lose it's typing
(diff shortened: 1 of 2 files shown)


diegoreymendez:
Thanks for the review!

PR #1146, opened 2019-02-26
Fix update of typing attributes when moving range.

### LINKED ISSUE #1140: Format bar update issues.
## Description:

Some styles are not properly updated by the format bar, when moving the caret back.

styleinformatbar

SergioEstevao:
Fixes #1140 

Fix the update of typing attributes the moment selection is changed.

To test:
 - Open the demo app
 - Move the cursor/caret position along different styles (bold, italic, link, underlines) etc.
 - Check if the update of style on the format bar happens the moment you the cursors moves to a new style position.

@diegoreymendez do you mind to review this? Feel free to do any changes you feel are adequate to it.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -692,7 +692,7 @@ open class TextView: UITextView {
     ///
     public func setHTMLUndoable(_ html: String) {
         replace(storage.rangeOfEntireString, withHTML: html)
-        recalculateTypingAttributes()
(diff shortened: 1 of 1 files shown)


SergioEstevao:
@diegoreymendez do you have time to check this?

diegoreymendez:
@SergioEstevao - Since you mentioned it'd be fine to make adjustments, I expanded a bit on the documentation and added a convenience method in the case where we want to recalculate the typing attributes for the current selection range.

Hope you don't mind.

Very nice job!  This was working as intended from the start, and made it clear why the toolbar was failing.

PR #1147, opened 2019-03-01
Fix oblique type bug for Chinese

shengyang998:
The bug is: When you type Chinese (or other language that has not `italic` type) and English at the same line with `italic` format enable, and then when you delete all the English characters, the `italic` format button won't be able to hit again. 

To test:
Just type Chinese and English with `italic` open and delete all the English characters. 
After doing so, try to hit the `italic` format button again and you will see the different.

### DIFF
--- a/Aztec/Classes/Extensions/UIFont+Traits.swift
+++ b/Aztec/Classes/Extensions/UIFont+Traits.swift
@@ -30,6 +30,15 @@ extension UIFont {
             // This is a very viable scenario, and our default handling mechanism for it is to
             // return the original, unmodified font.
             //
+            
+            // MARK: Fix oblique type for Chinese
+            if newTraits.contains(.traitItalic) && familyDescriptor.postscriptName.hasPrefix(".PingFangSC") {
+                newTraits.remove(.traitItalic)
+                if let newDescriptor = familyDescriptor.withSymbolicTraits(newTraits) {
+                    return UIFont(descriptor: newDescriptor, size: pointSize)
+                }
+            }
+            
             return self
         }
 


SergioEstevao:
@shengyang998 thank you for the report and the PR! Do you mind adding a unit test to validate or just shows us some text where the problem occurs?

shengyang998:
> @shengyang998 thank you for the report and the PR! Do you mind adding a unit test to validate or just shows us some text where the problem occurs?

Here is my test case: 

1. Type this in (with italic format button turned on)
```
早上好 SergioEstevao
```

2. and then delete backward until the cursor hit the Chinese character, like:
```
早上好
     ^
```

3. right now the italic format button should be turned off automatically and you can **never** toggle it again. 

The problem is caused by `return self` in `guard let`. So I decided to remove the `traitItalic` and return a properly initialized `UIFont` instead. 

Moreover, the problem is not only for Chinese character, any character that don't have a `Italic` trait may cause the problem. However, I don't have resources and test them through. That is the reason why I use the code to specify the font name `PingFangSC` with `.postscriptName.hasPrefix(".PingFangSC")` (`SC` is for `Simplified Chinese`). 

There should be an improvement in my code, but I am not a linguist and don't have too much time to deal with all kind of languages. 

Yes, it would be better if there is a method to test whether a font has a italic trait or not, but I did not find it. If you find a better solution, let me know. Thanks. ❤️

maxme:
We recently switched the project license. Aztec is now under the MPL v2.0. More info on this ticket: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1305

If this PR gets merged, the code will be under the MPL v2.0.

PR #1157, opened 2019-03-27
Empty list

### LINKED ISSUE #1159: Empty list disappear
**Describe the bug**
Empty list HTML like this: `<ul><li></li><ul>` disappear on the visual editor and don't come back when switching to HTML

**To Reproduce**
1. Go to the empty demo
2. Switch to HTML mode
3. Write the following `<ul><li></li><ul>`
4. Switch to visual mode and see that nothing is show.

**Expected behavior**
Expect to see a empty bullet item like this: 

- 

**Screenshots**

 - Device: iPhone XS
 - OS: iOS 12
 - Version 1.5.0

SergioEstevao:
Fixes #1159

This PR implements a fix to support empty list elements like this: `<ul><li></li></ul>`

To test:
- Open an empty demo 
- Switch to HTML and add this: `<ul><li></li></ul>`
- Switch to Visual and see if the element is displayed.

### DIFF
--- a/Aztec/Classes/Extensions/String+EndOfLine.swift
+++ b/Aztec/Classes/Extensions/String+EndOfLine.swift
@@ -9,7 +9,7 @@ extension String {
     ///
     /// - Returns: `true` if the specified index is in an empty line, `false` otherwise.
     ///
-    func isEmptyLine(at index: String.Index) -> Bool {
+    public func isEmptyLine(at index: String.Index) -> Bool {
         return isStartOfNewLine(at: index) && isEndOfLine(at: index)
     }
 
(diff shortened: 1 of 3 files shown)


diegoreymendez:
So you were right when you mentioned to me that the conditions there should work right out of the box.

The problem seems to be with [the second line](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1157/files#diff-e8332b673f79516a496c1f635257543eR142) in the condition... I’m not sure why we have it.

But if I remove it 1 test fails… the test doesn’t make a ton of sense to me, but we both worked on it - do you have any recollection of why we had it.

My recommendation for testing a bit more would be to try disabling the conditions I linked above, and see if the breaking test is really necessary.

Let's chat whenever you have a second.

PR #1165, opened 2019-04-01
Add test to check nested list output.

### LINKED ISSUE #1164: Nested Lists are being break in several items
**Describe the bug**
When writing a nested list like this:

```
<ul>
  <li>List 1</li>
  <li>List 2
           <ul>
              <li>List 3</li>
              <li>List 4</li>
            </ul>
  </li>
</ul>
```

While the format is visually correct, the output HTML is incorrect and it's the following:

```
<ul>
  <li>List 1</li>
  <li>List 2</li>
  <li><ul><li>List 3</li></ul></li>
  <li><ul><li>List 4</li></ul></li>
  </li>
</ul>
```

**To Reproduce**

**Expected behavior**
The original HTML should be kept

 - Device: iPhoneXs
 - OS: iOS 12
 - Version 1.5.1

**Additional context**

SergioEstevao:
Fixes #1164 

At the moment this only adds a unit test to verify the bug

To test:
 - Open the demo empty app
 - Copy paste the following HTML:
```
<ul>
  <li>List 1</li>
  <li>List 2
           <ul>
              <li>List 3</li>
              <li>List 4</li>
            </ul>
  </li>
</ul>
```
 - Switch to visual and back to HTML and see if the output is correct.


etoledom:
Adding commit that fixes nested lists parsing.
The problem was with the `<li>` merging logic.  We want to merge list items when its child is a new list.
Added an exception for this together with the exception for `<figure>`.

I also added the possibility to a new bullet style for nested unordered lists, so it looks the same as on Gutenberg Web.
Since Android doesn't have support for this yet, we won't change the indented bullet character yet, but it's ready to be changed anytime 👍 

To test:
- Check that lists in general perform as expected.
- Check this PR https://github.com/wordpress-mobile/gutenberg-mobile/pull/863 and see that the list block is performing correctly.

@diegoreymendez if you could take a look would be great!

Aztec/Classes/TextKit/ParagraphProperty/TextList.swift
```diff
@@ -14,10 +14,21 @@ open class TextList: ParagraphProperty {
         case ordered
         case unordered
 
-        func markerText(forItemNumber number: Int) -> String {
+        func markerText(forItemNumber number: Int, level: Int) -> String {
             switch self {
             case .ordered:      return "\t\(number)."
-            case .unordered:    return "\t\u{2022}"
+            case .unordered:    return unorderedMarker(for: level)
```
diegoreymendez: Is the code here related to the bug fix?  I'm not sure I follow how the logic here affects the issue at hand.

etoledom: **TL:DR:** Not at all.

**Long story:**
I added the level information to `TextList` while searching of a way to fix the original bug. When I noticed that it wasn't necessary, instead of removing it, I saw the opportunity to implement the bullets-per-level support. On gutenberg they use a different bullet for indented unordered lists (`{25E6}`), so I thought it would be a good addition.

Talking with @hypest about it, we decided to keep the old one until Aztec-Android also has support for different bullet per indentation level.

We could move this code to a different PR if it's necessary. No problem with that! 👍

diegoreymendez: I hate to be adding extra work to your plate, but I'd encourage moving the remaining changes to a different PR indeed, for documentation and history purposes - and also to provide more clarity to this PR for future reference.

etoledom:
@diegoreymendez - Thanks for the review!

I removed the commits that added the bullets improvement and moved them here: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1173 . Added @SergioEstevao as reviewer since it's not a priority, but feel free to review that one too if you have the time/curiosity 👍

etoledom:
Thank you! 🙏

PR #1166, opened 2019-04-01
Fix a crash when system pasteboard is empty

jkmassel:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/8826

### To Test

**Before**
- Launch the WP app on a simulator, immediately after doing Hardware > Erase all Content and Settings (and thus has a clean pasteboard)
- Copy some text out of an Aztec post
- App will crash due to out-of-bounds exception

**After**
- Launch the test WP app as before (don't forget to Erase all Content and Settings)
- Copy some text out of Aztec post
- App will not crash

You could also run the updated Unit Tests – they demonstrate the issue in a more reliable way – the steps above didn't cause the crash _every_ time for some reason – it may have to do with content synced between the simulator and the host computer? 🤔

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -542,9 +542,8 @@ open class TextView: UITextView {
 
     // MARK: - Pasteboard Helpers
 
-    private func storeInPasteboard(encoded data: Data) {
-        let pasteboard = UIPasteboard.general
-        pasteboard.items[0][NSAttributedString.pastesboardUTI] = data
+    internal func storeInPasteboard(encoded data: Data, pasteboard: UIPasteboard = UIPasteboard.general) {
+        pasteboard.setData(data, forPasteboardType: NSAttributedString.pastesboardUTI)
     }
 
     // MARK: - Intercept keyboard operations
--- a/AztecTests/TextKit/TextViewTests.swift
+++ b/AztecTests/TextKit/TextViewTests.swift
@@ -1745,6 +1745,21 @@ class TextViewTests: XCTestCase {
         textView.cut(nil)
     }
 
+    /// This test verifies that trying to store data in an empty pasteboard
+    /// doesn't cause Aztec to crash.
+    ///
+    func testWritingIntoAnEmptyPasteboardDoesNotCauseAztecToCrash() {
+
+        let pasteboard = UIPasteboard.general
+        pasteboard.items.removeAll()
+
+        let data = "Foo".data(using: .utf8)!
+        let textView = TextViewStub(withSampleHTML: true)
+        textView.storeInPasteboard(encoded: data, pasteboard: pasteboard)
+
+        XCTAssertEqual(pasteboard.items.count, 1)
+    }
+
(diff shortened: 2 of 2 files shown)


jkmassel:
To _very_ consistently demonstrate the issue, place `pasteboard.items.removeAll()` at the start of `storeInPasteboard`, before anything else happens. It's artificial, but demonstrates the issue as well :)

etoledom:
Thank you @jkmassel for taking care of this! I'll review today. Sorry for the delay.

koke:
I was reviewing the follow-up in #1192 and I still don't understand how the crash might happen.
The pasteboard should not be empty since we just called `super.cut(_)` which should populate the pasteboard.

I only saw two crashes on fabric now, both on iPhone 6s and an old version of the app, so I'm a bit puzzled by this one

jkmassel:
> I was reviewing the follow-up in #1192 and I still don't understand how the crash might happen.
> The pasteboard should not be empty since we just called `super.cut(_)` which should populate the pasteboard.

I agree – I'm not sure why this was happening. Two possibilities I investigated in this:

1) The contents of the selection couldn't be copied to the pasteboard because....reasons?
2) The contents of the selection are somehow invalid and aren't causing a pasteboard item to be created? 

Given the fact that this crash was occasionally reproducible on example text after a clean launch makes me think the issue isn't #2. That unfortunately just leaves random cases of the pasteboard contents being dropped. Stepping through in the debugger confirmed the pasteboard was empty.

`5bcf3569f8b88c2963e1f9ca-fabric` still shows ~100 users on 12.2 and below hitting this issue, but none after. Somehow, this PR fixed _that_ crash, but like you, I'm at a bit of a loss as to how it does so.

koke:
> it appears a majority of the devices have less than 5% available RAM during the time of the crash.

> this crash was occasionally reproducible on example text after a clean launch

I think I have a new theory: the `copy` operation isn't synchronous. I just learned that the system pasteboard [persists upon restart](https://developer.apple.com/documentation/uikit/uipasteboard/1622106-generalpasteboard), and that combined with the handoff features makes me thing that perhaps, `cut`/`copy` immediately populates 
the pasteboard in the most common cases, but that's not guaranteed anywhere in the docs.

So in cases of memory pressure, on when you first use the pasteboard and it needs to bootstrap, there will be a small delay between calling `copy` and having the contents there. If that's the case, I wonder if the new fix will have unexpected side effects in those cases, as we might get the same text copied as two separate items (one from UIKit, one from our code with the custom Aztec attributed string).

I think I'll let this go, since testing/debugging this would take a lot of effort, and the side effects might not be that bad anyway, but if we see some reports about copy/pasting doing the wrong thing, we might want to investigate further.

koke:
To add some more insights, I had added an observer to `UIPasteboard.changedNotification` while copying from Gutenberg, and it gets called multiple times every time you copy. What's more interesting is that I left the debugger paused for a few minutes while looking at other things, and this happened:

```
(lldb) po notification.userInfo
2019-05-28 11:16:17.035531+0200 WordPress[48640:11468225] PBItemCollectionServicer connection disconnected.
nil```

PR #1173, opened 2019-04-17
Adding support for different bullets by UL indentation level

etoledom:
This PR adds support for different bullets by UL indentation level for parity with Gutenberg Web:

The indented bullet will still be the same than the previous one (black bullet) until Android adds support for this too.

bullets

To test:
- `TextList.swift`.
- Search for `return "\t\u{2022}"` and change the unicode from `2022` to `25E6` (white bullet).
- Open the example project and check that the indented bullets are the white dot. (as in the screenshot)

### DIFF
--- a/Aztec/Classes/Formatters/Implementations/TextListFormatter.swift
+++ b/Aztec/Classes/Formatters/Implementations/TextListFormatter.swift
@@ -34,7 +34,7 @@ class TextListFormatter: ParagraphAttributeFormatter {
             newParagraphStyle.setParagraphStyle(paragraphStyle)
         }
 
-        let newList = TextList(style: self.listStyle, with: representation)
+        let newList = TextList(style: self.listStyle, with: representation, level: newParagraphStyle.lists.count)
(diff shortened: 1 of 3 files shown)


SergioEstevao:
Could this be done in a way that the bullet characters are configurable by the user of the library?

diegoreymendez (commented):
While the idea behind this PR is great, my understanding is that right now this change doesn't really change bullets at all - only the code in preparation for something that we're expecting will come up.

I understand and agree with the direction but my suggestion for this PR would be to either:

- Make it possible to customize the bullets as @SergioEstevao is proposing, or
- Merge this once we can swap to actually having a new bullet.

That said, @SergioEstevao 's suggestion sounds much better IMHO for a generic library.

etoledom:
I agree with you both!
Now that this is on its own PR, it makes sense to extend it to add a proper support for customizable bullets.

SergioEstevao:
@etoledom are you still working on this?

etoledom:
I haven't retaken this task yet.
Feel free to take it 😁

PR #1175, opened 2019-04-18
Refactor TextView key commands.

etoledom:
Refactor TextView key commands to make key commands publicly and independently accessible.

This is needed to override `override open var keyCommands: [UIKeyCommand]? ` using the key commands defined internally by Aztec, without depending on the array indexing of them.

This PR has no change in logic.

To test:
- All unit tests should pass.
- Smoke test the example app.
- Test steps in `gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/889


etoledom:
Thank you!

PR #1178, opened 2019-04-23
Fix AutoPProcessor for Swift 5

etoledom:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/11515
WPiOS related PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/11529

The original issue describes the introduction of `e>` characters when saving a post.

I was able to reproduce this issue under these conditions:
- Any two pair of tags where the first tag holds text (as the `blockquote` tag)
- Both tags separated by a new line.
- The first tags contains a utf16 character as part of its content (as `—`).

In this case, the range used to insert `</blockquote>` when inserting the p tags was shifted by that character, inserting it earlier, and leaving the last two characters behind (`e>`).

The issue was that even though `newString` is a copy of the original string, the range created by `let matchRange = Range(match.range, in: self)` was different for each of them:

Screen Shot 2019-04-23 at 5 34 21 PM

Screen Shot 2019-04-23 at 5 42 24 PM

The issue was fixed by calculating the range from `newString`, that's where the range will be used to replace the matched tags:

`newString.replaceSubrange(matchRange, with: block(matchString, submatchStrings))`

**To test:**
- Check that the new test pass. (You can also see it fail by rolling back the changes).
- Open the Empty Demo in the WordPressEditor section (Calypso & Gutenberg).
- Paste this snippet in the html view:
```html
Thanks for joining me!
<blockquote>Good company in a journey makes the way seem shorter. — Izaak Walton</blockquote>
<img class="size-full wp-image-7" src="https://twentysixteendemo.files.wordpress.com/2015/11/post.png" alt="post" width="1000" height="563" />
```
- Go to visual mode and html mode several times.
- Check that no extra `e>` is added to the content.
- Smoke test Gutenberg to verify that it's not affected. (https://github.com/wordpress-mobile/WordPress-iOS/pull/11529)

cc @jkmassel

### DIFF
--- a/WordPress-Aztec-iOS.podspec
+++ b/WordPress-Aztec-iOS.podspec
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'WordPress-Aztec-iOS'
-  s.version          = '1.6.0'
+  s.version          = '1.6.1'
   s.summary          = 'The native HTML Editor.'
 
 # This description is used to generate tags and improve search results.
--- a/WordPress-Editor-iOS.podspec
+++ b/WordPress-Editor-iOS.podspec
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'WordPress-Editor-iOS'
-  s.version          = '1.6.0'
+  s.version          = '1.6.1'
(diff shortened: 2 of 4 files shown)


etoledom:
Thank you!

PR #1181, opened 2019-05-02
Fix nested lists rendering

### LINKED ISSUE #1180: Nested lists are not rendered properly
**Describe the bug**
If you set the HTML like this: 

```
<ul>
  <li>One</li>
  <li>
    <ul>
      <li>Two</li>
    </ul>
  </li>
</ul>
```

You should get this rendered:

Screenshot 2019-05-01 at 12 27 52

instead, we are getting this:

Screenshot 2019-05-01 at 12 28 13

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the empty demo
2. Switch to HTML mode
3. Copy/Paste the HTML above to it
4. Switch to visual and see the error

**Expected behaviour**
This list should display properly

 - Device: iPhone Xx
 - OS: iOS 12
 - Version 1.6.0

SergioEstevao:
Fixes #1180 

This PR addresses the issues reported on the ticket above by creating a dedicated LIConverter in order to check when to add a intrinsic representation for the <li> before or after the children.

The main logic here is that if the Li element has children elements that are nested lists and no text defined we need to add "\n" to the content in order for the marker to be shown.

To test:
 - If you set the HTML like this: 

```
<ul>
  <li>One</li>
  <li>
    <ul>
      <li>Two</li>
    </ul>
  </li>
</ul>
```

You should get something like this rendered:

Screenshot 2019-05-01 at 12 27 52

Besides this test the remaining logic of lists should be retested:
 - Create lists
 - Indent element
 - Change styles
 - Remove list item
 - Add new list item


SergioEstevao:
@etoledom Thanks for the check! I fixed the issue can you check it again?

etoledom:
Hey @SergioEstevao - I did a second pass to this PR after all the new commits, and I did find one issue:

After pressing enter to scape the list and coming back again, when I try to press double enter to scape the list again, the bullet stays with the cursor:

Lists

I noticed that this also happens on develop branch, so it shouldn't be a blocker to merge this PR ✅

PR #1193, opened 2019-05-24
Fix copy paste

SergioEstevao:
Fixes #

This PR brings the fix for copy paste to develop.

To test:
 - Run the demo app
 - Copy some content from one of the demo content
 - Paste it in another app, for example Notes or Reminders
 - See that the content is pasted in plain text.

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -555,7 +555,11 @@ open class TextView: UITextView {
     // MARK: - Pasteboard Helpers
 
     internal func storeInPasteboard(encoded data: Data, pasteboard: UIPasteboard = UIPasteboard.general) {
-        pasteboard.setData(data, forPasteboardType: NSAttributedString.pastesboardUTI)
+        if pasteboard.numberOfItems > 0 {
+            pasteboard.items[0][NSAttributedString.pastesboardUTI] = data;
+        } else {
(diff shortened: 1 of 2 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -555,7 +555,11 @@ open class TextView: UITextView {
     // MARK: - Pasteboard Helpers
 
     internal func storeInPasteboard(encoded data: Data, pasteboard: UIPasteboard = UIPasteboard.general) {
-        pasteboard.setData(data, forPasteboardType: NSAttributedString.pastesboardUTI)
+        if pasteboard.numberOfItems > 0 {
+            pasteboard.items[0][NSAttributedString.pastesboardUTI] = data;
```
koke: I don't understand why we're doing this one. Why not use [setItems(_:options:)](https://developer.apple.com/documentation/uikit/uipasteboard/1829417-setitems) in both cases?

SergioEstevao:
setItems was overwritting the values already there, unless you copy them
first... and then apply again.

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: I tried that, but setItems will reset everything, so unless you first copy what you already have there and add the extra UTI to it.

PR #1195, opened 2019-05-30
Add html to pasteboard when doing copy or cut.

SergioEstevao:
This PR implements copy/cut of HTML out of Aztec TextView.

This is done by getting the HTML of the selected range and then adding it to the pasteboard.

To test:
 - On the demo content copy or cut a selection that has HTML attributes (bold, italic, etc..)
 - Paste it on another app that supports it. For example ~Pages~ GB web on Safari
 - See that it was properly pasted.

### DIFF
--- a/AztecTests/TextKit/TextViewTests.swift
+++ b/AztecTests/TextKit/TextViewTests.swift
@@ -2017,7 +2017,7 @@ class TextViewTests: XCTestCase {
         XCTAssertEqual(output, expected)
     }
 
-    // MARK: - Copy tests
+    // MARK: - Copy/Paste tests
 
     func testCopyAndPasteToPlainText() {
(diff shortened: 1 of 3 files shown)


AztecTests/TextKit/TextViewTests.swift
```diff
@@ -2027,4 +2027,32 @@ class TextViewTests: XCTestCase {
 
         XCTAssertEqual(UIPasteboard.general.string, "This is text with attributes: bold")
     }
+
+    func testCopyHTML() {
+        let sourceTextView = TextViewStub(withHTML: "<p>This is text with attributes: <strong>bold</strong> and <italic>italic</italic></p>")
```
mkevins: Did you mean `<em>` instead of `<italic>`?

SergioEstevao: I think that for the test case, it doesn't make any difference

PR #1199, opened 2019-06-07
Make available method to extract HTML from a string.

SergioEstevao:
This PR just makes another method available to the outside to retrieve HTML from any attributed string.

To test:
 - There is no change in the logic of aztec in any way so no tests needed.
 - You can test the use of this method using this GB-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1097


SergioEstevao:
> I'm OK with this, but also wondering if the PR that's using this could just instantiate HTMLConverter directly

I thought about this, but I was afraid of missing any extra plugin configuration that can be done on it. So in the ended prefer to play it safe an use whatever is in the TextView.

PR #1207, opened 2019-08-22
Allow customisation of format bar colors

frosty:
This PR exposes the color properties for `FormatBarItem`. This is required for implementing dark mode in WPiOS. You can test it with this WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12373

### DIFF
--- a/WordPress-Aztec-iOS.podspec
+++ b/WordPress-Aztec-iOS.podspec
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'WordPress-Aztec-iOS'
-  s.version          = '1.7.0'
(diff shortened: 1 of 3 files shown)


WordPress-Aztec-iOS.podspec
```diff
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'WordPress-Aztec-iOS'
-  s.version          = '1.7.0'
+  s.version          = '1.7.1-beta.1'
```
etoledom: I believe we can release directly the next version 1.7.1, since today we have to release it anyway.

PR #1209, opened 2019-09-04
[Dark Mode] Expose editor html storage as property

danielebogo:
Refs. [WPiOS #12462](https://github.com/wordpress-mobile/WordPress-iOS/pull/12462)

This PR exposes the Editor html storage as let property in order to customize its colors due to the introduction of DarkMode. This is required for implementing dark mode in [WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/12462).

## To test:
• Run the Unit Tests
• Run [WPiOS #12462](https://github.com/wordpress-mobile/WordPress-iOS/pull/12462)

### DIFF
--- a/Aztec/Classes/EditorView/EditorView.swift
+++ b/Aztec/Classes/EditorView/EditorView.swift
@@ -6,12 +6,14 @@ import UIKit
 public class EditorView: UIView {
     public let htmlTextView: UITextView
     public let richTextView: TextView
+    public let htmlStorage: HTMLStorage
     
     // MARK: - Encoding / Decoding
     
     static let htmlTextViewKey = "Aztec.EditorView.htmlTextView"
     static let richTextViewKey = "Aztec.EditorView.richTextView"
-    
+    static let htmlStorageKey = "Aztec.EditorView.htmlStorage"
+
     // MARK: - Content Insets
     
     public var contentInset: UIEdgeInsets {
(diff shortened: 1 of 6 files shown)


Aztec/Classes/EditorView/EditorView.swift
```diff
@@ -6,12 +6,14 @@ import UIKit
 public class EditorView: UIView {
     public let htmlTextView: UITextView
     public let richTextView: TextView
+    public let htmlStorage: HTMLStorage
```
etoledom: Do you think it would be better to have a computed property instead? Something like

```swift
public var htmlStorage: HTMLStorage {
    guard let htmlStorage = htmlTextView.textStorage as? HTMLStorage else {
        fatalError("If this happens, something is very off on the init config")
    }
    return htmlStorage
}
```

I don't particularly like the `fatalError`, even though the guard should always succeed. Alternatively we can make it an optional, but that wouldn't make much sense for the "user", and how to handle the hypothetical `nil` case?

But it would remove the hassle with the coder init.

None look like a great option to me really, so I'm happy to keep it as it is if you think it is the best one 🙂

danielebogo: Yep I like it!

danielebogo:
Thanks @etoledom for reviewing this! I updated the porperty and the Podspecs. Tests passed successfully!

PR #1210, opened 2019-09-23
Update project to Swift 5 and xCode11

SergioEstevao:
To test:
 - Check if the project compiles and works correctly in Xcode 11.


.circleci/config.yml
```diff
@@ -9,24 +9,24 @@ workflows:
     jobs:
```
jtreanor: You will need to update the orb version above this to `ios: wordpress-mobile/ios@0.0.32` so it has the needed fixes.

.circleci/config.yml
```diff
@@ -9,24 +9,24 @@ workflows:
...
           name: Test
-          xcode-version: "10.2.0"
+          xcode-version: "11.0"
           carthage-update: true
           carthage-working-directory: Example
           bundle-install: false
           pod-install: false
           workspace: Aztec.xcworkspace
           scheme: AztecExample
           device: iPhone XS
```
jtreanor: This should be `iPhone 11`. Xcode 11 does not come with an iPhone XS simulator by default.

jkmassel:
One other issue – in order to compile this, we'll need to update Gridicons as well. I can take care of that – I'll tag you in the dependent PR?

SergioEstevao:
@jtreanor I followed your tips, but I'm still having trouble with CircleCI... any more ideas?

jtreanor:
@SergioEstevao It looks like you will need to update CocoaPods to at least 1.7.x for Xcode 11 compatibility.

You can update [this line](https://github.com/wordpress-mobile/AztecEditor-iOS/blob/develop/Gemfile#L2) to `1.7.5` and run `bundle install`

SergioEstevao:
@jtreanor it worked! Do you mind review it now?

PR #1221, opened 2019-10-08
Allow configuration of default fonts used in TextView

SergioEstevao:
This PR brings to TextView the following:

 - The possibility to replace the defaultFont being used after the TextView is created
 - Introduction of a FontProvider to allow configuration of fonts being used in the system
 - Configuration of the default font being used for monospace 

To test:
 - Run the sample app
 - Check the demo content
 - There is new button on the top right (+) press it and see if the default font changes in all the demo content.

### DIFF
--- a/Aztec/Classes/TextKit/FontProvider.swift
+++ b/Aztec/Classes/TextKit/FontProvider.swift
@@ -1,0 +1,33 @@
+import Foundation
+import UIKit
+
+public class FontProvider {
+
+    private init() {
+
+    }
+
+    static var shared = FontProvider()
+
+    lazy var monospaceFont: UIFont = {
+        let baseFont = UIFont(descriptor:UIFontDescriptor(name: "Menlo", size: 14), size:14)
+        let font: UIFont
+        if #available(iOS 11.0, *) {
+            font = UIFontMetrics.default.scaledFont(for: baseFont)
+        } else {
+            font = baseFont
+        }
+        return font
+    }()
+
+    lazy var defaultFont: UIFont = {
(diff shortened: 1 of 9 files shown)


Aztec/Classes/TextKit/FontProvider.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import UIKit
+
+public class FontProvider {
+
+    private init() {
+
+    }
+
+    static var shared = FontProvider()
```
etoledom: The font fields are public, but the shared instance is internal. We should probably make this one public too?

SergioEstevao:
@etoledom code updated ready for another look.

Example/Example/EditorDemoController.swift
```diff
@@ -195,11 +195,15 @@ class EditorDemoController: UIViewController {
...
             html = ""
         }
-        
+        navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .add, target: self, action: #selector(changeFont))
         editorView.setHTML(html)
         editorView.becomeFirstResponder()
     }
 
+    @objc func changeFont() {
+        editorView.richTextView.defaultFont = UIFont.preferredFont(forTextStyle: .callout)
```
etoledom: Thinking about this... It might feel counter-intuitive that setting the `defaultFont` also modifies the current content's font.
Maybe a method `setBaseFont()` or something similar would be more intuitive 🤔

What do you think?
No hard feelings though 👍

SergioEstevao: We talked about this and decide to keep the current name for now and add more documentation about the uses of default font.

PR #1233, opened 2019-11-13
Creates isLoaded property

leandroalonso:
Adds a property called `isLoaded` in `MediaAttachment`.

It returns `true` if the media is loaded, otherwise false.

To test:

The green build should be enough.

- [x] If it's feasible, I have added unit tests.

### DIFF
--- a/Aztec/Classes/TextKit/MediaAttachment.swift
+++ b/Aztec/Classes/TextKit/MediaAttachment.swift
@@ -45,6 +45,12 @@ open class MediaAttachment: NSTextAttachment {
         }
     }
 
+    /// Indicates if the media is loaded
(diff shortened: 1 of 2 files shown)


yaelirub:
Tests pass but worth mentioning that there are 3 failing tests when running on `iPhone 11 pro`:
'testResizingImageWorks`
`testResizingImageWorks2`
and `testImageForAttachment` 

I believe this failures were not introduced by this PR. Please consider creating an issue to fix these tests if one doesn't exist.

leandroalonso:
@yaelirub 

Created: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/1234

PR #1235, opened 2019-11-18
Update reference images for image resize tests.

### LINKED ISSUE #1234: Tests failing in larger screens (Plus and Max)
**Describe the bug**
Tests `testResizingImageWorks` and `testResizingImageWorks2` fails for devices with big screen (Plus and Max)

**To Reproduce**
1. Run the tests in any iPhone Plus or Max
2. Check that the above-mentioned tests fail

**Expected behavior**
The test should succeed.

SergioEstevao:
Fixes #1234 

This PR updates the reference images being used for 3x resolution devices to make tests pass.

To test:
 - Make sure unit test are now passing.


leandroalonso:
@SergioEstevao :shipit:

PR #1237, opened 2019-11-20
Fix copy/paste of attributed string from app without support for dark/light mode

SergioEstevao:
Fixes #https://github.com/wordpress-mobile/WordPress-iOS/issues/12688

This PR implements three fixes to sort issues around text color:
 - When copy/paste attributed string it makes sure that text color is set to default color if no color is set
 - It also make sure that when all text is deleted the typing attributes are reset to default attributes that include the default color
 - Make sure that the defaultColor is used for default attributes where converting from HTML mode.

To test:
 - Copy paste text from other apps. Examples: Notes app, Tumblr in post edit mode, inside different sections of Aztec
 - Open the demo app, put the cursor over the text with the red color, switch to HTML mode, and switch back to visual. Check that text is displayed in the correct color.
 - Open the demo app, select all the text and delete it. Start to type check that the text shows with the default attributes and not with the attributes of the last selection.

@maxme do you want these fixes to land on release 13.7 of WPiOS (currently frozen) or the next one 13.8?

### DIFF
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -647,7 +647,7 @@ open class TextView: UITextView {
         evaluateRemovalOfSingleLineParagraphAttributesAfterSelectionChange()
         ensureRemovalOfParagraphAttributesWhenPressingBackspaceAndEmptyingTheDocument()
         ensureCursorRedraw(afterEditing: deletedString.string)
-
+        recalculateTypingAttributes()
         notifyTextViewDidChange()
     }
 
@@ -1247,6 +1247,7 @@ open class TextView: UITextView {
     private func recalculateTypingAttributes(at location: Int) {
         
         guard storage.length > 0 else {
+            typingAttributes = defaultAttributes
             return
         }
         
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -249,13 +249,11 @@ open class TextView: UITextView {
     }
 
     open lazy var defaultTextColor: UIColor? = {
-        let color: UIColor
         if #available(iOS 13.0, *) {
-            color = UIColor.label
+            return UIColor.label
         } else {
-            color = UIColor.darkText
-        }
-        return color
+            return UIColor.darkText
+        }        
     }()
 
     // MARK: - Plugin Loading
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -248,16 +248,16 @@ open class TextView: UITextView {
         return attributes
     }
 
-    open var defaultTextColor: UIColor?
+    open lazy var defaultTextColor: UIColor? = {
+        let color: UIColor
+        if #available(iOS 13.0, *) {
+            color = UIColor.label
+        } else {
+            color = UIColor.darkText
+        }
+        return color
+    }()
 
-    override open var textColor: UIColor? {
-        get {
-            return super.textColor
-        }
-        set {
-            super.textColor = newValue            
-        }
-    }
     // MARK: - Plugin Loading
     
     var pluginManager: PluginManager {
--- a/Aztec/Classes/TextKit/TextView.swift
+++ b/Aztec/Classes/TextKit/TextView.swift
@@ -242,12 +242,22 @@ open class TextView: UITextView {
             .font: defaultFont,
             .paragraphStyle: defaultParagraphStyle,
         ]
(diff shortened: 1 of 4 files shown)


Aztec/Classes/TextKit/TextView.swift
```diff
@@ -242,12 +242,22 @@ open class TextView: UITextView {
...
         ]
-        if let color = textColor {
+        if let color = defaultTextColor {
             attributes[.foregroundColor] = color
         }
         return attributes
     }
-    
+
+    open var defaultTextColor: UIColor?
```
etoledom: Should we have a "smart" `defaultColor` pre-set?
This way we could avoid the need to always set a defaultTextColor on every place we use Aztec, and just set it if we want a different default.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -242,12 +242,22 @@ open class TextView: UITextView {
...
+
+    override open var textColor: UIColor? {
+        get {
+            return super.textColor
+        }
+        set {
+            super.textColor = newValue            
+        }
+    }
```
etoledom: Do we need to override `textColor`? It seems to not do anything special 🤔

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: Good suggestion, I'm going to change it.

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: Good catch, I was trying to catch when textColor was being changed and add those, let me remove them.

SergioEstevao: Addressed by https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1237/commits/ba8524aba1332550efb05cad583a85ed3511beb5

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: Addressed by: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1237/commits/ba8524aba1332550efb05cad583a85ed3511beb5

maxme:
> @maxme do you want these fixes to land on release 13.7 of WPiOS (currently frozen) or the next one 13.8?

IMO, It got the criterias to get merged in the current frozen version:
- It's a bug fix.
- You added tests.
- It's not too big and not impacting the rest of the app too much (less prone to create a blocker).

So let's merge it in 13.7.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -242,12 +242,22 @@ open class TextView: UITextView {
...
             attributes[.foregroundColor] = color
         }
         return attributes
     }
-    
+
+    open lazy var defaultTextColor: UIColor? = {
+        let color: UIColor
+        if #available(iOS 13.0, *) {
+            color = UIColor.label
```
frosty: You could just return directly in each branch of this if statement, and you can drop the `UIColor`? (e.g. `return .label`)

Aztec/Classes/TextKit/TextViewPasteboardDelegate.swift
```diff
@@ -80,6 +80,21 @@ open class AztecTextViewPasteboardDelegate: TextViewPasteboardDelegate {
...
+    private func fixColors(in string: NSAttributedString, using baseColor: UIColor?) -> NSAttributedString {
+        guard #available(iOS 13.0, *) else {
+            return string
+        }
+        let colorToUse = baseColor ?? UIColor.label
+
+        let newString = NSMutableAttributedString(attributedString: string)
+        newString.enumerateAttributes(in: newString.rangeOfEntireString, options: []) { (attributes, range, stop) in
+            if attributes[.foregroundColor] == nil {
+                newString.setAttributes([.foregroundColor: colorToUse], range: range)
```
frosty: Out of interest, do we need to set color attributes, or can we just remove any that already exist there?

SergioEstevao:
From my investigation we need to set them explicitly or else we will get
the bug!

Aztec/Classes/TextKit/TextViewPasteboardDelegate.swift
SergioEstevao: From my investigation we need to set them explicitly or else we will get
the bug!

Aztec/Classes/TextKit/TextView.swift
SergioEstevao: Done

PR #1243, opened 2019-12-02
Implements a Color provider

SergioEstevao:
Fixes #https://github.com/wordpress-mobile/gutenberg-mobile/issues/1603#issuecomment-559841868

This PR adds support for the standard HTML colors and also provides a way for provide default colors for formatters.

Screenshot 2019-12-02 at 16 45 09

To test:
 - Start the demo app
 - Check if the colors are correctly set on the Colors line
 - Check if the code tag has the correct background color.


Aztec/Classes/TextKit/ColorProvider.swift
```diff
@@ -0,0 +1,57 @@
...
+        guard let colorDataURL = Bundle.aztecBundle.url(forResource: "html_colors", withExtension: "json"),
+            let data = try? Data(contentsOf: colorDataURL),
+            let object = try? JSONSerialization.jsonObject(with: data, options: .allowFragments),
+            let colorDict = object as? [String: String] else {
+                return [String:UIColor]()
+        }
+
+        return colorDict.compactMapValues { (value) -> UIColor? in
+            return UIColor.init(hexString: value)
+        }
```
etoledom: This could be written as:
`return colorDict.compactMapValues { UIColor(hexString: $0) }`
A bit more direct and still readable.
If you prefer to keep it as it is (which is also fine), the `init` is still not needed 👍

Aztec/Classes/TextKit/ColorProvider.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// Returns a color designated by name if available.
+    ///  By default the the color available are the one defined in the HTML specification: https://www.w3schools.com/colors/colors_names.asp
+    /// - Parameter name: the name of the color, or nil if the name is not available
+    public func color(named name: String) -> UIColor? {
+        if let color = htmlColors[name.lowercased()] {
+            return color
+        }
+
+        return extraColors[name]
+    }
```
etoledom: In the comment `/// - Parameter name: the name of the color, or nil if the name is not available`
It seems like the second part refers to the return value, and not to the `name` parameter, right?

Aztec/Classes/TextKit/ColorProvider.swift
```diff
@@ -0,0 +1,57 @@
...
+
+        return colorDict.compactMapValues { (value) -> UIColor? in
+            return UIColor.init(hexString: value)
+        }
+    }()
+
+    private var extraColors = [String:UIColor]()
+
+    /// Adds extra named colors to be used by the `color(named:)` method.
+    ///  If the there is a name collision the value on the added dictionary will overwrite the current color.
```
etoledom: Small typo :) 
```suggestion
    ///  If there is a name collision, the value in the added dictionary will overwrite the current color.
```

Aztec/Classes/TextKit/ColorProvider.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// - Parameter namedColors: a dictionary of colors keyed by name
+    public func add(namedColors: [String: UIColor]) {
+        extraColors.merge(namedColors) { (base, other) -> UIColor in
+            return other
+        }
+    }
```
etoledom: By the current implementation, if a user adds a new color called `"yellow"` with a different yellow tone (for example), when fetching the `color(named: "yellow")`, it will still return the default yellow value.

This sounds like it's fine since we are following the standard colours for the default. But it might be unexpected.

Not sure if this is really a problem. What do you think?

Aztec/Classes/TextKit/ColorProvider.swift
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        return extraColors[name]
+    }
+
+    public lazy var codeBackgroungColor: UIColor = {
+        if #available(iOS 13.0, *) {
+            return UIColor.secondarySystemBackground
+        } else {
+            return UIColor.lightGray
```
etoledom: The `UIColor`s can be inferred from the property type.

etoledom: Also, this prop (`codeBackgroungColor`) feels a bit out of place, is more like a style property for the Code tag while this class seems to be related with HTML Color stuff.

But the class name, it seems to be a generic colours helper? Do we plan to use this for more color related work?

SergioEstevao: The idea of this class is to be similar to FontProvider, so we can use to configure and read default color values for formatters in Aztec.

Aztec/Classes/TextKit/ColorProvider.swift
SergioEstevao: I want to keep the default HTML colors "static", and allow to add extra colors, for example if Gutenberg defines some standard CSS colors that we will need to use.

etoledom:
Thank you for the update! Looks great 🎉 :shipit:

PR #1246, opened 2019-12-06
Improve “More” button accessibility

shiki:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/12953

Currently, when using VoiceOver, hovering on the _More_ button would make VoiceOver speak “Possibly, More“:

This changes that so that it can be configured to say “More”. 

The button will now also report to VoiceOver that the `FormatBar` is currently expanded or not. When hovering on an expanded bar's More button, VoiceOver would speak “selected“. 

## Testing

1. Navigate to the Standard Demo.
2. Turn VoiceOver on. Please confirm the changes described above. 

## Reviewing 

Based on what I understand, the localized “More” accessibility label should be configured in the hosting app (e.g. WPiOS). In here, I added a configurable `FormatBar` property called `overflowToggleAccessibilityLabel`. Please let me know if there's a more appropriate way to do this.

## Submitter Checklist

- [x] If it's feasible, I have added unit tests.


Example/Example/EditorDemoController.swift
```diff
@@ -993,6 +993,8 @@ extension EditorDemoController {
         }
 
         toolbar.overflowToggleIcon = Gridicon.iconOfType(.ellipsis)
+        toolbar.overflowToggleAccessibilityLabel = NSLocalizedString("More", comment: "Accessibility label for the More button on formatting toolbar.")
```
SergioEstevao: I was thinking, while it's good to allow this to be configurable from the outside shouldn't we have a default value for the label inside the library code?

shiki: That's a good idea. I'm not familiar with how localizations work for libraries. Is using `NSLocalizedString` like here enough?

shiki: Based on [WPiOS's `Localizable.strings`](https://github.com/wordpress-mobile/WordPress-iOS/blob/44a4dc6bf1bb44b677293faacf86e761cd7255e8/WordPress/Resources/ja.lproj/Localizable.strings#L5706-L5707), it looks like the `NSLocalizedString()` calls in Aztec are copied over to it. I initially thought they weren't. It looks like we don't need to have the customizability after all. 

- Moved the `NSLocalizedString()` to the `overflowToggleItem` 82208f1d6ff6ce99dcef738a58722cc3712d9fa6
- Removed unnecessary property f2d46f5960886eaf839d5b5c7cd7036d76b66091

Thank you for pointing this out, @SergioEstevao. 🙂

shiki:
Thank you, @SergioEstevao! 🙏

PR #1247, opened 2019-12-06
Compatibility with xcode 10

etoledom:
This PR adds a XCODE11 flag to be use to continue supporting Xcode 10.

This is a workaround to continue using Xcode 10 in gutenberg-mobile CI. And we should roll back these changes as soon as we update it.

To test:
- Test that dark-mode colours are not broken.
- Test CI on gutenberg-mobile and check that e2e iOS tests on Device are ✅  (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1657)


etoledom:
Thank you! I'll prepare the hotfix release next 👍

Issue #1086, opened 2018-11-18
Nested blockquote style


sergstav:
Hello to all, thank for awesome library. I use it for displaying static html text, by using isEditable=false.
I try to display nested blockqutotes, with same style as parent one.
### Expected behavior
Expected behavior: nested blockquotes has the same style as parent one: at least left border line.

<blockquote><blockquote>многострочная цитатка<blockquote>теперь оч сложный вариант<br />сначала тут текст у нас<br />потом внизу цитата</blockquote><br /><div><br /></blockquote><br />вло<br /><br /></blockquote>

### Actual behavior
Nested quotes display as usual text, but with tab.
![screen shot 2018-11-18 at 5 31 34 pm](https://user-images.githubusercontent.com/3393265/48674730-25492480-eb58-11e8-81a4-92f62c7bd967.png)

### Steps to reproduce the behavior
Create nested blockquote text. For example:
<blockquote>hi test<blockquote>nested one</blockquote></blockquote>

##### Tested on [device], iOS [version], Aztec [version]
Tested on iOS Simulator iPhone XR
iOS 12.1
Aztec version 1.1

MadeByDouglas:
I agree, this would be nice. im investigating a solution, will update here if I come up with anything.

MadeByDouglas:
OK PR is open #1252

PR #1252, opened 2019-12-10
Multi level blockquote support

MadeByDouglas:
Fixes issue #1086 

To test: try tabbing and shift tabbing to go through nested quotes

Features:

- supports up to 9 levels of indentation, arbitrary, could be more, but seemed enough
- supports from 1 to 5 levels of colored quotes, editable by user of the API, default is 5
- style can be made to look like apple mail app or other mail apps by removing background color and customizing colors

Screen Shot 2019-12-09 at 7 56 50 PM

Known issue:
when pasting quotes from either internally or externally such as a mail client, the text color doesn't change properly. Something with the tryPastingHTML method and replace not correctly getting color for new string. Would appreciate help fixing this.

### DIFF
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -10,12 +10,14 @@ class LayoutManager: NSLayoutManager {
     /// Blockquote's Left Border Color
     /// Include 3 nested values
     
-    var blockquoteBorderColor = [0 : UIColor(hexString: "66D6FF")!,
-                                1 : UIColor(hexString: "5198E8")!,
-                                2 : UIColor(hexString: "597CFF")!]
+    var blockquoteBorderColor = [0 : UIColor(hexString: "9252FF")!,
+                                1 : UIColor(hexString: "5A4AE8")!,
+                                2 : UIColor(hexString: "5D77FE")!,
+                                3 : UIColor(hexString: "4A8EE8")!,
+                                4 : UIColor(hexString: "52CAFF")!]
     
-    /// Update this accordingly to the dictionary above
-    static let quoteMaxDepth: Int = 2
+    /// Update this accordingly to the highest index value in the dictionary above
+    static let quoteMaxDepth: Int = 4
 
 
     /// Blockquote's Background Color
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -17,7 +17,7 @@ class LayoutManager: NSLayoutManager {
                                 4 : UIColor(hexString: "52CAFF")!]
     
     /// Update this accordingly to the highest index value in the dictionary above
-    static let quoteMaxDepth: Int = 4
+    var quoteMaxDepth: Int = 4
 
 
     /// Blockquote's Background Color
@@ -147,7 +147,7 @@ private extension LayoutManager {
     
     private func drawNestedBlockquote(in rect: CGRect, with context: CGContext, at depth: Int) {
         
-        let index = depth > LayoutManager.quoteMaxDepth ? LayoutManager.quoteMaxDepth : depth
+        let index = depth > quoteMaxDepth ? quoteMaxDepth : depth
 
         if let blockquoteBorderColor = blockquoteBorderColor[index] {
             let borderRect = CGRect(origin: rect.origin, size: CGSize(width: blockquoteBorderWidth, height: rect.height))
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -73,6 +73,19 @@ private extension LayoutManager {
                 let blockquoteRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: blockquoteIndent, lineEndsParagraph: lineEndsParagraph)
 
                 self.drawBlockquote(in: blockquoteRect.integral, with: context)
+                
+                //nested blockquote
+                let nestDepth = paragraphStyle.blockquoteNestedIndent.depth
+                guard nestDepth >= 1 else {return}
+                for index in 1...nestDepth {
+                    let indent = CGFloat(index) * Metrics.listTextIndentation
+                    
+                    let nestRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: indent, lineEndsParagraph: lineEndsParagraph)
+                    
+                    self.drawBlockquote(in: nestRect.integral, with: context)
+                }
+                
+            
             }
         }
 
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -8,12 +8,14 @@ import QuartzCore
 class LayoutManager: NSLayoutManager {
 
     /// Blockquote's Left Border Color
-    ///
-    var blockquoteBorderColor: UIColor? = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
+    /// Include 3 nested values
     
-    var blockquoteBorderColor1: UIColor? = UIColor(red: 0.62, green: 0.75, blue: 0.83, alpha: 1.0)
-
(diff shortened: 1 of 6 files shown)


SergioEstevao:
@MadeByDouglas thank you so much for the PR, I'm going to have a look and review it.

One small thing that you could do to make the review easier is to create a separate branch (feature branch) for your PR instead of using the develop branch.

You can name your branch something like: ```issue\1086_blockquote_multilevel_support``` and then submit the PR saying that you want to merge from there to our develop branch.

Thanks again.

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -8,8 +8,17 @@ import QuartzCore
 class LayoutManager: NSLayoutManager {
 
     /// Blockquote's Left Border Color
-    ///
-    var blockquoteBorderColor: UIColor? = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
+    /// Include 3 nested values
+    
+    var blockquoteBorderColor = [0 : UIColor(hexString: "9252FF")!,
```
SergioEstevao: Why couldn't be this a simple array like this:
```
var blockquoteBorderColor: [UIColor] = [UIColor(hexString: "9252FF")!, UIColor(hexString: "5A4AE8")!, UIColor(hexString: "5D77FE")!, UIColor(hexString: "4A8EE8")!, UIColor(hexString: "52CAFF")!]

File: Aztec/Classes/TextKit/LayoutManager.swift [hunk 2]
```diff
@@ -8,8 +8,17 @@ import QuartzCore
...
+                                1 : UIColor(hexString: "5A4AE8")!,
+                                2 : UIColor(hexString: "5D77FE")!,
+                                3 : UIColor(hexString: "4A8EE8")!,
+                                4 : UIColor(hexString: "52CAFF")!]
+    
+    /// Update this accordingly to the highest index value in the dictionary above
+    var quoteMaxDepth: Int = 4
```
SergioEstevao: This could be implemented like this:
``` var quoteMaxDepth = { get { return blockquoteBorderColor.count } }
in order to autosync with the above array/dictionary

File: Aztec/Classes/TextKit/LayoutManager.swift [hunk 3]
```diff
@@ -132,7 +166,7 @@ private extension LayoutManager {
 
         }
 
-        if let blockquoteBorderColor = blockquoteBorderColor {
+        if let blockquoteBorderColor = blockquoteBorderColor[0] {
```
SergioEstevao: Instead of doing ```blockquoteBorderColor[0]``` you could do ```blockquoteBorderColor.first``` that is more explicit and clear

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -223,6 +223,10 @@ open class TextView: UITextView {
     // MARK: - Properties: Text Lists
 
     var maximumListIndentationLevels = 7
+    
+    // MARK: - Properties: Blockquotes
+    
+    var maximumBlockquoteIndentationLevels = 9
```
SergioEstevao: This probably should be synced with the values on the LayoutManager like this:
```
var maximumBlockquoteIndentationLevels = { get { return layout. quoteMaxDepth } }
```

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -281,17 +285,69 @@ open class TextView: UITextView {
...
+        }
+    }
+    
     /// Blockquote Blocks Border Color.
     ///
-    @objc dynamic public var blockquoteBorderColor: UIColor? {
+    @objc dynamic public var blockquoteBorderColor: UIColor {
         get {
-            return layout.blockquoteBorderColor
+            return layout.blockquoteBorderColor[0]!
```
SergioEstevao: Use first as mentioned above.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -281,17 +285,69 @@ open class TextView: UITextView {
...
         }
         set {
-            layout.blockquoteBorderColor = newValue
+            layout.blockquoteBorderColor[0] = newValue
         }
     }
 
+    @objc dynamic public var blockquoteBorderColorNested1: UIColor {
```
SergioEstevao: I think it will probably be easier and more flexible to expose the full color array for quotes in layout, instead of having this "shortcut" for each level.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -527,41 +583,142 @@ open class TextView: UITextView {
...
+    private func indent(blockquote: Blockquote, increase: Bool = true) {
+        
+        let paragraphStyle = typingAttributes.paragraphStyle()
+        let currentDepth = paragraphStyle.blockquoteNestedIndent.depth
+        
+        let textColor = getTextColorForQuoteDepth(currentDepth)
+                
+        var previousAttributes = typingAttributes
+        previousAttributes.removeValue(forKey: .foregroundColor)
+        previousAttributes[.foregroundColor] = textColor.prev
```
SergioEstevao: I don't think we should change the color of text in the quote, only the border/bar.

This could be a future nice to have, but will bring us problems with color attributes in the text implemented by style attributes and also can bring issues with dark mode support.

Even if we decide to go ahead and do this, the change of attributes needs to be done inside the formatter class not in this location.

MadeByDouglas:
> You can name your branch something like: `issue\1086_blockquote_multilevel_support` and then submit the PR saying that you want to merge from there to our develop branch.

Oh definitely, sorry totally forgot to make a branch since I was on my fork. Should I still go ahead and do that and make a new PR? or just remember that for next time?

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: hmm, now that you mention it, I guess this could work. For some reason at the time I thought I needed a dictionary to ensure I kept the order / accessed the right color at the right level. Let me investigate and make the change.

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: oh right thats better, only thing it would have to be `count-1` because the way I use this var is to point to a particular index so 0-4 rather than 1-5

SergioEstevao:
Please create the new branch and a new PR, If needed I can add the review
comments again.

Aztec/Classes/TextKit/TextView.swift
MadeByDouglas: well, so actually this var `maximumBlockquoteIndentationLevels` means something different than `quoteMaxDepth` 

`maximumBlockquoteIndentationLevels` means how many possible nested quote levels there could be, how many times tab keeps indenting.

`quoteMaxDepth` is how many different colors it supports before subsequent nested tabs just use the same color.

Other editors, like Froala, or apple mail app, support many quote levels, 20+, (more than I thought actually) however only 3 levels of colors. I thought supporting 5 levels of colors provided a better experience. And actually, I guess I should increase the other limit to something much higher than 9

https://www.froala.com/wysiwyg-editor

Aztec/Classes/TextKit/TextView.swift
MadeByDouglas: Hmm, I guess this would work. Originally I wanted to keep the colors limited to max of 5 and I wasn't sure how to limit the length of the array a user might submit. However, if thats not a concern, I guess it would work to just let them submit any array they choose?

Aztec/Classes/TextKit/TextView.swift
MadeByDouglas: Hmm, I really need this feature though. My project which uses aztec needs the nested quoting to look like the apple mail app, which changes text color rather than the background color. Can you point me in the way of doing this in the formatter class? I will take a look myself now that you mention it as well.

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: ok with the array this var really isn't necessary anymore; will open the new PR

MadeByDouglas:
I am closing this one since PR #1253 is live

PR #1253, opened 2019-12-12
Issue/1086 blockquote multilevel support

### LINKED ISSUE #1086: Nested blockquote style

MadeByDouglas:
Fixes #1086 

To test: new array method of tracking nested colors and next / previous index logic cleaned up to be more readable

replaces PR #1252 

**Original PR Description**

To test: try tabbing and shift tabbing to go through nested quotes

**Features:**

supports up to 9 levels of indentation, arbitrary, could be more, but seemed enough
supports from 1 to 5 levels of colored quotes, editable by user of the API, default is 5
style can be made to look like apple mail app or other mail apps by removing background color and customizing colors

Screen Shot 2019-12-09 at 7 56 50 PM

**Known issue:**
when pasting quotes from either internally or externally such as a mail client, the text color doesn't change properly. Something with the tryPastingHTML method and replace not correctly getting color for new string. Would appreciate help fixing this.

### DIFF
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -70,16 +70,19 @@ private extension LayoutManager {
             let blockquoteGlyphRange = glyphRange(forCharacterRange: range, actualCharacterRange: nil)
 
             enumerateLineFragments(forGlyphRange: blockquoteGlyphRange) { (rect, usedRect, textContainer, glyphRange, stop) in
+                
+                let startIndent = paragraphStyle.blockquoteIndent
+
                 let lineRange = self.characterRange(forGlyphRange: glyphRange, actualGlyphRange: nil)
                 let lineCharacters = textStorage.attributedSubstring(from: lineRange).string
                 let lineEndsParagraph = lineCharacters.isEndOfParagraph(before: lineCharacters.endIndex)
-                let blockquoteRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: 0.0, lineEndsParagraph: lineEndsParagraph)
+                let blockquoteRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: startIndent, lineEndsParagraph: lineEndsParagraph)
 
                 self.drawBlockquoteBackground(in: blockquoteRect.integral, with: context)
                 
-                let nestDepth = paragraphStyle.blockquoteDepth
+                let nestDepth = paragraphStyle.blockquoteNestDepth
                 for index in 0...nestDepth {
-                  let indent = CGFloat(index) * Metrics.listTextIndentation
+                  let indent = startIndent + CGFloat(index) * Metrics.listTextIndentation
 
                   let nestRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: indent, lineEndsParagraph: lineEndsParagraph)
 
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -140,26 +140,24 @@ private extension LayoutManager {
     }
     
     private func drawBlockquoteBorder(in rect: CGRect, with context: CGContext, at depth: Int) {
-        
         let quoteCount = blockquoteBorderColors.count
         let index = min(depth, quoteCount-1)
         
-        if blockquoteBorderColors.indices.contains(index) {
-            let borderColor = blockquoteBorderColors[index]
-            let borderRect = CGRect(origin: rect.origin, size: CGSize(width: blockquoteBorderWidth, height: rect.height))
-            borderColor.setFill()
-            context.fill(borderRect)
-        }
-
+        guard index < quoteCount else {return}
+        
+        let borderColor = blockquoteBorderColors[index]
+        let borderRect = CGRect(origin: rect.origin, size: CGSize(width: blockquoteBorderWidth, height: rect.height))
+        borderColor.setFill()
+        context.fill(borderRect)
     }
 
     /// Draws a single Blockquote Line Fragment, in the specified Rectangle, using a given Graphics Context.
     ///
     private func drawBlockquoteBackground(in rect: CGRect, with context: CGContext) {
-        if let blockquoteBackgroundColor = blockquoteBackgroundColor {
-            blockquoteBackgroundColor.setFill()
-            context.fill(rect)
-        }
+        guard let color = blockquoteBackgroundColor else {return}
+        
+        color.setFill()
+        context.fill(rect)
     }
 }
 
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -10,12 +10,14 @@ class LayoutManager: NSLayoutManager {
     /// Blockquote's Left Border Color
     /// Include 3 nested values
     
-    var blockquoteBorderColor = [0 : UIColor(hexString: "66D6FF")!,
-                                1 : UIColor(hexString: "5198E8")!,
-                                2 : UIColor(hexString: "597CFF")!]
+    var blockquoteBorderColor = [0 : UIColor(hexString: "9252FF")!,
+                                1 : UIColor(hexString: "5A4AE8")!,
+                                2 : UIColor(hexString: "5D77FE")!,
+                                3 : UIColor(hexString: "4A8EE8")!,
+                                4 : UIColor(hexString: "52CAFF")!]
     
-    /// Update this accordingly to the dictionary above
-    static let quoteMaxDepth: Int = 2
+    /// Update this accordingly to the highest index value in the dictionary above
+    static let quoteMaxDepth: Int = 4
 
 
     /// Blockquote's Background Color
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -17,7 +17,7 @@ class LayoutManager: NSLayoutManager {
                                 4 : UIColor(hexString: "52CAFF")!]
     
     /// Update this accordingly to the highest index value in the dictionary above
-    static let quoteMaxDepth: Int = 4
+    var quoteMaxDepth: Int = 4
 
 
     /// Blockquote's Background Color
@@ -147,7 +147,7 @@ private extension LayoutManager {
     
     private func drawNestedBlockquote(in rect: CGRect, with context: CGContext, at depth: Int) {
         
-        let index = depth > LayoutManager.quoteMaxDepth ? LayoutManager.quoteMaxDepth : depth
+        let index = depth > quoteMaxDepth ? quoteMaxDepth : depth
 
         if let blockquoteBorderColor = blockquoteBorderColor[index] {
             let borderRect = CGRect(origin: rect.origin, size: CGSize(width: blockquoteBorderWidth, height: rect.height))
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -76,7 +76,9 @@ private extension LayoutManager {
                 let lineEndsParagraph = lineCharacters.isEndOfParagraph(before: lineCharacters.endIndex)
                 let blockquoteRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: blockquoteIndent, lineEndsParagraph: lineEndsParagraph)
 
-                self.drawBlockquote(in: blockquoteRect.integral, with: context)
+                self.drawBlockquoteBackground(in: blockquoteRect.integral, with: context)
+                self.drawBlockquoteBorder(in: blockquoteRect.integral, with: context, at: 0)
+
                 
                 //nested blockquote
                 let nestDepth = paragraphStyle.blockquoteNestedIndent.depth
@@ -86,7 +88,7 @@ private extension LayoutManager {
                     
                     let nestRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: indent, lineEndsParagraph: lineEndsParagraph)
                     
-                    self.drawNestedBlockquote(in: nestRect.integral, with: context, at: index)
+                    self.drawBlockquoteBorder(in: nestRect.integral, with: context, at: index)
                 }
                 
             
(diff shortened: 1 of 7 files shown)


Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -8,8 +8,13 @@ import QuartzCore
 class LayoutManager: NSLayoutManager {
 
     /// Blockquote's Left Border Color
-    ///
-    var blockquoteBorderColor: UIColor? = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
+    /// Include 3 nested values
```
SergioEstevao: This comment is outdated, probably should indicate what is the purpose of the variable and don't mention the number of levels that can be changed depending of the needs.

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -73,6 +78,19 @@ private extension LayoutManager {
...
+                guard nestDepth >= 1 else {return}
+                for index in 1...nestDepth {
+                    let indent = CGFloat(index) * Metrics.listTextIndentation
+                    
+                    let nestRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: indent, lineEndsParagraph: lineEndsParagraph)
+                    
+                    self.drawNestedBlockquote(in: nestRect.integral, with: context, at: index)
+                }
+                
+
```
SergioEstevao: Instead of having this two different ways of drawing the borders I suggest we do the following changes:

 - Rename the method `drawBlockquote()` to be `drawBlockquoteBackground()` and change it to only draw the background of the blockquote.
 - Rename the `drawNestedBlockquote()` to be `drawBlockquoteBorder()` and use it to draw **all** the borders of the blockquote.

This will simplify the code and it make more obvious the steps involved.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -223,6 +223,12 @@ open class TextView: UITextView {
     // MARK: - Properties: Text Lists
 
     var maximumListIndentationLevels = 7
+    
+    // MARK: - Properties: Blockquotes
+    
+    /// The max levels of quote indentation allowed
+    /// Set to 9 for iPhone sized screen, 24 for other devices
+    var maximumBlockquoteIndentationLevels = UIDevice.current.userInterfaceIdiom == .phone ? 9 : 24
```
SergioEstevao: I think it's better to have a single value here and let users of the library change it according to their needs. 

Also on the iPad we have to support the multitask modes where the screen can have the same width of a phone and 24 levels don't make sense then.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -983,7 +1095,21 @@ open class TextView: UITextView {
...
+        //get original color or previous nested color as appropriate when toggling back and forth
+        var originalQuoteAttributes = typingAttributes
+        originalQuoteAttributes.removeValue(forKey: .foregroundColor)
+        
+        let textColor = getTextColorForQuoteDepth(currentDepth, toggle: true)
+        originalQuoteAttributes[.foregroundColor] = textColor.prev
+        
+        let formatter = BlockquoteFormatter(placeholderAttributes: originalQuoteAttributes)
+        //use the first color when toggling on
+        formatter.nextTextColor = textColor.next
```
SergioEstevao: At the moment I don't think we should change the color of the text in the blockquote. I understand that you have a need for this for your work, but can we handle this in a separate follow up PR?

The color changes need to be done inside the formatter and we will need to send the value of color to apply/revert to the formatter, but again I think that change should be done on a separate PR and we should focus on the border color on this one.

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -122,6 +140,20 @@ private extension LayoutManager {
 
         return blockquoteRect
     }
+    
+    private func drawNestedBlockquote(in rect: CGRect, with context: CGContext, at depth: Int) {
+        
+        let quoteCount = blockquoteBorderColors.count
+        let index = depth < quoteCount ? depth : quoteCount-1
```
SergioEstevao: I think this will be easier to write like this: ```min(depth, quoteCount-1)```

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: oh good idea

Aztec/Classes/TextKit/TextView.swift
MadeByDouglas: hmm, true I wasn't sure how to do the compact width check here. Also if user can change this and it happens to be less than whats in the `blockquoteBorderColors` array it will crash. I think I will just make this value be the count of `blockquoteBorderColors` +9, that should cover any reasonable use case, and on compact widths with high indent, it will just look a little funny, but thats unlikely to happen anyway

MadeByDouglas: ok, that makes sense. I will just make it public var then.

Yeah, I wasn't sure how to get compactWidth in that instance didnt spend too much time on it

Aztec/Classes/TextKit/TextView.swift
MadeByDouglas: ok, I will separate that out into another PR

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: oh yes much better 😅

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -72,7 +76,22 @@ private extension LayoutManager {
...
+                //nested blockquote
+                let nestDepth = paragraphStyle.blockquoteNestedIndent.depth
+                guard nestDepth >= 1 else {return}
+                for index in 1...nestDepth {
+                    let indent = CGFloat(index) * Metrics.listTextIndentation
+                    
+                    let nestRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: indent, lineEndsParagraph: lineEndsParagraph)
+                    
+                    self.drawBlockquoteBorder(in: nestRect.integral, with: context, at: index)
+                }
```
SergioEstevao: We could merge the drawing of the border 0 and other depths by just doing this: 
```
               let nestDepth = paragraphStyle.blockquoteNestedIndent.depth               
                for index in 0...nestDepth {
                    let indent = CGFloat(index) * Metrics.listTextIndentation

                    let nestRect = self.blockquoteRect(origin: origin, lineRect: rect, blockquoteIndent: indent, lineEndsParagraph: lineEndsParagraph)

                    self.drawBlockquoteBorder(in: nestRect.integral, with: context, at: index)
                }
```

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -122,20 +142,27 @@ private extension LayoutManager {
 
         return blockquoteRect
     }
+    
+    private func drawBlockquoteBorder(in rect: CGRect, with context: CGContext, at depth: Int) {
+        
+        let quoteCount = blockquoteBorderColors.count
+        let index = min(depth, quoteCount-1)
+        
+        if blockquoteBorderColors.indices.contains(index) {
```
SergioEstevao: this could be rewritten like this:
```
guard index < blockquoteBorderColors.count else {
  return
}
```

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: hmm yep!

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: thats true, I guess is that more performant than `indices.contains` also? (although its a small array)

I was trying to keep similar style to the `func drawBlockquoteBackground` on line 158 which uses an `if let`.

but I guess it doesn't really matter and actually I think a guard makes more sense for both.

Aztec/Classes/TextKit/ParagraphStyle.swift
```diff
@@ -191,6 +191,23 @@ open class ParagraphStyle: NSMutableParagraphStyle, CustomReflectable {
...
+        var depth = 0
+        for position in (0..<listAndBlockquotes.count).reversed() {
+            if listAndBlockquotes[position] is Blockquote {
+                depth = position
+                break
+            }
+        }
+        let indent = CGFloat(depth) * Metrics.listTextIndentation
+        return (depth, indent)
+    }
```
SergioEstevao: Can we please split this in two methods, one to calculate  the depth and another for the indent value?

The one for indent can use the method for depth inside.

MadeByDouglas: yeah, I was just following the established pattern for these vars as they were used for the list and thought a tuple was an elegant solution to get the depth info, but I can split it up. Actually it reminds me I need to merge it with the `blockquoteIndent` var right above it

MadeByDouglas: ok interesting, when I have only `blockquoteIndent` var rather than both that and `blockquoteNestedIndent` then the background rect shifts as you can see here. I guess I will keep both vars to keep original behavior. 
Screen Shot 2019-12-16 at 5 13 53 PM

MadeByDouglas: well you can see the solution I came up with; just one var, but the block quote background always starts at indent of 0

SergioEstevao:
It looks we have a small bug when we have blockquotes inside lists, look at the screenshot bellow from the Standard Demo content:

Screenshot 2019-12-18 at 22 23 22

MadeByDouglas:
oh interesting. I guess its because they share how they indent lists? Which you need so the spacing is correct, but somehow block quote inside a list needs to know to push the content over further from left and yet still start depth at its own count. Hmm.
```
public var blockquoteDepth: Int {
        let listAndBlockquotes = properties.filter({ property in
            return property is Blockquote || property is TextList
        })
        var depth = 0
        for position in (0..<listAndBlockquotes.count).reversed() {
            if listAndBlockquotes[position] is Blockquote {
                depth = position
                break
            }
        }
        return depth
    }

    /// The amount of indent for the list of the paragraph if any.
    ///
    public var listIndent: CGFloat {
        let listAndBlockquotes = properties.filter({ property in
            return property is Blockquote || property is TextList
        })
        var depth = 0
        for position in (0..<listAndBlockquotes.count).reversed() {
            if listAndBlockquotes[position] is TextList {
                depth = position
                break
            }
        }

        return CGFloat(depth) * Metrics.listTextIndentation
    }
```

MadeByDouglas:
should be fixed, check the standard demo now

Aztec/Classes/TextKit/LayoutManager.swift
```diff
@@ -8,8 +8,12 @@ import QuartzCore
 class LayoutManager: NSLayoutManager {
 
     /// Blockquote's Left Border Color
-    ///
-    var blockquoteBorderColor: UIColor? = UIColor(red: 0.52, green: 0.65, blue: 0.73, alpha: 1.0)
+    /// Set the array with how many colors you like, they appear in the order they are set in the array.
+    var blockquoteBorderColors = [UIColor(hexString: "9252FF")!,
```
SergioEstevao: As a default I would like to have just a single color like this: `[UIColor.gray]` then users of the library can configure to their needs.

Example/Example/EditorDemoController.swift
```diff
@@ -181,7 +181,7 @@ class EditorDemoController: UIViewController {
             editorView.richTextView.textColor = UIColor.label
             editorView.richTextView.blockquoteBackgroundColor = UIColor.secondarySystemBackground
             editorView.richTextView.preBackgroundColor = UIColor.secondarySystemBackground
-            editorView.richTextView.blockquoteBorderColor = UIColor.secondarySystemFill
+            editorView.richTextView.blockquoteBorderColors = [.red, .blue, .green]
```
SergioEstevao: To keep the demo consistent with what we have before can we put the `UIColor.secondarySystemFill` as the first color on the array, and then for other colors use the system version to adapt to the dark/light modes.
So it should be something like this in the end:
```
[UIColor.secondarySystemFill, UIColor.systemRed, UIColor.systemBlue, UIColor.systemGreen]
```

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: oh, I had picked such a nice default color theme though. But yeah, I guess that makes sense

MadeByDouglas: should I use `systemGray` instead of `gray` ?

SergioEstevao: At this location because the library still supports iOS11 we should stick to gray that is available on all version of the OS since iOS11.

If you use systemGray then you need to put code to branch it depending of building to iOS 13 or a previous version of the OS.

Example/Example/EditorDemoController.swift
MadeByDouglas: OK I dropped the red because it contrasts a lot and just use teal and blue which go with the gray better I think for default theme

Aztec/Classes/TextKit/LayoutManager.swift
MadeByDouglas: well I thought that but it seems system gray works on older iOS releases, it let me put it without saying had to be iOS 13.

also here it says since iOS 7.0, but documentation says that for all system colors which doesn't make sense https://developer.apple.com/documentation/uikit/uicolor/3173143-systemgray

MadeByDouglas: it seems that semantic colors like `systemBackground` require iOS 13 where as the other system colors like `systemRed` or `systemGray` have a fallback / or have been around that allows them to be used since iOS 7, although it depends, `systemIndigo` requires iOS 13 and `systemPurple` iOS 9 etc.
Screen Shot 2020-01-07 at 4 02 44 PM

SergioEstevao:
I just pushed a small update to fix a scenario where the border color array was emtpy.
All looking good now!

MadeByDouglas:
awesome! thanks for the small fix

PR #1262, opened 2020-01-08
Allow to use headers without bold/shadow traits

### LINKED ISSUE #1185: Allow header styles without bolding
This [merged PR](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1127) updated headers so that they are displayed with a bold font weight.

The [original issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/351) that this addressed proposed that headers should be:

> Bold by default, but customization to non-bold will also be available

I don't see this non-bold option available. We should provide it for those who don't want their headers automatically bolded.

SergioEstevao:
Fixes #1185

This PR adds an option on the library to select the use of bold/shadow on headings formaters configurable for library users.

To test:
 - Start the demo app
 - Check if headers are showing correctly and bold format works correctly
 - Try it again but now with the option `Configuration.headersWithBoldTrait` set to true
 - Check the HTML generated to see if headers don't get extra `strong` elements.


SergioEstevao:
It looks this also fixes: #1613

PR #1264, opened 2020-01-17
Fix ordered list big numbers display

SergioEstevao:
Refs #https://github.com/wordpress-mobile/gutenberg-mobile/issues/1679

This PR increase the spacing provide for items marker to ensure that items up to four digits are displaying using the default font size. It also makes sure that numbers that are large don't overlap the text content of the list.

Simulator Screen Shot - iPhone 11 Pro - 2020-01-17 at 12 15 03

To test:
 - Run the demo app
 - Select the big lists demo
 - Check that items are displayed and when they large they don't overlap the list text but instead start to get cropped on the left side.


SergioEstevao:
@etoledom and @iamthomasbishop I will need your review on this PR when you have some free time.

etoledom:
Hey @SergioEstevao , it works great!
My only concern is that the numbers can get cropped at the left if they have 5 digits or more. Is there anything we can do about it or is it a good compromise already?

Definitely better than overlapping the text!

iamthomasbishop:
As I mentioned in Slack, while this isn't an ideal solution (cuts off when 5+ digits, inconsistent margins, etc.), it's at least an improvement over what we have now. My vote is to ship this improvement and create another ticket to refine further.

SergioEstevao:
@etoledom , @iamthomasbishop I refactored the code and now it supports dynamic indent of lists depending on the list type and the number of digits expected to be in the list.

Here is a screenshot:

Simulator Screen Shot - iPhone 11 - 2020-01-31 at 14 01 02

SergioEstevao:
After some discussion with @iamthomasbishop we got to a final agreement to be align the list up to 3 digits and from there grow dinamically:

image

iamthomasbishop:
This has my blessing — it's about as good as it's gonna get, a nice compromise without cutting off numbers and aligns pretty well (up to 4 digits as mentioned). I did another check of margins/spacing for bullets vs. digits and it checks out. My vote is to :shipit: !

image

PR #1274, opened 2020-03-27
Update gridicons to 1.0

SergioEstevao:
Update the example app to use Gridicons library version 1.0

To test:
 - Just build and run locally.
 - See if icons on the toolbar look correct on the demo app.


SergioEstevao:
@frosty just updated the last deprecated usage.

PR #1275, opened 2020-03-27
Fix underline bug

SergioEstevao:
It looks the [underline fix](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1271) introduced a bug when drawing links that are at the end of the content.
This PR checks the condition where the fix is applied to avoid using it on the wrong scenarios.

To test:
 - Start the demo app
 - Open the NBSP demo content
 - Move the cursor to the end of content ( last empty line)
 - Press backspace
 - Check that the underline is drawn to the end of the linked text

### DIFF
--- a/Example/Example/SampleContent/underline.html
+++ b/Example/Example/SampleContent/underline.html
@@ -2,3 +2,4 @@
 <p><span style="text-decoration: underline;">Alternative&nbsp;underline&nbsp;text</span></p>
 <p>Before link <a href="www.wordpress.com">Link&nbsp;to&nbsp;WordPress</a></p>
 <p>Before link <a href="www.jetpack.com">Link&nbsp;to&nbsp;Jetpack</a> After link</p>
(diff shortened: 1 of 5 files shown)


Example/Example/SampleContent/underline.html
```diff
@@ -2,3 +2,4 @@
 <p><span style="text-decoration: underline;">Alternative&nbsp;underline&nbsp;text</span></p>
 <p>Before link <a href="www.wordpress.com">Link&nbsp;to&nbsp;WordPress</a></p>
 <p>Before link <a href="www.jetpack.com">Link&nbsp;to&nbsp;Jetpack</a> After link</p>
+Link at end of text<a href="www.jetpack.com">Link&nbsp;to&nbsp;Jetpack</a>
```
etoledom: Is it relevant for the test that the las line is not in a `p` tag and `text` and `Link to Jetpack` are together?

SergioEstevao: Yes, because it needs not to have any extra new lines for the bug to be visible.

PR #1282, opened 2020-04-27
Implement support for sup and sub HTML tags.

SergioEstevao:
Adds support for superscript and subscript elements on Aztec.

Screenshot 2020-04-28 at 15 52 26

To test:
 - Run the demo app
 - Check the Standard demo
 - See if the demo superscript and subscript content show up correctly
 - Play around with the elements
 - Swift to HTML mode and see if the HTML is still correct
 - Switch to Visual mode and see if all show correctly.


SergioEstevao:
> Trying to continue typing after the last subscript I couldn't escape it:

I had a look on this, and I think this is expected behaviour, but because we lack buttons on the toolbar to enable disable it we cannot get rid of it.

This was also more evident because the html missed a `<br>` at the end of the line.
Just moving the example to another line and add adding the breakline it works like the others.

PR #1285, opened 2020-05-11
Fix collapse of white spaces when parsing HTML text.

SergioEstevao:
Makes sure that collapsing/removal of whitespaces doesn't happen at the parser level

To test:
 - Check if CI server tests are green.

### DIFF
--- a/Aztec/Classes/Libxml2/Converters/In/HTMLParser.swift
+++ b/Aztec/Classes/Libxml2/Converters/In/HTMLParser.swift
@@ -45,12 +45,16 @@ open class HTMLParser {
         //
         htmlHandleOmittedElem(0)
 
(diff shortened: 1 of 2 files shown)


guarani:
### Unit Tests

I'm not clear about how this change should be unit tested. I added a unit test like this, which has a `<pre>` block with two lines, both of which have leading and trailing whitespace:

```
    func testDisablingCollapseSpacesPreservesWhitespace() {
        let initialHTML = "<pre>    Hello    <br>    World    </pre>"
                           
        // Setup
        let defaultAttributes: [NSAttributedString.Key: Any] = [.font: UIFont.systemFont(ofSize: 14),
                                                                .paragraphStyle: ParagraphStyle.default]

        storage.htmlConverter.shouldCollapseSpaces = false
        storage.setHTML(initialHTML, defaultAttributes: defaultAttributes)

        let expectedResult = "<pre>    Hello    <br>    World    </pre>"
        let result = storage.getHTML()
        XCTAssertEqual(expectedResult, result)
    }
```
If I temporarily revert the change you did in this PR, the above test _should_ fail – but it still passes 🤔. This tells me that tapping the bold format button in the toolbar is doing something different to this test, otherwise this test should fail. I've look at the use of `getHTML` and `setHTML` in `RCTTextView` but haven't found relevant differences.

PR #1294, opened 2020-06-08
Add release documentation.

SergioEstevao:
Add a release document to the repo explaining the steps needed to create a new version.

To test:
 - Review the document to see if all is correct.

### DIFF
--- a/Documentation/ReleaseProcess.md
+++ b/Documentation/ReleaseProcess.md
@@ -2,7 +2,7 @@
 
 ### Step 1: Creating a release branch ###
 
-The branch structure in Aztec iOS is based on the GIT Flow methodology.
+The branch structure in Aztec iOS is based on the [Git Flow](https://nvie.com/posts/a-successful-git-branching-model/) methodology.
 
 These are the main branches involved in the release process:
 ```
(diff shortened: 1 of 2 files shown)


Documentation/ReleaseProcess.md
```diff
@@ -0,0 +1,112 @@
...
+WordPress-Aztec-iOS.podspec
+WordPress-Editor-iOS.podespec
+```
+
+Open a PR on github where you target the `master` branch with the `release/x.y.z` branch and call it `Release x.y.z`
+
+
+### Step 2: Testing the Integration ###
+
+Before going any further, It's normally good practice to test the Aztec integration into WordPress-iOS project, to make sure I won’t have to do the release process twice.
```
mchowning, line 37: * Lowercase "i" in "It's"
* With the change from "I" to "it's" we now also need to update "make sure I won't...". Could be something like "make sure we won't...".

Documentation/ReleaseProcess.md
```diff
@@ -0,0 +1,112 @@
...
+If all is good, push it
+```
+bundle exec pod trunk push WordPress-Editor-iOS.podspec
+```
+
+### Step 5: Closing the milestone ###
+
+For simplicity Aztec uses a single milestone named “Next Stable”. The purpose of this milestone is to be used to assign to it all of the regular issues that are closed and merged into develop during the regular development process.
+
+At this point in the release process you should rename the milestone to the version you’re about to release.
```
mchowning, line 100: Nit: "you're about to release" might be better as "you just released" since we've already merged to `master` and pushed the POD at this point.

PR #1305, opened 2020-09-14
Change the license from GPL v2 to MPL v2

maxme:
All contributors agreed for the license switch in the ticket #1299 . We're now changing the license from GPL v2 to MPL v2.


hypest:
Looks good so far @maxme!

I see a mention of the older LICENSE file in https://github.com/wordpress-mobile/AztecEditor-iOS/blob/185fef360a1e3517c7669daf072aef0e2a488629/Aztec.xcodeproj/project.pbxproj#L524, perhaps we need to open the project in Xcode to update the project file?

hypest:
Also, it seems that the README file is displaying an icon for the license file and picks it up from CocoaPods so, I guess we'll need to push the Pod to CocoaPods before the proper icon is shown in the final README. It shouldn't be a blocker for the PR though.

maxme:
>  , perhaps we need to open the project in Xcode to update the project file?

Fixed in c3ddcde

>  It shouldn't be a blocker for the PR though.

I agree, let's merge this one and then push a new release to CocoaPods.

hypest (approved):
LGTM!

Issue #1224, opened 2019-10-13
Documentation required


mrramdhas:
Expecting detailed documentation regarding how to use, what features available etc.

netspencer:
yes! documentation would be very helpful

virskor:
yes! it's bad to guess functions~

MadeByDouglas:
Yes. Looking through the example, its clear this is a capable and powerful library which I would very much like to use. And yet, holy flying fuck you have made this complicated. The `EditorDemoController.swift` has nearly 2000 lines of code!!! Like, clearly you are smart people at Wordpress as you wrote this whole shindig. And yet, did anyone not take a step back, after writing a very cool library, and then think, hmm, maybe our example is a bit too much. Maybe a simpler example that just shows the basics of how to get started, add a simple toolbar, and doesn't require several other dependencies, might serve people better? anyway... Ok. rant over.

SergioEstevao:
> and doesn't require several other dependencies

@MadeByDouglas what dependencies you are referring to?

> Maybe a simpler example that just shows the basics

@MadeByDouglas what do you think should be in the basic demo?

MadeByDouglas:
So, my frustration is really centered around the `EditorDemoController.swift`

I tried to extract just the essence of it, to make a simpler demo of my own, and it became clear everything was spaghetti and all dependent on each other. I ended up just copying the whole file with 1700+ plus lines feeling there must be a better way.

Ok so **dependencies**, the toolbar uses `Gridicons` pod, not horrible, but now without that I have nothing as far as a demo text editor with controls. Would be great to see using SF Symbols or something built in for a basic format bar. 

Next, even if I don't want to do anything with Wordpress, it turns out the attachment VCs and I believe the VC used for the headers all require `WordPressEditor` So now, it turns out to use this as is, I need to import these other libraries.

Furthermore, a refactor of `EditorDemoController` would be very helpful, to help people learn what everything does, what is needed for a basic editor, and what all the possible configurations can be. Also so people could just copy a snippet or so and have a working implementation.

I would **suggest** refactoring the `EditorDemoController` as is, but also making a new demo that just shows the simplest possible implementation just with the library but still gives what I imagine most users want, a featured rich text editor, a formatting toolbar out of the box, and ability to see source html it generates.

diegoreymendez:
Thanks for the feedback.

[...]

PR #1319, opened 2021-01-07
Replace gridicons with SFSymbols on the Example app

SergioEstevao:
Refs: #1224

This PR replaces the use of Gridicons on the example app with SFSymbols.

This will remove the only third party dependency of this project and will also allow us to remove the use of Carthage on the build system.

| Simulator Screen Shot - iPhone 8 - 2021-01-07 at 14 44 20 |  Simulator Screen Shot - iPhone 8 - 2021-01-07 at 14 44 57 |

Notes:
 - This not changes the minimum version supported by the library, Aztec and WordPressEditor are still supporting iOS 11 and above
 - We are only removing the gridicons on Aztec local Example project, the main app will continue to use Gridicons for the legacy editor
 
To test:
 - Check if the project builds correctly locally.
 - Check if the Example app runs and you can use the toolbar to edit
 - Check the overlays on the image using the Image Overlay sample content.


loremattei (approved):
This looks good to me and I like that we can simplify the build system removing Carthage. 

✅ Check if the project builds correctly locally.
✅ Check if the Example app runs and you can use the toolbar to edit
✅ Check the overlays on the image using the Image Overlay sample content.

I noticed that there's a Warning when building the Example app because of the switch to iOS 13: 
`AztecEditor-iOS/Example/Example/EditorDemoController.swift:455:22: 'init(input:modifierFlags:action:discoverabilityTitle:)' was deprecated in iOS 13.0`.

Since it's a deprecation notice and it's on the sample app, I'm approving the PR anyway, but I'm reporting it as we may want to fix it in this PR.

SergioEstevao:
> I noticed that there's a Warning when building the Example app because of the switch to iOS 13:
AztecEditor-iOS/Example/Example/EditorDemoController.swift:455:22: 'init(input:modifierFlags:action:discoverabilityTitle:)' was deprecated in iOS 13.0.

Good catch @loremattei , I updated the code to remove that warning and removed some missing references to Carthage.

SergioEstevao:
@ceyhun do you mind giving it a look?

ceyhun (approved):
Tested with Xcode 12.3 and Example app built fine without Carthage, icons in toolbar and image overlays looking good. The project builds fine as well :shipit:

PR #1327, opened 2021-03-17
Fix iOS 13 deprecation

knox:
Fixes this fixes a depreaction warning which breaks treating warnings as errors.

To test: 
* Change iOS deployment target to 13 or above
* Enable "Treat warnings as errors"


guarani (approved):
Thanks for the PR! Looks good to me 👍

knox:
@guarani Could we get a release containing this, please?

PR #1333, opened 2021-10-05
Feature - Adds support for the Mark HTML tag

geriux:
This PR adds support for the `mark` HTML tag.

It currently supports setting the text color and there are plans to add support for background-color as well.

<kbd></kbd>

**To test:**
- Open the demo app
- Check either the standard demo or the Gutenberg demo
- Scroll to the bottom to find the new tag
- Expect to see the text with the colors

**Question**: In the console, I'm still getting `Message: Tag mark invalid` is there something else that's missing to avoid that message?


SergioEstevao (approved):
Looking good! :shipit:

SergioEstevao:
> Question: In the console, I'm still getting Message: Tag mark invalid is there something else that's missing to avoid that message?

This warning message comes from the XML parser for some reason it does not recognize the mark element as valid HTML. But it does not have any impact on the parsing reading of the HTML.

PR #1345, opened 2021-12-28
Allow using the default font for the PreFormatter/HeaderFormatter

geriux:
Currently, there's not an option to use a custom font size for the HeaderFormatter since it uses the default values for each level, as well as for the PreFormatter which uses the default monospace font.

This introduces a new configuration flag `useDefaultFont` that will allow to use the `defaultFont` for both formatters.

To test use this [WordPress build](https://github.com/wordpress-mobile/WordPress-iOS/pull/17707#issuecomment-1003071218).


antonis (approved):
Thank you for wrangling this @geriux 👍 
I tested the [generated build](https://github.com/wordpress-mobile/WordPress-iOS/pull/17707) and the app behaves as expected. The code changes look consistent to me too 🎉

PR #1349, opened 2022-02-05
Add Buildkite

jkmassel:
Adds Buildkite support.

In order to make that work, this PR also:

- Disables some tests that don't work on iOS 15.
- Fixes Swift 5 `AnyObject` declarations.

We can either fix the tests, or wait for a future CI image that has an iOS 14 runtime. Might be worth discussing?


Aztec.xcodeproj/xcshareddata/xcschemes/Aztec.xcscheme
```diff
@@ -56,6 +56,17 @@
                BlueprintName = "AztecTests"
                ReferencedContainer = "container:Aztec.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
```
jkmassel, line 59: These tests don't work under iOS 15, so they're disabled for now

fastlane/Fastfile
```diff
@@ -0,0 +1,74 @@
...
+      buildlog_path: File.join(__dir__, '.build', 'logs'),
+      derived_data_path: File.join(__dir__, '.build', 'derived-data'),
+      ensure_devices_found: true
+    )
+
+    destroy_simulator(device)
+  end
+end
+
+def create_simulator(runtime, device)
```
jkmassel, line 37: I was surprised that this wasn't already a `fastlane` action – might be worth moving this into the release toolkit in the future (in order to be able to guarantee that we run tests on the desired runtime).

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,80 @@
...
+  # Common environment values to use with the `env` key.
+  env: &common_env
+    IMAGE_ID: xcode-13
+
+# This is the default pipeline – it will build and test the app
+steps:
+  #################
+  # Build and Test
+  #################
+  - label: "🧪 Build and Test"
```
mokagio, line 14: Muscle memory was going to make me suggest to use 🔬 but I don't know whether that's appropriate for a "Build and Test" step 🤔 😅

Aztec.xcodeproj/project.pbxproj
```diff
@@ -1456,7 +1456,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0800;
-				LastUpgradeCheck = 1230;
+				LastUpgradeCheck = 1310;
```
mokagio, line 1459: Time to upgrade to Xcode 13.2 @jkmassel 😝

fastlane/Fastfile
```diff
@@ -0,0 +1,74 @@
...
+
+    run_tests(
+      workspace: 'Aztec.xcworkspace',
+      scheme: 'WordPressEditor',
+      device: device.name,
+      prelaunch_simulator: true,
+      buildlog_path: File.join(__dir__, '.build', 'logs'),
+      derived_data_path: File.join(__dir__, '.build', 'derived-data'),
+      ensure_devices_found: true
+    )
```
mokagio, line 31: Have you considered splitting these in two lanes and running them in parallel in CI?

I'm guessing because the tests themselves don't take that long to run (the lane runs less than a minute with no `DerivedData` on my 2019 Intel) doing this might actually result in more total CI setup time for the parallel agents, which would be wasteful. 👍

mokagio (approved):
> We can either fix the tests, or wait for a future CI image that has an iOS 14 runtime. Might be worth discussing?

My vote is for fixing the tests. 

I'm also not sure what the state of Aztec is, but I think it's getting closer and closer to be removable? Or do we plan to leave it around for those who want to switch back to the "classic editor" in the app? 

If it's days are numbered, we might as well carry on like this and direct effort to decommissioning it.

.buildkite/pipeline.yml
AliSoftware: "🛠️ " ? 😇

Aztec.xcodeproj/xcshareddata/xcschemes/Aztec.xcscheme
mokagio: I was going to open a PR similar to https://github.com/wordpress-mobile/MediaEditor-iOS/pull/29 here, but I'll wait for input on my [question regarding the future of this library](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1349#pullrequestreview-874770130).

fastlane/Fastfile
jkmassel: Considered, but the overhead is a lot – I'd be open to splitting them if this repo starts doing a lot of builds though? 🤷‍♂️

PR #1363, opened 2022-10-03
Tools upgrade: Ruby 2.7.4, `bash-cache` 2.8.0

mokagio:
I noticed the repo was still using Ruby 2.6.4. There's nothing inherently wrong with it given there hasn't been any update on this in a while.

Still, 2.6.4 doesn't work on Apple Silicon out of the box and that's the recommended development platform internally. Upgrading was straightforward and quick, and might save us some time if/when we'll need to make more changes here.

I tried to upgrade to Xcode 14 as well, but run into [some failures](https://buildkite.com/automattic/aztec-editor-ios/builds/48#01839d1b-3179-4c19-8355-d7f730675d36). Given we are not actively working on this repo, I think we can merge this PR as-is, and follow up on Xcode 14 in due time via #1364.


AliSoftware (approved):
🧹 🧹 🧹

PR #1367, opened 2023-03-06
Point `bash-cache` plugin to new `a8c-ci-toolkit` location

mokagio:
While I was at it, I also updated the version to the latest, 2.13.0.

This was done via:

```
find . -type f -name "*.yml" -exec sed -i '' 's/automattic\/bash-cache#[0-9.]\{1,\}/automattic\/a8c-ci-toolkit#2.13.0/g' {} +
```

If CI is green, we're good to merge.

Internal reference: paaHJt-4z0-p2

---


rynaardb (approved):
LGTM!

PR #1369, opened 2023-06-22
Prevent crash due to malformed data.

Roger-Blinto:
Happened in the wild when pasting bullet list copied from the very same view.


spongemook:
`unarchiveTopLevelObjectWithData` was deprecated in iOS 12.0

suggest code

```swift
do {
    let attributedString = try NSKeyedUnarchiver.unarchivedObject(ofClass: NSAttributedString.self, from: data)
    return attributedString
} catch {
    debugPrint(error.localizedDescription)
    return nil
}
```

Roger-Blinto:
@spongemook Sure, that is more future proof. I would have preferred it as a one-liner, as the original.
The library is for iOS 11 and onwards. If that version is increased there will be one less thing to change.
Not counting this commit, 12 other API:s deprecated in iOS 15 or earlier are being used. Considering the number of issues, non merged pull requests and recent commit frequency I would say the repository is close to being abandoned...

jkmassel:
Hi @Roger-Blinto – sorry about the wait on this – we'd be happy to merge it, but it'd be really helpful to see a test case to go along with it that demonstrates the issue.

Roger-Blinto:
Fixed in a similar way in the master. Not included in latest release yet.

PR #1373, opened 2023-09-08
Use NSString.paragraphRange method due to iOS 17 crash

### LINKED ISSUE #1374: iOS 17 crashes when calling `String.paragraphRange`
**Describe the bug**
iOS 17 crashes when `String.paragraphRange(for:)` is called when the range passed in is for a single paragraph separator character: `\u{2029}`

**To Reproduce**
1. Run the AztecExample demo app in the Simulator or on a real device
2. Tap "Standard Demo"
3. Place the cursor at the end of the top line "Welcome to Aztec"
4. Erase characters until the app crashes

**Expected behavior**
The app to erase the characters without crashing.

**Screenshots**

| Video | Exception |
| - | - |
| <video src="https://github.com/wordpress-mobile/AztecEditor-iOS/assets/2092798/681a25d7-eaa6-4fde-ae61-a10eab2b214c" /> | Screenshot 2023-09-11 at 16 36 49 |

 - Device: iPhone 14
 - OS: iOS 17.0 (21A5326a)

**Additional context**

- This also applies to [String.lineRange(for:)](https://developer.apple.com/documentation/swift/stringprotocol/linerange(for:))
- The NSString equivalent functions do not crash
- Apple Feedback has been submitted: `FB13142115`

The following Playground in Xcode 16 will reproduce the crash:

```swift
import Foundation

/// Fails when using Xcode 16.0 beta 8 Playground and when running on iOS 17 beta.
let paragraphSeparator = "\u{2029}"

/// Try adding a character before or after the separator - it should not crash.
let string = paragraphSeparator
let range = string.startIndex ..< string.endIndex

/// Throws an exception.
let outRange = string.paragraphRange(for: range)
/// Success if this was reached.
print(string)
```

twstokes:
Fixes #1374

Addresses a crash that occurs when calling [`String.paragraphRange(for:)`](https://developer.apple.com/documentation/swift/stringprotocol/paragraphrange(for:)). The workaround for now is to call the [`NSString` equivalent](https://developer.apple.com/documentation/foundation/nsstring/1408548-paragraphrange).

Related issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/21261

- The crash also occurs when calling [`String.lineRange(for:)`](https://developer.apple.com/documentation/swift/stringprotocol/linerange(for:))
- It only seems to happen when a lone [paragraph separator](https://www.compart.com/en/unicode/U+2029) is passed to it: "\u{2029}"

Xcode 16 Playground code that will reproduce the crash:

```swift
import Foundation

/// Fails when using Xcode 16.0 beta 8 Playground and when running on iOS 17 beta.
let paragraphSeparator = "\u{2029}"

/// Try adding a character before or after the separator - it should not crash.
let string = paragraphSeparator
let range = string.startIndex ..< string.endIndex

/// Throws an exception.
let outRange = string.paragraphRange(for: range)
/// Success if this was reached.
print(string)
```

To test:

1. Run the AztecExample demo app in the Simulator or on a real device running iOS 17
2. Tap "Standard Demo"
3. Place the cursor at the end of the top line "Welcome to Aztec"
4. Erase characters until there are no more to erase
5. **Expect:** No exception is thrown and the app doesn't crash.

Repeat the test for iOS 16.

### DIFF
--- a/Aztec/Classes/Extensions/NSAttributedString+ParagraphRange.swift
+++ b/Aztec/Classes/Extensions/NSAttributedString+ParagraphRange.swift
@@ -22,14 +22,13 @@ extension NSAttributedString {
     ///
     func paragraphRanges(intersecting range: NSRange, includeParagraphSeparator: Bool = true) -> [NSRange] {
         var paragraphRanges = [NSRange]()
-        let swiftRange = string.range(fromUTF16NSRange: range)
-        let paragraphsRange = string.paragraphRange(for: swiftRange)
-        
-        string.enumerateSubstrings(in: paragraphsRange, options: .byParagraphs) { [unowned self] (substring, substringRange, enclosingRange, stop) in
+        let paragraphsRange = foundationString.paragraphRange(for: range)
+
+        foundationString.enumerateSubstrings(in: paragraphsRange, options: .byParagraphs) { (substring, substringRange, enclosingRange, stop) in
             let paragraphRange = includeParagraphSeparator ? enclosingRange : substringRange
-            paragraphRanges.append(self.string.utf16NSRange(from: paragraphRange))
+            paragraphRanges.append(paragraphRange)
         }
-        
+
         return paragraphRanges
     }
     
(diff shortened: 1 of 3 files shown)


geriux (approved):
LGTM!

I've tested these changes in both iOS 17 and 16 and it works as expected 🚀 Nice work!

PR #1377, opened 2023-11-04
Use Xcode 15.0.1

jkmassel:
### Migrates the project to Xcode 15 and Apple Silicon in CI

Also includes:
- An adjustment to the `UITextView` subclass that injects a `UIPasteboard` object. This allows us to mock access to the system pasteboard, which requires user interaction in recent iOS versions
- An adjustment to string index handling – in newer versions of Xcode, there are runtime assertions about string index overflow, so we need to be a bit more careful when replacing text 
- A tiny little fix for a failing test, presumably related to something about how iOS generates PNG files.

This can be reviewed by anyone, but I'd appreciate if @mokagio or @crazytonyli took a look at the project code adjustments!

### Testing Details

Ensure CI passes

---

- [x] Please check here if your pull request includes additional test coverage.


Aztec/Classes/Extensions/String+RangeConversion.swift
```diff
@@ -241,4 +249,12 @@ public extension String {
...
         return startIndex ..< endIndex
     }
+
+    func isValidRange(_ range: Range<String.UTF16View.Index>) -> Bool {
+        isValidIndex(range.lowerBound) && isValidIndex(range.upperBound)
+    }
+
+    func isValidIndex(_ index: String.UTF16View.Index) -> Bool {
+        (self.startIndex ... self.endIndex).contains(index)
+    }
```
crazytonyli, line 259: [The `endIndex` is not an accessible index.](https://developer.apple.com/documentation/swift/string/endindex)

> A string’s “past the end” position—that is, the position one greater than the last valid subscript argument.

crazytonyli (changes requested):

Aztec/Classes/Extensions/String+RangeConversion.swift
```diff
@@ -241,4 +249,12 @@ public extension String {
 
         return startIndex ..< endIndex
     }
+
+    func isValidRange(_ range: Range<String.UTF16View.Index>) -> Bool {
+        isValidIndex(range.lowerBound) && isValidIndex(range.upperBound)
+    }
+
+    func isValidIndex(_ index: String.UTF16View.Index) -> Bool {
+        (self.startIndex ... self.endIndex).contains(index)
```
AliSoftware:
```suggestion
        (self.startIndex ..< self.endIndex).contains(index)
```

.buildkite/pipeline.yml
```diff
@@ -24,9 +24,16 @@ steps:
   - label: "🔬 Validate Podspecs"
     key: "validate"
     command: |
-      validate_podspec
+      # validate_podspec
+      echo '+++ ⚠️ validate_podspec was bypassed ⚠️'
+      # post a message in the logs
+      cat .buildkite/validate_podspec_annotation.md
+      # and also as an annotation
+      cat .buildkite/validate_podspec_annotation.md | buildkite-agent annotate --style 'warning'
```
crazytonyli: There is a new workaround for this issue: You can pass a `--patch-cocoapods` option to `validate_podspec` and the publish pod utilities. See https://github.com/wordpress-mobile/WordPressKit-iOS/pull/652/files.

Aztec/Classes/Extensions/String+RangeConversion.swift
jkmassel: This change causes the following test to fail:

```swift
    func testThatEndIndexIsReportedAsValid() {
        let candidate = "This is a string 👍🏽"
        XCTAssertTrue(candidate.isValidIndex(candidate.endIndex))
    }
```

.buildkite/pipeline.yml
```diff
@@ -24,9 +24,11 @@ steps:
   - label: "🔬 Validate Podspecs"
     key: "validate"
     command: |
-      validate_podspec
+      validate_podspec --patch-cocoapods
     env: *common_env
     plugins: *common_plugins
+    agents:
+      queue: upload
```
crazytonyli: This `agents` configuration needs to be removed.

crazytonyli (changes requested):
Maybe updating the publishing pod commands in`.buildkite/publish-aztec-pod.sh` too, with the same `--patch-cocoapods` option? Otherwise, future releasing jobs would fail.

Aztec/Classes/TextKit/TextView.swift
```diff
@@ -495,7 +495,7 @@ open class TextView: UITextView {
     // MARK: - Intercept copy paste operations
 
     open override func cut(_ sender: Any?) {
-        let data = storage.attributedSubstring(from: selectedRange).archivedData()
+        let data = try! storage.attributedSubstring(from: selectedRange).archivedData()
```
crazytonyli, line 498: Does that mean this line crashes if the selected attributed string contains an "attribute value" that doesn't conform to `NSCoding`? I don't really know anything about this library's implementation, not sure how likely that's going to happen...

AliSoftware: That's a good question, but also before the migration to Xcode 15 and newer Swift back when `archivedData` was not marked as `throws`, I'd expect the code would already crash (but maybe with an ObjC NSException rather than a Swift error being thrown)?

That being said, it would be a nice improvement to use `let data = try? …` instead and then make the subsequent call to `storeInPasteboard(encoded: data)` conditional on `if let data`; but not sure if that fix belongs in this PR or separate.

AliSoftware: Also, looking at the deprecated vs new method signature for `NSKeyedArchiver.archivedData`, I'm wondering if the only case where it could `throw` is if you passed `requiresSecureCoding: true` and the objects conformed to `NSCoding`… but not `NSSecureCoding`.

But since here we use `requiresSecureCoding: false` in the implementation of `archivedData`, maybe that means we'll never throw in practice? Or if we do, that would be under the same conditions as the ones the current version of the library already does, i.e. not related to Xcode 13->Xcode15 migration?

AliSoftware (approved):

AliSoftware:
Breadcrumbs: I reverted the attribution of the cluster to this repo in Buildkite (https://github.com/Automattic/buildkite-ci/pull/331) — aka reverted this repo to use Intel infra — while this PR is still not merged.

Once we land this PR, don't forget to redo the change in `aztec-editor-ios.tf` to switch it back to use the cluster again.

crazytonyli (approved):

mokagio:
I wanted to help here and tried to address the unit tests failures via https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1382 but didn't get very far.

mokagio:
Going to merge this despite unit tests failures, to unblock moving to the cluster in CI, which in turns unblocks adopting the latest Xcode, https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1392, which, finally, will unblock https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1391.

I'm still experiencing unit tests issue locally, but I want to see how/if the situation changes once this lands.

PR #1384, opened 2024-02-02
Fix crash when underlining text with special glyphs

### LINKED ISSUE #11866: NSRangeException: Aztec.TextStorage attributedSubstringFromRange:: Out of bounds
Sentry Issue: [WOOCOMMERCE-IOS-1MDH](https://a8c.sentry.io/issues/4895304366/?referrer=github_integration)

```
NSRangeException: Aztec.TextStorage attributedSubstringFromRange:: Out of bounds
  File "LayoutManager.swift", line 369, in LayoutManager.underlineGlyphRange
  File "<compiler-generated>", in LayoutManager.underlineGlyphRange
  File "main.swift", line 7, in main
...
(28 additional frame(s) were not displayed)
```

itsmeichigo:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/11866

## Description
We have a [custom implementation](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1271) for underlining text, and the logic involves checking for a range of glyphs in strings. For text with special glyphs (e.g. Arabic language), this range is sometimes out of the bound of the text (I have yet to find an exact explanation for this).

The fix is to fall back to using the default implementation if we detect that a glyph range is invalid.

## Testing steps
- Run the demo app
- Open the content Underline NBSP
- Check that underlines are drawn correctly and the app doesn't crash.

---

### DIFF
--- a/Aztec/Classes/TextKit/LayoutManager.swift
+++ b/Aztec/Classes/TextKit/LayoutManager.swift
@@ -366,6 +366,12 @@ extension LayoutManager {
         guard let textStorage = textStorage else {
             return
         }
+
+        guard Range(glyphRange, in: textStorage.string) != nil else {
+            // range out of bound for the glyph, fallback to default behavior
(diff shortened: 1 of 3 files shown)


selanthiraiyan (approved):
The changes look good to me. :shipit:

PR #1400, opened 2024-08-13
Rename an API that has the same name as a new API in iOS SDK

crazytonyli:
iOS 18 SDK has a new API [`NSAttributedString.init(attachment:attributes:)`](https://developer.apple.com/documentation/foundation/nsattributedstring/4357343-init?changes=latest_minor). This library has a public API that has the same name.

This PR did a quick fix which adds a meaning less `aztec: Void` argument to the API, so that the library get to keep the existing functionality and won't affect library consumers in using iOS 18 SDK.

---


Example/Example/UnknownEditorViewController.swift
```diff
@@ -11,14 +11,14 @@ class UnknownEditorViewController: UIViewController {
     ///
     fileprivate(set) var saveButton: UIBarButtonItem = {
         let saveTitle = NSLocalizedString("Save", comment: "Save Action")
-        return UIBarButtonItem(title: saveTitle, style: .plain, target: self, action: #selector(saveWasPressed))
+        return UIBarButtonItem(title: saveTitle, style: .plain, target: UnknownEditorViewController.self, action: #selector(saveWasPressed))
```
kean: This doesn't seem right 🤔 
It's now setting a class as a target.

Aztec/Classes/Extensions/NSAttributedString+Attachments.swift
```diff
@@ -15,7 +15,15 @@ extension NSAttributedString
...
     ///
+    @available(*, deprecated, renamed: "init(attachment:attributes:aztec:)", message: "This function has the same name as a new API in iOS 18")
     public convenience init(attachment: NSTextAttachment, attributes: [NSAttributedString.Key: Any]) {
+        self.init(attachment: attachment, attributes: attributes, aztec: ())
+    }
+
+    /// Helper Initializer: returns an Attributed String, with the specified attachment, styled with a given
+    /// collection of attributes.
+    ///
+    public convenience init(attachment: NSTextAttachment, attributes: [NSAttributedString.Key: Any], aztec: Void) {
```
kean, line 26: (nit) I would suggest converting it to a factory method with a prefix: `NSAttributedString.aztec_make(attachment:attributes:)`. Does it need to be public?

kean (changes requested):

Example/Example/UnknownEditorViewController.swift
crazytonyli: Good catch! These are Xcode's code suggestions, which I blindly took... Addressed in https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1400/commits/fc8861a90e17bc9faeab205f3d0faa457ee3e5ee.

Aztec/Classes/Extensions/NSAttributedString+Attachments.swift
crazytonyli: This initialiser is available to both NSAttributedString and NSMutableAttributedString. It's not very straightforward to make a static function to return an instance of its receiver.

> Does it need to be public?

I'm not sure. But I don't want to change the public API as part of this PR though, which is why I marked the original function as deprecated.

crazytonyli:
The unit test failures are not related to this PR. https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1392 seems to fix them.

mokagio (approved):
Looks like all comments have been addressed.

Will need to wait for #1404 and #1391 to merge, then update here, then merge this one, too.

PR #1379, opened 2023-12-05
Support SPM installation

allenhumphreys:
Fixes #1372 #1370

To test:

You can add `.package(url: "https://github.com/allenhumphreys/AztecEditor-iOS", branch: "ah/spm")` to the dependencies in a sample application. If needed I can create a repo with a sample application that consumes the package.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #1406, opened 2025-01-22
Support SPM installation

mokagio:
This is a copy of @allenhumphreys' https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1379 but with `develop` already merged into it, including conflicts resolution and a necessary commit to add a file to the project so that the tests (which at the moment run via the Xcode project setup, not SwiftPM) could build in CI.

I'd love for @allenhumphreys' work to be acknowledged in the GitHub paper-trail. Given this PR is based off #1379, I would expect #1379 to result as merged once this is merged. Alternatively, we could update the base for this to #1379, merge it into it, then merge #1379 into `develop`.

## Where to go from here

After this lands, we should:

- Update the setup to get rid of the `Aztec.xcodeproj` and `Aztec.xcworkspace`. The demo app is already secluded in its own folder and with its own project
- Adopt the standard `Sources/` `Tests/` SwiftPM setup

---


allenhumphreys:
Thanks for the thought, but I’m ok with my name being on the commit and not the PR. The PR numbers you’ve added have taken care of the rest of the history. ❤️

Package.swift
```diff
@@ -0,0 +1,63 @@
+// swift-tools-version: 5.8
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let swiftSettings: [SwiftSetting] = [
+    .define("SPM"),
```
crazytonyli: I don't think you need this. There is an existing built-in flag `SWIFT_PACKAGE`.

allenhumphreys: https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/Usage.md#packaging-legacy-code

Nice!

mokagio: Thanks @crazytonyli !

Addressed via cce8320caae791836c1888d60fb61003b20610bb

crazytonyli (approved):

PR #1408, opened 2025-02-04
Bump version to 1.20.0

mokagio:
Just a process breadcrumb PR. I will admin merge on CI is green.


mokagio:
CI is red, but that's due to the first 1.20.0 attempt partially succeeding on Aztec but failing on Editor, https://buildkite.com/automattic/aztec-editor-ios/builds/194.

The second run, then, failed on Aztec because it was already available on the official CDN, but passed on Editor, https://buildkite.com/automattic/aztec-editor-ios/builds/196.

So at the end of the day, the final result is what we wanted it to be: both Aztec and Editor are available on the official CocoaPods specs at version 1.20.0

PR #1354, opened 2022-03-29
Support alternate bullet style per level

cbess:
Similar to web browsers, draw different bullets at different levels.

To test:
- Create an unordered list
- Add multiple levels to it
- Notice the bullet style for level one is different than level two and three

| Before | After |
| ------ | ------ |
| old | new |

PR #1409, opened 2025-02-22
Support standard bullet styles

cbess:
Reopening: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1354

- Addressed each point in the above PR.
- Added alternate ordered list style support per level
---
Similar to web browsers, draw different bullets at different levels.

To test:
- Set the `listIndentStyle` to `.varied`
- Create an ordered or unordered list
- Add multiple levels to it
- Notice the bullet style for level one is different than level two and three
- Also, execute "Big Lists" demo, works as expected

| Before | After |
| ------ | ------ |
| old | new |
| Screenshot 2025-02-22 at 4 53 29 PM | Screenshot 2025-02-22 at 4 52 42 PM |
---


cbess:
@twstokes it should be ready now :)

twstokes:
> @twstokes it should be ready now :)

Sounds good @cbess! I'll aim to look at it soon.

cbess:
@twstokes any updates?

Aztec/Classes/TextKit/ParagraphProperty/TextList.swift
```diff
@@ -14,14 +15,51 @@ open class TextList: ParagraphProperty {
...
+            case .ordered:
+                if indentLevel == nil {
+                    return "\(number)."
+                }
+                
+                switch indentLevel {
+                case 1:
+                    return "\(number)."
+                case 2:
+                    let text = getLetter(for: number)
```
twstokes, line 30: What do you think about using `NSTextList(markerFormat: .lowercaseAlpha, options: 0)` here so we offload that part to [UIKit](https://developer.apple.com/documentation/uikit/nstextlist)? We'd get the string via `marker(forItemNumber: number)`.

Aztec/Classes/TextKit/ParagraphProperty/TextList.swift
```diff
@@ -14,14 +15,51 @@ open class TextList: ParagraphProperty {
...
+                    return "\(text)."
+                default:
+                    // marker for all levels > 2
+                    let text = getRomanNumeral(for: number)
```
twstokes: Same question as above, but with `NSTextList(markerFormat: .lowercaseRoman, options: 0)`.

Aztec/Classes/TextKit/ParagraphProperty/TextList.swift
cbess: Tried it, but it doesn't match the HTML spec behavior: https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element

Namely, after the 26th item, it should concatenate the letters, then traverse them. So after `z` we get `aa`, then `ab`, etc. The algorithm I implemented reproduces the HTML standard.

Aztec/Classes/TextKit/ParagraphProperty/TextList.swift
```diff
@@ -90,3 +129,22 @@ open class TextList: ParagraphProperty {
         return lhs.style == rhs.style && lhs.start == rhs.start && lhs.reversed == rhs.reversed
     }
 }
+
+/// Returns the letters to use as the ordered list marker text
+fileprivate func getLetter(for number: Int) -> String {
+    let listChars = "abcdefghijklmnopqrstuvwxyz"
```
twstokes, line 135: We could include a `guard` here for positive `int`s, otherwise we'll crash.

Example/Example/EditorDemoController.swift
```diff
@@ -257,7 +258,7 @@ class EditorDemoController: UIViewController {
         rightMargin -= view.safeAreaInsets.right
 
         scrollInsets.right = -rightMargin
-        editorView.scrollIndicatorInsets = scrollInsets
+        editorView.horizontalScrollIndicatorInsets = scrollInsets
```
twstokes, line 261: Thanks for fixing the example code! 😄

twstokes (commented):
@cbess thanks for your work on this.

The only part I'm unsure of is ordered lists:

> Similar to web browsers, draw different bullets at different levels.

The HTML spec supports different bullet styles as you mentioned for ordered lists, but in my testing browsers are rendering what Aztec is already doing by default.

I tested Chrome and Safari with:

```
<ol>
  <li>One</li>
  <ol>
    <li>Two</li>
    <ol>
      <li>Three</li>
    </ol>
  </ol>
</ol>
```

And it renders:

image

This is also reflected in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol). Lastly, I get the same behavior in the iOS Notes app for ordered list nesting. I think in order to fully support ordered list styles we'd need to let the user choose them and have the appropriate HTML markup reflect the choice. _If I'm misunderstanding something please let me know._

For unordered lists I think your changes work well in that they vary the bullets styles, which seems to be standard behavior. 👍

cbess:
@twstokes You're welcome! I appreciate this library, thanks for maintaining it!

You make a fair point about the `ol` bullets. And yes, that was intentional. I should have said:
> Similar to popular wysiwyg HTML editors, draw different bullets at different levels. 

I was really copying examples like [CKEditor](https://ckeditor.com/ckeditor-5/demo/feature-rich/) 😅 

Screenshot 2025-03-31 at 5 44 48 PM

With that being said, perhaps the style can be renamed to `mixed` or `varied` or `assorted` or `multiform` or [something else]?

Aztec/Classes/TextKit/ParagraphProperty/TextList.swift
cbess: Good point. So, I think I did one better, since a valid/useful value needs to be returned.

twstokes:
> With that being said, perhaps the style can be renamed to `mixed` or `varied` or `assorted` or `multiform` or [something else]?

`varied` makes sense to me. 👍

cbess:
Updated enum member and the doc string

twstokes (approved):
LGTM. Thanks @cbess!

PR #1412, opened 2025-05-28
Extract `HTMLParser` in dedicated library

mokagio:
Fixes https://linear.app/a8c/issue/AINFRA-639

The code is already in use in the WooCommerce iOS watchOS app, see https://github.com/woocommerce/woocommerce-ios/pull/15678/commits/ec7a641c28017cb3775b9513ef90fe049d0e11dd

I decide to extract the files under the standard `Sources/` and `Tests/` root folders that SwiftPM expects. To keep the `Package.swift` and `podspec`s consistent, I moved all the other sources and tests there, too.

I updated the plumbing in the example app and ran it locally to ensure it still works, for what is worth.

Screenshot 2025-06-04 at 8 12 01 AM

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. – N.A. many files were moved, but the source change was minimal.


mokagio:
CI is failing with

```
Compiling for iOS 12.0, but module 'Aztec' has a minimum deployment target of iOS 13.0: /Users/builder/Library/Developer/Xcode/DerivedData/App-etmfcijrdcehzwescvrvmhwzlcrt/Build/Products/Debug-iphonesimulator/WordPress-Aztec-iOS/Aztec.framework/Modules/Aztec.swiftmodule/arm64-apple-ios-simulator.swiftmodule
```

but I don't get this locally 🤔 

---

Update: Gave up on trying to fix this. Simply bumped everything to iOS 13 to satisfy it. [bd3eacc](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1412/commits/bd3eaccc72c24841fff4457ca431e46e451cba69)

.buildkite/shared-pipeline-vars
```diff
@@ -5,7 +5,7 @@
 
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E 's/^~> ?//' .xcode-version)
-CI_TOOLKIT_PLUGIN_VERSION="3.8.0"
+CI_TOOLKIT_PLUGIN_VERSION="5.3.1"
```
mokagio, line 8: Routine update.

.buildkite/shared-pipeline-vars
```diff
@@ -5,7 +5,7 @@
...
 
-export IMAGE_ID="xcode-$XCODE_VERSION-macos-14.7.1-v1"
+export IMAGE_ID="xcode-$XCODE_VERSION"
```
mokagio, line 10: Suffix removed because I also updated the tooling to 16.4 which does not need it.

Package.swift
```diff
@@ -1,51 +1,44 @@
-// swift-tools-version: 5.8
+// swift-tools-version: 5.10
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
 
 let package = Package(
     name: "WordPress-AztecEditor-iOS",
-    platforms: [.iOS(.v11)],
+    platforms: [.iOS(.v13)],
```
mokagio, line 8: The reason for the version bump is layered.

This PR extracts HTMLParser as a standalone library so that the WooCommerce watchOS app can use it via SwiftPM, since watchOS does not support UIKit.

I knew the restructuring works when the library is fetched via SwiftPM because both the example app and Woo build and run. 

But the only way I had to verify the CocoaPods setup, which required a few conditional imports (see the rest of the diff) was to run the unit tests as part of the spec verification.

But in CI, the tests failed with:

```
    - ERROR | xcodebuild:  /opt/ci/builds/builder/automattic/aztec-editor-ios/Tests/AztecTests/Extensions/ArrayHelperTests.swift:2:18: error: compiling for iOS 12.0, but module 'Aztec' has a minimum deployment target of iOS 13.0: /Users/builder/Library/Developer/Xcode/DerivedData/App-fnvsuwzsjydtrdcwmvzmpjrcydqz/Build/Products/Debug-iphonesimulator/WordPress-Aztec-iOS/Aztec.framework/Modules/Aztec.swiftmodule/arm64-apple-ios-simulator.swiftmodule
```

However on my local machine the same `bundle exec pod lib lint` command passes.

After a number of unsuccessful attempts to address the CI issue I simply decided to bump the minimum deployment target and move on. After all, all clients of this app already expect version way above iOS 13.

mokagio:
@AliSoftware @iangmaia @twstokes could I ask you for a review?

The code is already in use in WooCommerce, referenced by a commit, but it would be good to merge this. Publishing a new tag would also be neat.

Tests/AztecTests/Converters/AttributesToStringAttributes/BoldElementAttributeConverterTests.swift
```diff
@@ -1,5 +1,8 @@
 import XCTest
 @testable import Aztec
+#if SWIFT_PACKAGE
+@testable import HTMLParser
+#endif
```
AliSoftware, line 5: I'm unsure why this `@testable import` has to be gated by `#if SWIFT_PACKAGE` (while other imports don't)? 🤔

Tests/AztecTests/Importer/InAttributeConverterTests.swift
```diff
@@ -1,6 +1,10 @@
 import XCTest
-@testable import Aztec
 import libxml2
+#if SWIFT_PACKAGE
+@testable import HTMLParser
+#else
+@testable import Aztec
+#endif
```
AliSoftware, line 7: This one is even more confusing to me… why do we import one package in one context (`#if SWIFT_PACKAGE`) and a different one in another? 🤔

AliSoftware (approved):
Ran `bundle exec fastlane test` and passed as expected (well after I updated `TEST_RUNTIME = 'iOS 18.3'` because I didn't have the `18.2` runtime installed on my Mac with the Xcode versions I have and didn't want to lose time downloading the whole runtime just for that while I won't use it afterwards and it'd take lots of disk space)

Left a couple of questions regarding the `#if SWIFT_PACKAGE` imports though, so I disabled auto-merge so you can review them first if needed, but approving to unblock.

Tests/AztecTests/Converters/AttributesToStringAttributes/BoldElementAttributeConverterTests.swift
mokagio: It has to do with the difference between how CocoaPods and SwiftPM build the library, or rather libraries.

In SwiftPM, we now have the standalone HTMLParser library and Aztec depends on it.

But in CocoaPods, I didn't expose HTMLParser. That's a bit of an inconsistency, I'll admit. But it seemed fine to only work in the context of SwiftPM, between the state of CocoaPods and the fact that this change was just to unblock SwiftPM in Woo.

But this situation resulted in a few build differences. The HTMLParser sources are part of Aztec when building with CocoaPods, but when building for SwiftPM (the default in Xcode and xcodebuild/CLI) they aren't.

Tests/AztecTests/Importer/InAttributeConverterTests.swift
mokagio: After the [explanation above](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1412#discussion_r2146039992), this should be clearer.

This test only exercises code that is part of the HTMLParser library. In SwiftPM that is a standalone module. But when building for CocoaPods, that's part of Aztec.

Tests/AztecTests/Converters/AttributesToStringAttributes/BoldElementAttributeConverterTests.swift
AliSoftware: Ok I think I see (even though this is a bit much for my mushy Friday evening brain to take on fully 😅)

This seems like an important explanation, and something someone other that you (who worked on this migration directly and are thus familiar with the kind of hybrid/unbalanced setup you had to end up with) would require some digging to understand this subtlety and the reason behind it.

As such, I'll suggest to document it somewhere in the code. Like maybe a code comment explaining all that in the key/main/root place in the codebase where this `#if` case happens for the first time, then code comments next to all other `#if` you had to similarly add just pointing to that longer comment/explanation (to avoid copy/pasting same long explanatory comment at every single place)? wdyt?

mokagio: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1413

